PR #24, opened 2019-03-30
Mild change to decoder

graebm:
major:
- decode() returns after each response (or request), and must be called again to continue. This gives the connection a chance to check whether we're still in a valid state before continuing.
- rename `aws_http_client_connection_close()` -> `aws_http_connection_close()`.

minor
- remove "100-continue" stuff from decoder. This will eventually be handled by connection.


source/connection_h1.c
```diff
@@ -1130,63 +1128,65 @@ static int s_handler_process_read_message(
...
-    if (err) {
-        AWS_LOGF_ERROR(
-            AWS_LS_HTTP_CONNECTION,
-            "id=%p: Message processing failed, error %d (%s). Closing connection.",
-            (void *)&connection->base,
-            err,
-            aws_error_name(err));
+    /* Run decoder until all message data is processed */
+    struct aws_byte_cursor message_cursor = aws_byte_cursor_from_buf(&message->message_data);
+    while (message_cursor.len > 0) {
```
justinboswell: How much data could conceivably be backed up here? This appears to have removed the shutdown check, so that would mean shutdown has to decode potentially a lot of data and deliver it before completing? Is that desirable?

graebm: We're actually checking for shutdown more frequently now.

It used to be that a call to decode() would process ALL the data passed in, possibly processing multiple responses.

Now decode() exits after each response is processed, so we call it in a loop, and in each loop we check that the connection isn't shutting down, and that there is in fact a pending-request to receive the response

PR #26, opened 2019-04-02
log errors in decoder

graebm:
- log errors in the decoder
- misc modifications while I'm in there and thinking about it


source/decode.c
```diff
@@ -497,17 +616,22 @@ static int s_linestate_response(struct aws_http_decoder *decoder, struct aws_byt
...
     if (!aws_byte_cursor_eq(&version, &version_expected)) {
+        AWS_LOGF_ERROR(
+            AWS_LS_HTTP_STREAM, "id=%p: Incoming response uses unsupported HTTP version.", decoder->logging_id);
         return aws_raise_error(AWS_ERROR_HTTP_PARSE);
     }
 
-    int64_t code_val;
-    int err = s_read_int64(code, &code_val);
+    size_t code_val;
+    err = s_read_size(code, &code_val, 10);
```
JonathanHenson: why 10?

source/decode.c
```diff
@@ -404,48 +467,89 @@ static int s_linestate_header(struct aws_http_decoder *decoder, struct aws_byte_
...
+                } else if (
+                    aws_string_eq_byte_cursor_ignore_case(s_transfer_coding_gzip, &coding) ||
+                    aws_string_eq_byte_cursor_ignore_case(s_transfer_coding_x_gzip, &coding)) {
+                    /* A recipient SHOULD consider "x-gzip" to be equivalent to "gzip". RFC-7230 4.2.3 */
+                    decoder->transfer_encoding |= AWS_HTTP_TRANSFER_ENCODING_GZIP;
+
+                } else if (coding.len > 0) {
+                    AWS_LOGF_ERROR(
+                        AWS_LS_HTTP_STREAM,
+                        "id=%p: Incoming transfer-encoding header lists unrecognized coding.",
```
justinboswell: [nitpick] Wouldn't it be an unrecognied encoding?

source/decode.c
```diff
@@ -463,15 +567,27 @@ static int s_linestate_header(struct aws_http_decoder *decoder, struct aws_byte_
...
+    }
+
     struct aws_byte_cursor method = cursors[0];
     struct aws_byte_cursor uri = cursors[1];
     struct aws_byte_cursor version = cursors[2];
 
     struct aws_byte_cursor version_expected = aws_http_version_to_str(AWS_HTTP_VERSION_1_1);
     if (!aws_byte_cursor_eq(&version, &version_expected)) {
+        AWS_LOGF_ERROR(
+            AWS_LS_HTTP_STREAM, "id=%p: Incoming request uses unsupported HTTP version.", decoder->logging_id);
```
justinboswell: Why not log the unexpected version?

source/decode.c
```diff
@@ -579,3 +703,7 @@ int aws_http_decoder_get_encoding_flags(const struct aws_http_decoder *decoder)
 size_t aws_http_decoder_get_content_length(const struct aws_http_decoder *decoder) {
     return decoder->content_length;
 }
+
+void aws_http_decoder_set_logging_id(struct aws_http_decoder *decoder, void *id) {
```
justinboswell: AFAICT this function is never called. Is that intentional?

source/decode.c
```diff
@@ -497,17 +616,22 @@ static int s_linestate_response(struct aws_http_decoder *decoder, struct aws_byt
...
     if (err) {
+        AWS_LOGF_ERROR(
+            AWS_LS_HTTP_STREAM, "id=%p: Failed to parse status code from incoming response.", decoder->logging_id);
         return AWS_OP_ERR;
     }
 
     /* Status-code is a 3-digit integer. RFC7230 section 3.1.2 */
-    if (code.len != 3 || code_val < 0 || code_val > 999) {
+    if (code.len != 3 || code_val > 999) {
+        AWS_LOGF_ERROR(AWS_LS_HTTP_STREAM, "id=%p: Incoming response has invalid status code", decoder->logging_id);
```
justinboswell: Worth logging some form of what that code was?

source/decode.c
```diff
@@ -404,48 +467,89 @@ static int s_linestate_header(struct aws_http_decoder *decoder, struct aws_byte_
...
+                    AWS_LS_HTTP_STREAM,
+                    "id=%p: Incoming headers for both content-length and transfer-encoding received. This is illegal.",
+                    decoder->logging_id);
+                return aws_raise_error(AWS_ERROR_HTTP_PARSE);
+            }
+
+            if (s_read_size(header.value_data, &decoder->content_length, 10) != AWS_OP_SUCCESS) {
+                AWS_LOGF_ERROR(
+                    AWS_LS_HTTP_STREAM,
+                    "id=%p: Incoming content-length header has invalid value.",
```
justinboswell: Do you want to log what the invalid value was?

source/decode.c
graebm: basis 10
I thought this was your interview question

source/decode.c
graebm: They come in a comma separated list, and apparently empty entries are ok

source/decode.c
graebm: Youuuuuu ARE RIGHT!
thanks
fixin it

source/decode.c
graebm: I was on the fence about whether to log invalid stuff. Like, if a header is invalid, do I log the bad string? Is this any kind of security risk? Or might I spit out 1000 lines of binary data?

thoughts?

graebm: logging it.
It looks lovely btw

source/decode.c
graebm: did

source/decode.c
graebm: did

source/decode.c
graebm: renamed function so this isn't confusing

source/decode.c
justinboswell: I just mean linguistically :)

PR #27, opened 2019-04-03
Stop decoding if user callback closes connection

graebm:
Brings work to an immediate stop if user calls `aws_http_connection_close()` from a callback.


source/connection_h1.c
```diff
@@ -1247,7 +1276,8 @@ static int s_handler_shutdown(
     (void)free_scarce_resources_immediately;
     struct h1_connection *connection = handler->impl;
 
-    /* Shut everything down the first time we get this callback (DIR_READ). */
+    /* Shutdown happens first in DIR_READ, then in DIR_WRITE with an event-loop tick in between.
+     * Invoke completion callbacks on 2nd pass so they don't fire from within a user's call to close_connection() */
```
JonathanHenson: I really don't like the idea of relying on behavior that isn't enforced by a contract. 
If you want a task then I suggest scheduling one explicitly.

graebm: And I would have gotten away with it too, if I hadn't commented what I was doing.



FIIIIIIIIIIIIIIINE

PR #29, opened 2019-04-05
aws_http_connection_is_open()

graebm:
- Added `aws_http_connection_is_open()`
- Modified `aws_http_connection_close()` so it made the connection **immediately not open**, even if called from outside the event-loop thread.


include/aws/http/connection.h
```diff
@@ -143,6 +143,12 @@ void aws_http_connection_release(struct aws_http_connection *connection);
 AWS_HTTP_API
 void aws_http_connection_close(struct aws_http_connection *connection);
 
+/**
+ * Returns true if the connection is thought to be good (not closed or closing).
+ */
+AWS_HTTP_API
+bool aws_http_connection_ok(const struct aws_http_connection *connection);
```
justinboswell: Perhaps aws_http_connection_is_open or _connected or something? Ok is a bit colloquial, and less clear than it could be.

graebm: is_open() sounds good
I wasn't in love with "ok"

PR #30, opened 2019-04-08
Misc cleanup of aws-c-http

graebm:
- Remove "TODO" comments that exist as tasks on our planning board.
- Write unit tests to address some TODO comments.
- Check for overflow when summing numbers.
- Don't log contents of bad data at the ERROR level, log it at DEBUG. This is to avoid log injection attacks.


tests/test_h1_client.c
```diff
@@ -876,6 +941,121 @@ H1_CLIENT_TEST_CASE(h1_client_response_with_too_much_data_shuts_down_connection)
...
+    /* send response */
+    ASSERT_SUCCESS(s_send_response_str(&tester, "HTTP/1.1 200 OK\r\n\r\n"));
+
+    /* tick loop until body finishes sending.*/
+    while (body_sender.cursor.len > 0) {
+        testing_channel_execute_queued_tasks(&tester.testing_channel);
+    }
+
+    /* check result */
+    const char *expected = "PUT /plan.txt HTTP/1.1\r\n"
```
justinboswell: TODO by way of unit test. I like it.

PR #33, opened 2019-04-15
websocket decoder

graebm:
https://tools.ietf.org/html/rfc6455#page-28
```
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-------+-+-------------+-------------------------------+
 |F|R|R|R| opcode|M| Payload len |    Extended payload length    |
 |I|S|S|S|  (4)  |A|     (7)     |             (16/64)           |
 |N|V|V|V|       |S|             |   (if payload len==126/127)   |
 | |1|2|3|       |K|             |                               |
 +-+-+-+-+-------+-+-------------+ - - - - - - - - - - - - - - - +
 |     Extended payload length continued, if payload len == 127  |
 + - - - - - - - - - - - - - - - +-------------------------------+
 |                               |Masking-key, if MASK set to 1  |
 +-------------------------------+-------------------------------+
 | Masking-key (continued)       |          Payload Data         |
 +-------------------------------- - - - - - - - - - - - - - - - +
 :                     Payload Data continued ...                :
 + - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - +
 |                     Payload Data continued ...                |
 +---------------------------------------------------------------+
```


include/aws/http/private/websocket_decoder.h
```diff
@@ -0,0 +1,111 @@
...
+    AWS_WEBSOCKET_OPCODE_CLOSE = 0x8,
+    AWS_WEBSOCKET_OPCODE_PING = 0x9,
+    AWS_WEBSOCKET_OPCODE_PONG = 0xA,
+};
+
+/**
+ * Full contents of a websocket frame, excluding the payload.
+ */
+struct aws_websocket_frame {
+    bool fin;
```
JonathanHenson: are you relying on this structure layout for serialization/deserialization? If not, it might be good to rearrange the member ordering for better packing.

include/aws/http/private/websocket_decoder.h
```diff
@@ -0,0 +1,111 @@
...
+    aws_websocket_decoder_frame_fn *on_frame,
+    aws_websocket_decoder_payload_fn *on_payload,
+    void *user_data);
+
+/**
+ * Returns when all data is processed, or a frame and its payload have completed.
+ * `data` will be advanced to reflect the amount of data processed by this call.
+ * `frame_complete` will be set true if this call returned due to completion of a frame.
+ * The `on_frame` and `on_payload` callbacks may each be invoked once as a result of this call.
+ * If an error occurs, the decoder is invalid forevermore.
```
JonathanHenson: "forevermore" sparks joy

source/websocket_decoder.c
```diff
@@ -0,0 +1,331 @@
...
+
+    /* first bit is a bool */
+    decoder->current_frame.masked = byte & 0x80;
+
+    /* remaining 7 bits are payload length */
+    decoder->current_frame.payload_length = byte & 0x7F;
+
+    /* If 7bit payload length has a high value, then the next few bytes contain the real payload length */
+    if (decoder->current_frame.payload_length < VALUE_FOR_2BYTE_EXTENDED_LENGTH) {
+        decoder->state += 2; /* Skip next state, which would have processed extended length */
```
JonathanHenson: I think I'd rather see the state increments be just setting to a state for better readability:

decoder->state = AWS_WEBSOCKET_PARSER_STATE_X;

source/websocket_decoder.c
```diff
@@ -0,0 +1,331 @@
...
+    if (decoder->current_frame.payload_length == VALUE_FOR_2BYTE_EXTENDED_LENGTH) {
+        total_bytes_extended_length = 2;
+        min_acceptable_value = VALUE_FOR_2BYTE_EXTENDED_LENGTH;
+        max_acceptable_value = UINT16_MAX;
+    } else {
+        assert(decoder->current_frame.payload_length == VALUE_FOR_8BYTE_EXTENDED_LENGTH);
+
+        total_bytes_extended_length = 8;
+        min_acceptable_value = UINT16_MAX + 1;
+        max_acceptable_value = 0x8000000000000000ULL - 1;
```
JonathanHenson: Why not just INT64_MAX?

source/websocket_decoder.c
```diff
@@ -0,0 +1,331 @@
...
+
+    /* If all data consumed, proceed to next state. */
+    if (decoder->state_bytes_processed == decoder->current_frame.payload_length) {
+        decoder->state++;
+    }
+
+    return AWS_OP_SUCCESS;
+}
+
+static state_fn *s_state_vtable[AWS_WEBSOCKET_DECODER_STATE_DONE] = {
```
JonathanHenson: this isn't really a vtable, but i'm not good at naming things, so.... you decide.

source/websocket_decoder.c
graebm: Spec says it's a crime to set the high bit, even though it should be treated as an unsigned u64

include/aws/http/private/websocket_decoder.h
graebm: I don't serialize this directly, but I put the members in this order because that's how they come in over the wire, and thought keeping the order might help people understand the code. I could have bit-packed them, but since I don't expect many of these to be in existence I didn't think the complication was worth it.

include/aws/http/private/websocket_decoder.h
graebm: Me too :)

source/websocket_decoder.c
graebm: I thought that the incrementing thing made it more flexible, if states were added/removed in the future. Every state does += 1, except BEGIN states that sometimes do += 2 to skip their CONTINUE states.

source/websocket_decoder.c
graebm: s_state_functions

PR #34, opened 2019-04-16
Fallback for ALPN is now http 1.1. We'll likely change this up a bit …

JonathanHenson:
…when we add h2 support.

### DIFF
--- a/source/connection.c
+++ b/source/connection.c
@@ -94,13 +94,12 @@ static struct aws_http_connection *s_connection_new(
             } else if (aws_string_eq_byte_buf(s_alpn_protocol_http_2, &protocol)) {
                 version = AWS_HTTP_VERSION_2;
             } else {
-                AWS_LOGF_ERROR(
+                AWS_LOGF_WARN(
                     AWS_LS_HTTP_CONNECTION,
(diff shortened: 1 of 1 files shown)


source/connection.c
```diff
@@ -94,13 +94,12 @@ static struct aws_http_connection *s_connection_new(
             } else if (aws_string_eq_byte_buf(s_alpn_protocol_http_2, &protocol)) {
                 version = AWS_HTTP_VERSION_2;
             } else {
-                AWS_LOGF_ERROR(
+                AWS_LOGF_WARN(
                     AWS_LS_HTTP_CONNECTION,
-                    "static: Unrecognized ALPN protocol '" PRInSTR "'.",
+                    "static: Unrecognized ALPN protocol '" PRInSTR "'. Assuming Http/1.1",
```
graebm: trivial: please capitalize like "HTTP/1.1"

source/connection.c
```diff
@@ -94,13 +94,12 @@ static struct aws_http_connection *s_connection_new(
...
                     AWS_BYTE_BUF_PRI(protocol));
```
graebm: while you're in here, maybe log the string separately at DEBUG level to reduce risk of log-injection attack

PR #35, opened 2019-04-18
Integration tests, fixed compile warning in websockets test, fixed se…

JonathanHenson:
…gfault in elasticurl.


tests/test_websocket_decoder.c
```diff
@@ -185,7 +185,8 @@ DECODER_TEST_CASE(websocket_decoder_rsv) {
         };
 
         /* Set the appropriate RSV */
-        input[0] |= (uint8_t)(1 << (6 - rsv));
+        /* the bit arithmetic is setup this way to avoid Conversion warnings from the compiler. */
+        input[0] = (uint8_t)(input[0] | (1 << (6 - rsv)));
```
graebm: I wonder if it's a better idea to just disable the compiler warnings in places than rewrite bit-twiddling code so it's even harder to understand. Bit-twiddling code is hard enough to grok as it is.

integration-testing/elasticurl_test.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
graebm: Is it worth writing this in Python so we don't need a separate script for Windows?

integration-testing/elasticurl_test.sh
```diff
@@ -0,0 +1,22 @@
...
+
+set -e
+set -x
+
+echo "elasticurl binary is located at $1"
+
+$1 -v TRACE example.com
+$1 -v TRACE -P -H "content-type: application/json" -i -d "{'test':'testval'}" http://httpbin.org/post
+$1 -v TRACE -o elastigirl.png https://s3.amazonaws.com/code-sharing-aws-crt/elastigirl.png
```
graebm: comments pls

`# Sanity check, see if request/response complete without error`

I'm not really sure what the 2nd cmd is testing

`# Download a file with elasticurl and curl, ensure the results are identical`

tests/test_websocket_decoder.c
JonathanHenson: I strongly dislike the wConversion option as worthless due to its false positive rate. If you all want to remove the warning fine. But it doesn’t build as is.

integration-testing/elasticurl_test.sh
JonathanHenson: 2nd one actually sends body data

PR #36, opened 2019-04-18
Websocket encoder

graebm:
- Implement encoder, works a lot like the decoder.
- Lots of tests.
- Move some stuff out from decoder code to shared private header.


source/websocket_decoder.c
```diff
@@ -108,27 +120,20 @@ static int s_state_extended_length(struct aws_websocket_decoder *decoder, struct
...
-    if (decoder->current_frame.payload_length == VALUE_FOR_2BYTE_EXTENDED_LENGTH) {
+    if (decoder->current_frame.payload_length == AWS_WEBSOCKET_7BIT_VALUE_FOR_2BYTE_EXTENDED_LENGTH) {
         total_bytes_extended_length = 2;
-        min_acceptable_value = VALUE_FOR_2BYTE_EXTENDED_LENGTH;
-        max_acceptable_value = UINT16_MAX;
+        min_acceptable_value = AWS_WEBSOCKET_2BYTE_EXTENDED_LENGTH_MIN_VALUE;
+        max_acceptable_value = AWS_WEBSOCKET_2BYTE_EXTENDED_LENGTH_MAX_VALUE;
     } else {
-        assert(decoder->current_frame.payload_length == VALUE_FOR_8BYTE_EXTENDED_LENGTH);
+        assert(decoder->current_frame.payload_length == AWS_WEBSOCKET_7BIT_VALUE_FOR_8BYTE_EXTENDED_LENGTH);
```
justinboswell: Probably shouldn't crash here, since this is directly caused by data? Instead raise an error about format and get out.

source/websocket_encoder.c
```diff
@@ -0,0 +1,347 @@
...
+    /* Shorten cursor if it won't all fit in out_buf */
+    bool all_data_written = true;
+    size_t space_available = out_buf->capacity - out_buf->len;
+    if (cursor.len > space_available) {
+        cursor.len = space_available;
+        all_data_written = false;
+    }
+
+    aws_byte_buf_write_from_whole_cursor(out_buf, cursor);
+    encoder->state_bytes_processed += cursor.len;
```
justinboswell: This took me a few minutes to work out, maybe a comment would help, because advance also shrinks the len. Offhand, it looks like you're adding too much here.

source/websocket_encoder.c
```diff
@@ -0,0 +1,347 @@
...
+    const struct aws_byte_buf prev_buf = *out_buf;
+
+    /* Invoke callback which will write to buffer */
+    bool user_says_done = false;
+    int err = encoder->stream_outgoing_payload(out_buf, &user_says_done, encoder->user_data);
+    if (err) {
+        return AWS_OP_ERR;
+    }
+
+    /* Ensure that user did not commit forbidden acts upon the out_buf */
```
justinboswell: :+1:

Is a crash the only way out here? Could we log something?

source/websocket_encoder.c
```diff
@@ -0,0 +1,347 @@
...
+    s_state_masking_key_check,
+    s_state_masking_key,
+    s_state_payload_check,
+    s_state_payload,
+};
+
+int aws_websocket_encoder_process(struct aws_websocket_encoder *encoder, struct aws_byte_buf *out_buf) {
+
+    /* Run state machine until frame is completely decoded, or the state stops changing.
+     * Note that we don't necessarily stop looping when out_buf is full, because not all state need to output data */
```
justinboswell: states

source/websocket_encoder.c
```diff
@@ -0,0 +1,347 @@
...
+
+        int err = s_state_functions[encoder->state](encoder, out_buf);
+        if (err) {
+            return AWS_OP_ERR;
+        }
+
+        if (prev_state == encoder->state) {
+            /* This assert checks that each state is doing as much work as it possibly can.
+             * Except for the PAYLOAD state, where it's up to the user to fill the buffer. */
+            assert((out_buf->len == out_buf->capacity) || (encoder->state == AWS_WEBSOCKET_ENCODER_STATE_PAYLOAD));
```
justinboswell: Is this just for you in dev? We'd rather not crash here later, right?

source/websocket_encoder.c
```diff
@@ -0,0 +1,347 @@
...
+    /* Error-check as much as possible before accepting next frame */
+    if (encoder->is_frame_in_progress) {
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
+
+    /* RFC-6455 Section 5.2 contains all these rules... */
+
+    /* Opcode must fit in 4bits */
+    if (frame->opcode != (frame->opcode & 0x0F)) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
```
justinboswell: Logging?

include/aws/http/private/websocket_encoder.h
```diff
@@ -0,0 +1,67 @@
...
+};
+
+struct aws_websocket_encoder {
+    enum aws_websocket_encoder_state state;
+    uint64_t state_bytes_processed;
+    struct aws_websocket_frame frame;
+    bool is_frame_in_progress;
+
+    /* True when the next data frame must be a CONTINUATION frame */
+    bool expecting_continuation_data_frames;
```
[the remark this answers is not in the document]
graebm: It means we're expected 1 or more.
I'll change it. Want it to be as clear as possible.

source/websocket_decoder.c
graebm: This particular value was set by the previous state (if user passes in large value, we encode a magic number that says how many bytes we'll use to encode the real value). I'll add a comment

source/websocket_encoder.c
graebm: Yeah, this should get an FATAL log. I'm going to add logging in a subsequent pass

source/websocket_encoder.c
graebm: Added comments.
We should probably be using byte_cursor advance more often. The class is supposed to be like a "cursor" that moves forward as you use it. It's got some weirdness I asked about in the c-chat the other day.

source/websocket_encoder.c
graebm: it's a dev-assert

source/websocket_encoder.c
graebm: will do in another PR.
There's a /* TODO: logging */ comment at the top of the file.

PR #37, opened 2019-04-26
Added unit tests for TLS negotiation failure and download of a medium…

justinboswell:
…-size file


source/connection_h1.c
```diff
@@ -445,7 +445,7 @@ struct aws_http_stream *s_new_client_request_stream(const struct aws_http_reques
 
     wrote_all &= aws_byte_buf_write_u8(&stream->outgoing_head_buf, '\r');
     wrote_all &= aws_byte_buf_write_u8(&stream->outgoing_head_buf, '\n');
-
+    (void)wrote_all;
```
bretambrose: What does this do?

codebuild/common-posix.sh
```diff
@@ -31,8 +31,8 @@ if [ "$TRAVIS_OS_NAME" != "osx" ]; then
     sudo apt-get install libssl-dev -y
     install_library s2n 7c9069618e68214802ac7fbf45705d5f8b53135f
 fi
-install_library aws-c-common
-install_library aws-c-io
+install_library aws-c-common v0.3.7
```
bretambrose: Was this intended to be temporary?

tests/test_tls.c
```diff
@@ -0,0 +1,315 @@
...
+    bool client_connection_is_shutdown;
+
+    struct aws_mutex wait_lock;
+    struct aws_condition_variable wait_cvar;
+    int wait_result;
+};
+
+static const size_t TEST_TIMEOUT_SEC = 20;
+
+void s_on_connection_setup(struct aws_http_connection *connection, int error_code, void *user_data) {
```
bretambrose: Something seems off about the concurrency here.  Main thread does a lock - cond wait - unlock as expected, but these notification callbacks never do a lock before modifying shared state and then they do an unlock which is maybe undefined since there was no corresponding lock?  Also mutex wasn't inited/cleaned up in the timeout test either, although bitwise zero is essentially init for our current platforms.

justinboswell: Good catch, the mutex was an omission of copy-pasta, and the mutex/cvar init/clean_up was missing from the first test...yow

codebuild/common-posix.sh
justinboswell: no, I want to pin these long term, we should not have a downstream build that isn't pinned, it leads to CI dying due to no fault of the comitter.

source/connection_h1.c
justinboswell: It prevents a compiler warning that the value is never read when asserts are compiled out.

tests/test_tls.c
```diff
@@ -305,6 +314,8 @@ static int s_test_tls_download_medium_file(struct aws_allocator *allocator, void
     aws_logger_set(NULL);
     aws_logger_clean_up(&test.logger);
 
+    aws_mutex_clean_up(&test.wait_lock);
```
bretambrose: These were actually cleaned up above (line 303)

PR #41, opened 2019-05-07
HPACK Static & Dynamic Tables

ColdenCullen:
This is just the initial implementation, serving just to prove out the static & dynamic tables.
Also included is:
* an implementation of integer encoding and decoding
* a rudimentary prototype (read:untested) of string encoding


ColdenCullen:
Note that this won't built until awslabs/aws-c-common#338 is merged.

source/hpack.c
```diff
@@ -0,0 +1,405 @@
...
+
+    return AWS_OP_ERR;
+}
+
+int aws_hpack_insert_header(struct aws_hpack_context *context, const struct aws_http_header *header) {
+
+    /* Cache state */
+    const size_t old_index_0 = context->dynamic_table.index_0;
+
+    /* Increment index 0, wrapping if necessary */
```
justinboswell: Decrementing?

tests/test_hpack.c
```diff
@@ -0,0 +1,294 @@
...
+    ASSERT_UINT_EQUALS(3, output_buf.len);
+    ASSERT_UINT_EQUALS(UINT8_MAX >> 3, output_buffer[0]);
+    ASSERT_UINT_EQUALS(154, output_buffer[1]);
+    ASSERT_UINT_EQUALS(10, output_buffer[2]);
+    ASSERT_UINT_EQUALS(0, output_buffer[3]);
+
+    return AWS_OP_SUCCESS;
+}
+
+AWS_TEST_CASE(hpack_decode_integer, test_hpack_decode_integer)
```
JonathanHenson: Do we need a test case proving the short buffer branches?

tests/test_hpack.c
```diff
@@ -0,0 +1,294 @@
...
+     * +---+---+---+---+---+---+---+---+
+     */
+    uint8_t test_3[] = {UINT8_MAX >> 3, 154, 10};
+    to_decode = aws_byte_cursor_from_array(test_3, AWS_ARRAY_SIZE(test_3));
+    ASSERT_SUCCESS(aws_hpack_decode_integer(&to_decode, 5, &result));
+    ASSERT_UINT_EQUALS(0, to_decode.len);
+    ASSERT_UINT_EQUALS(1337, result);
+
+    return AWS_OP_SUCCESS;
+}
```
JonathanHenson: negative tests?

include/aws/http/private/hpack.h
```diff
@@ -0,0 +1,49 @@
...
+/* General HPACK API */
+struct aws_hpack_context *aws_hpack_context_new(struct aws_allocator *allocator, size_t max_dynamic_elements);
+void aws_hpack_context_destroy(struct aws_hpack_context *context);
+struct aws_http_header *aws_hpack_get_header(struct aws_hpack_context *context, size_t index);
+int aws_hpack_find_index(struct aws_hpack_context *context, const struct aws_http_header *header, size_t *index);
+int aws_hpack_insert_header(struct aws_hpack_context *context, const struct aws_http_header *header);
+int aws_hpack_resize_dynamic_table(struct aws_hpack_context *context, size_t new_max_elements);
+
+/* Public for testing purposes */
+int aws_hpack_encode_integer(uint64_t integer, uint8_t prefix_size, struct aws_byte_buf *output);
```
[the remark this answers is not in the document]
ColdenCullen: No, it doesn't need to be exported.

source/hpack.c
```diff
@@ -0,0 +1,405 @@
...
+        return false;
+    }
+
+    /* If the header stored in the table doesn't have a value, then it's a match */
+    return !right->value.ptr || aws_byte_cursor_eq(&left->value, &right->value);
+}
+
+void aws_hpack_static_table_init(struct aws_allocator *allocator) {
+
+    aws_hash_table_init(
```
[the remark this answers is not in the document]
ColdenCullen: Yes, but I'm not quite sure what we would do about that? library init is a void function. @graebm is it possible to have library_init fail?

source/hpack.c
```diff
@@ -0,0 +1,405 @@
...
+            &context->dynamic_table.reverse_lookup,
+            &context->dynamic_table.buffer[context->dynamic_table.index_0],
+            NULL,
+            NULL)) {
+        goto error;
+    }
+
+    /* Write the new header */
+    if (aws_hash_table_put(
+            &context->dynamic_table.reverse_lookup, header, (void *)context->dynamic_table.index_0, NULL)) {
```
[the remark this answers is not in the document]
ColdenCullen: No, because it's being copied into the list. That said, we should probably use that pointer instead of this one.

tests/test_hpack.c
ColdenCullen: What do you mean? What part would fail?

include/aws/http/private/hpack.h
```diff
@@ -0,0 +1,49 @@
...
+int aws_hpack_decode_integer(struct aws_byte_cursor *to_decode, uint8_t prefix_size, uint64_t *integer);
```
graebm: What if the bytes to decoder are split across aws_io_message boundaries?

Will we know how many incoming bytes to buffer before calling this function?

source/hpack.c
```diff
@@ -0,0 +1,413 @@
...
+        *output->buffer |= prefix_mask;
+        ++output->len;
+
+        integer -= prefix_mask;
+
+        const uint64_t hi_57bit_mask = UINT64_MAX - (UINT8_MAX >> 1);
+
+        while (integer) {
+            if (output->len == output->capacity) {
+                return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
```
graebm: This can fail after writing SOME output

we should probably detect the amount of space we need and fail early
Or take some other approach where we use state to track what's partially encoded

Or a failure here i fatal to the user, which is fine if that's the way we expect to use it.

source/hpack.c
```diff
@@ -0,0 +1,413 @@
...
+    byte &= prefix_mask;
+
+    /* No matter what, the first byte's value is always added to the integer */
+    *integer = byte;
+
+    if (byte == prefix_mask) {
+        uint8_t bit_count = 0;
+        do {
+            if (!aws_byte_cursor_read_u8(to_decode, &byte)) {
+                return AWS_OP_ERR;
```
graebm: similar to another comment

you've partially modified the cursor, but then failed

source/hpack.c
```diff
@@ -0,0 +1,413 @@
...
+
+    if (to_decode->len == 0) {
+        return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
+    }
+    const uint8_t cut_bits = 8 - prefix_size;
+    const uint8_t prefix_mask = UINT8_MAX >> cut_bits;
+
+    uint8_t byte = 0;
+    if (!aws_byte_cursor_read_u8(to_decode, &byte)) {
+        return AWS_OP_ERR;
```
graebm: you need to raise an error here, aws_byte_cursor_read/write return bool and doesn't raise errors

source/hpack.c
```diff
@@ -0,0 +1,413 @@
...
+            NULL)) {
+        aws_mem_release(allocator, context->dynamic_table.buffer);
+        return NULL;
+    }
+
+    return context;
+}
+
+void aws_hpack_context_destroy(struct aws_hpack_context *context) {
+    aws_mem_release(context->allocator, context->dynamic_table.buffer);
```
graebm: trivial: `if (!context) return`

source/hpack.c
```diff
@@ -0,0 +1,413 @@
...
+    const struct aws_byte_cursor *to_encode,
+    bool huffman_encode,
+    struct aws_byte_buf *output) {
+
+    if (output->len == output->capacity) {
+        return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
+    }
+
+    /* Write the use_huffman bit */
+    *output->buffer |= huffman_encode << 7;
```
graebm: this isn't accounting for output->len

graebm: also, you're assuming the pre-existing value here is 0

source/hpack.c
```diff
@@ -86,7 +91,8 @@ int aws_hpack_decode_integer(struct aws_byte_cursor *to_decode, uint8_t prefix_s
 
     uint8_t byte = 0;
     if (!aws_byte_cursor_read_u8(to_decode, &byte)) {
-        return AWS_OP_ERR;
+        AWS_FATAL_ASSERT(false); /* Like 5 seconds ago we made sure there was data in here */
+        return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
```
graebm: Just a regular assert is fine for what would be a programming error. regular assert is compiled out. Also you've still got some error handling left over here
```
if (!read())
{
    assert(0); /* like 4 seconds ago */
}
```

or just don't bother checking that read succeeded here at all

PR #44, opened 2019-05-10
WIP websocket handler: reads incoming messages

graebm:
Websocket channel-handler reads incoming messages and passes frames to user via callbacks.

Doesn't yet deal with a channel-handler being installed to the right.


source/websocket.c
```diff
@@ -70,6 +74,13 @@ struct aws_websocket {
 
         struct aws_websocket_decoder decoder;
 
+        /* Amount to increment window after a channel message has been processed. */
+        size_t incoming_message_window_update;
+
+        /* Frame currently being decoded. */
+        struct aws_websocket_incoming_frame incoming_frame;
+        bool has_incoming_frame;
```
justinboswell: Pointer to structure? Set pointer when there's an incoming frame, clear when not, only access through pointer?

source/websocket.c
```diff
@@ -876,22 +918,163 @@ static int s_handler_process_read_message(
...
+    struct aws_websocket_incoming_frame *incoming_frame = &websocket->thread_data.incoming_frame;
+    incoming_frame->payload_length = frame->payload_length;
+    incoming_frame->opcode = frame->opcode;
+    incoming_frame->fin = frame->fin;
+    incoming_frame->rsv[0] = frame->rsv[0];
+    incoming_frame->rsv[1] = frame->rsv[1];
+    incoming_frame->rsv[2] = frame->rsv[2];
+
+    /* Invoke user cb */
+    if (websocket->on_incoming_frame_begin) {
```
justinboswell: If reading is stopped, should we even invoke this callback?

tests/test_websocket_handler.c
```diff
@@ -51,6 +60,18 @@ struct tester {
...
+    /* Frames reported via the websocket's on_incoming_frame callbacks are recorded here */
+    struct incoming_frame incoming_frames[100];
+    size_t num_incoming_frames;
+
+    /* For pushing messages downstream, to be read by websocket handler.
+     * readpush_frame is for tests to define websocket frames to be pushed downstream.
+     * An encoder is used to turn these into proper bits */
+    struct readpush_frame *readpush_frames;
+    size_t num_readpush_frames;
+    size_t readpush_frame_i;
```
justinboswell: readpush_frame_index, perhaps?

source/websocket.c
graebm: cute. I'll do that

source/websocket.c
graebm: If reading was stopped, we shouldn't get to this part of the code (wouldn't run decoder on incoming bytes, wouldn't get callback about frame-header). I added an assert to make sure that contract is being followed.

Checking is_reading_stopped is in case the user tries to close the websocket from the callback, we want that to stop the decoder from running over any further bytes.

tests/test_websocket_handler.c
graebm: ok

PR #50, opened 2019-05-15
Connection manager

bretambrose:
Initial draft of http connection manager


source/connection_manager.c
```diff
@@ -0,0 +1,876 @@
...
+     * values.  You could argue that we could use a sum of everything, but we still need the
+     * individual values for proper behavior and error checking during the ready state.  Also,
+     * a hybrid atomic/lock solution felt excessively complicated and delicate.
+     */
+    size_t external_ref_count;
+};
+
+/*
+ * Hard Requirement: Manager's lock must held somewhere in the call stack
+ */
```
graebm: just an idea: have a bool that tracks whether it's locked, and you can assert on that in your functions.
I guess you could never assert that it's unlocked, but it would be safe to assert that it was locked

```
static void s_lock(manager) {
  mutex_lock(manager->lock);
  manager->is_locked = true;
}

File: source/connection_manager.c [hunk 2]
```diff
@@ -0,0 +1,876 @@
...
+on_error:
+
+    s_aws_http_connection_manager_destroy(manager);
+
+    return NULL;
+}
+
+void aws_http_connection_manager_acquire(struct aws_http_connection_manager *manager) {
+    aws_mutex_lock(&manager->lock);
+    manager->external_ref_count += 1;
```
graebm: trivial: assert ref_count != 0

source/connection_manager.c
```diff
@@ -0,0 +1,876 @@
...
+    int result = AWS_OP_ERR;
+
+    struct aws_http_connection_acquisition *request =
+        aws_mem_acquire(manager->allocator, sizeof(struct aws_http_connection_acquisition));
+    if (request == NULL) {
+        callback(NULL, aws_last_error(), user_data);
+        return result;
+    }
+
+    AWS_LOGF_DEBUG(AWS_LS_HTTP_CONNECTION_MANAGER, "id=%p: Acquire connection", (void *)manager);
```
graebm: TRACE?

source/connection_manager.c
```diff
@@ -0,0 +1,876 @@
...
+    } else {
+        AWS_LOGF_ERROR(
+            AWS_LS_HTTP_CONNECTION_MANAGER,
+            "id=%p: Acquire connection called when manager in shut down state",
+            (void *)manager);
+
+        request->error_code = AWS_ERROR_HTTP_CONNECTION_MANAGER_INVALID_STATE_FOR_ACQUIRE;
+        aws_linked_list_push_back(&completions, &request->node);
+
+        aws_raise_error(AWS_ERROR_HTTP_CONNECTION_MANAGER_INVALID_STATE_FOR_ACQUIRE);
```
graebm: The completion callback shouldn't fire if aws_http_connection_manager_acquire_connection() returns an error code. I can't tell whether or not this is the case...

That's just the pattern we've been using. If error returned, then no completion callback.

source/connection_manager.c
```diff
@@ -0,0 +1,876 @@
...
+
+    if (should_destroy) {
+        s_aws_http_connection_manager_destroy(manager);
+    }
+}
+
+static void s_aws_http_connection_manager_build_task_set(
+    struct aws_http_connection_manager *manager,
+    struct aws_linked_list *completions,
+    size_t *new_connections) {
```
graebm: should `new_connections` get zeroed out? Its value is only manipulated from one branch

source/connection_manager.c
```diff
@@ -0,0 +1,876 @@
...
+    aws_http_on_client_connection_setup_fn *callback,
+    void *user_data) {
+
+    int result = AWS_OP_ERR;
+
+    struct aws_http_connection_acquisition *request =
+        aws_mem_acquire(manager->allocator, sizeof(struct aws_http_connection_acquisition));
+    if (request == NULL) {
+        callback(NULL, aws_last_error(), user_data);
+        return result;
```
graebm: If you're returning an error-code, don't also invoke the callback.

source/connection_manager.c
```diff
@@ -0,0 +1,876 @@
...
+
+    AWS_LOGF_INFO(AWS_LS_HTTP_CONNECTION_MANAGER, "id=%p: release", (void *)manager);
+
+    aws_mutex_lock(&manager->lock);
+
+    if (manager->external_ref_count > 0) {
+        manager->external_ref_count -= 1;
+
+        if (manager->external_ref_count == 0) {
+            AWS_LOGF_INFO(
```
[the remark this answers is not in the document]
bretambrose: Definitely, and come to think of it, adding DEBUG level logging of all ref-like counters (external ref_count, pending count, vended count, etc...) after every mutation would be ideal but we don't want to actually call log while holding the lock.  So maybe I'll add a snapshot structure that will take a copy of those and then we'll log the snapshot state post unlock.

source/connection_manager.c
bretambrose: Unfortunately, that wouldn't reliably verify that your thread is the one who took the lock though.

source/connection_manager.c
bretambrose: I'm not sure if I'm out of sync with others here, but I tend to think of trace as debug info that's just too noisy/high frequency to otherwise log at the debug level.  Since this is per connection-acquire request, it feels like it can go in debug.

source/connection_manager.c
bretambrose: We always zero it before invoking this function but that wouldn't stop us from accidentally forgetting if things changed in the future, so I'll do it there as well.

source/connection_manager.c
bretambrose: Ok, yeah, it's definitely calling the callback on several different synchronous errors.  I'll remove.

bretambrose: Actually scratch that.  Per discussion, we'll remove the return value on this function and instead always notify on the callback.

source/connection_manager.c
bretambrose: Remove return value per above

PR #68, opened 2019-05-31
Unlock mutex even if early out from connection release function

bretambrose:


tests/test_connection_manager.c
```diff
@@ -161,13 +164,12 @@ void s_release_connections(size_t count, bool close_first) {
     }
 
     if (release_count == 0) {
-        return;
+        goto release;
```
graebm: Does this silence actual errors? Or just make the test resilient to expected periodic failures?

bretambrose: It should be just resilience.  The connection requests are for real targets so they may fail occasionally, which would result in nothing to release.

PR #70, opened 2019-06-04
Fix for mem-release crash on h1 decode failure

bretambrose:
return SUCCESS on process read failures so that sender does not try and clean up message

### DIFF
--- a/source/h1_connection.c
+++ b/source/h1_connection.c
@@ -1294,7 +1294,7 @@ static int s_handler_process_read_message(
                 (void *)&connection->base);
 
             aws_raise_error(AWS_ERROR_HTTP_CONNECTION_CLOSED);
-            goto error;
+            goto shutdown;
         }
 
         if (!connection->thread_data.incoming_stream) {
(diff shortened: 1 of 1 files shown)


source/h1_connection.c
```diff
@@ -1339,16 +1339,17 @@ static int s_handler_process_read_message(
...
 
     aws_mem_release(message->allocator, message);
     return AWS_OP_SUCCESS;
-error:
+
+shutdown:
     aws_mem_release(message->allocator, message);
     s_shutdown_connection(connection, aws_last_error());
-    return AWS_OP_ERR;
+    return AWS_OP_SUCCESS;
```
justinboswell: So I see raising errors, and then returning success anyway... how does that work out?

bretambrose: Per Mike, errors are passed to shutdown, but we return sucess to indicate we consumed the message.  There's some talk of figuring out how to communicate/enforce the io message contract better (error = sender frees, success = recipient frees)

justinboswell: Perhaps add some comments to that effect, so the next person isn't confused?

PR #78, opened 2019-06-13
Support for HTTP/1.1 connection upgrade, aka switching protocols

graebm:
In preparation for websockets...

* HTTP handler detects when the connection has upgraded to a new protocol.
  * HTTP handler doesn't know or care what specific protocol was negotiated.
* Then it's up to the new protocol handler to install itself into the channel.
  * If no handler is installed when new-protocol-data starts arriving, the channel shuts down.

I could have been much more prescriptive about how this "conversion to midchannel handler" takes place, but trying to think through every use-case was overwhelming so I kept this simple and easy to change in the future.


source/h1_connection.c
```diff
@@ -704,6 +719,62 @@ static void s_stream_write_outgoing_data(struct h1_stream *stream, struct aws_io
...
+                } /* END CRITICAL SECTION */
+            }
+
+            if (has_pending_streams) {
+                AWS_LOGF_ERROR(
+                    AWS_LS_HTTP_CONNECTION,
+                    "id=%p: Cannot switch protocols while streams are pending, closing connection.",
+                    (void *)&connection->base);
+
+                error_code = AWS_ERROR_INVALID_STATE;
```
justinboswell: Why not ```AWS_ERROR_HTTP_SWITCHED_PROTOCOLS```?

source/h1_connection.c
```diff
@@ -1294,33 +1461,58 @@ static int s_handler_process_read_message(
...
+         *
+         * We need to do this check in the middle of the normal processing loop,
+         * in case the switch happens in the middle of processing a message. */
+        if (connection->thread_data.has_switched_protocols) {
+            /* Don't auto-increment read window for parts of message using the new protocol. */
+            if (message_cursor.len < connection->thread_data.incoming_message_window_update) {
+                connection->thread_data.incoming_message_window_update -= message_cursor.len;
+            } else {
+                connection->thread_data.incoming_message_window_update = 0;
+            }
```
justinboswell: It took me longer than I want to document publicly to figure this out. This is setting copy_mark to just after the current message_cursor's worth of data in the message, is that correct? Could you comment that for future us?

source/h1_connection.c
graebm: This is the error for why the connection closed.

The fact that we switched protocols isn't the error, it's that we switched protocols while other HTTP requests were still queued up.

source/h1_connection.c
graebm: yes

graebm: I changed the auto-window-update stuff in an attempt to make it less confusing

source/h1_connection.c
justinboswell: That's kinda what I thought, but I had to double check :)

tests/test_h1_client.c
```diff
@@ -1553,10 +1553,43 @@ H1_CLIENT_TEST_CASE(h1_client_close_from_on_thread_makes_not_open) {
...
-/* Send "Connection: Upgrade" request and receive "101 Switching Protocols" response. */
-static int s_switch_protocols(struct tester *tester) {
-    /* send request */
-    struct aws_http_header headers[] = {
+struct protocol_switcher {
+    /* Settings */
+    struct tester *tester;
+    size_t downstream_handler_window_size;
+    const char *data_after_upgrade_response;
+    bool dont_install_downstream_handler;
```
justinboswell: negative bools :( Perhaps ```needs_downstream_handler```, or ```should_install_downstream_handler```, or keep it flipped, but something like ```no_downstream_handler``` or ```is_terminal_handler```

graebm: ok

PR #81, opened 2019-06-16
Tests for aws_websocket_client_connect()

graebm:
* Use "function table" so all calls into external systems like HTTP can be mocked.
* Test failure everywhere the function table is used.
* Test OOM by having a "timebomb" allocator that fails after N allocations.


tests/test_websocket_bootstrap.c
```diff
@@ -0,0 +1,635 @@
...
+    bool websocket_setup_invoked;
+    int websocket_setup_error_code;
+
+    bool websocket_shutdown_invoked;
+    int websocket_shutdown_error_code;
+} s_tester;
+
+static int s_tester_init(struct aws_allocator *alloc) {
+    aws_load_error_strings();
+    aws_io_load_error_strings();
```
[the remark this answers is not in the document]
bretambrose: Taking it a little further, is it worth/ok wrapping all of the string loading into the per-library init ala aws_http_library_init that also invokes all downstream compile-time linked aws-c-* libraries as well (inits would need to be safe to call multiple times).  That way full initialization resolves to one or two direct-dependency calls without worrying about everything all the way down to common.

tests/test_websocket_bootstrap.c
```diff
@@ -0,0 +1,635 @@
...
+
+    if (s_tester.fail_at_step == BOOT_STEP_VALIDATE_RESPONSE_STATUS) {
+        *out_status = 403;
+    } else {
+        *out_status = 101;
+    }
+    return AWS_OP_SUCCESS;
+}
+
+static struct aws_websocket *s_mock_websocket_handler_new(const struct aws_websocket_handler_options *options) {
```
bretambrose: These are so much nicer than what I did previously.

tests/test_websocket_bootstrap.c
```diff
@@ -0,0 +1,635 @@
...
+    BOOT_STEP_HTTP_CONNECT_COMPLETE,
+    BOOT_STEP_REQUEST_NEW,
+    BOOT_STEP_REQUEST_COMPLETE,
+    /* If the response validation steps fail, we expect a specific error */
+    BOOT_STEP_VALIDATE_RESPONSE_STATUS = AWS_ERROR_HTTP_WEBSOCKET_UPGRADE_FAILURE,
+    BOOT_STEP_WEBSOCKET_NEW = 0x50000000, /* Back to using made-up error-codes */
+    BOOT_STEP_HTTP_SHUTDOWN,
+};
+
+/* Needs to be a static singleton so that mock functions can access it */
```
bretambrose: Isn't it the user data though?

tests/test_websocket_bootstrap.c
```diff
@@ -0,0 +1,635 @@
...
+ * Reports the reason for the failure via `out_error_code`. */
+static int s_drive_websocket_connect(int *out_error_code) {
+    bool websocket_connect_called_successfully = false;
+    bool http_connect_setup_reported_success = false;
+
+    /* Call websocket_connect() */
+    struct aws_byte_cursor uri_cursor = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("server.example.com");
+    struct aws_uri uri;
+    int err = aws_uri_init_parse(&uri, s_tester.alloc, &uri_cursor);
+    if (err) {
```
bretambrose: This feels like it should be an assert.  If there's a failure you'll only detect it "downstream" with asserts on later expected state.

tests/test_websocket_bootstrap.c
```diff
@@ -0,0 +1,635 @@
...
+    return timebomb->real_allocator->mem_acquire(timebomb->real_allocator, size);
+}
+
+static void s_timebomb_mem_release(struct aws_allocator *alloc, void *ptr) {
+    struct timebomb_impl *timebomb = alloc->impl;
+    timebomb->real_allocator->mem_release(timebomb->real_allocator, ptr);
+}
+
+/* Run connection process with an allocator that fakes running out of memory after N allocations. */
+TEST_CASE(websocket_boot_fail_because_oom) {
```
bretambrose: This is amazing.  We should do this for.... everything.

bretambrose: In fact, can we move the allocator and the test pattern (as a macro with invocation hooks maybe) up into common?

source/websocket_bootstrap.c
```diff
@@ -37,6 +37,27 @@ static const struct scheme_port s_scheme_ports[] = {
...
+    .aws_http_connection_close = aws_http_connection_close,
+    .aws_http_connection_get_channel = aws_http_connection_get_channel,
+    .aws_http_stream_new_client_request = aws_http_stream_new_client_request,
+    .aws_http_stream_release = aws_http_stream_release,
+    .aws_http_stream_get_connection = aws_http_stream_get_connection,
+    .aws_http_stream_get_incoming_response_status = aws_http_stream_get_incoming_response_status,
+    .aws_websocket_handler_new = aws_websocket_handler_new,
+};
+
+static struct aws_websocket_client_bootstrap_function_table *s_function_table = &s_default_function_table;
```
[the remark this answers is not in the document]
graebm: It's worth storing per-instance if you wanted to run a test where multiple things ran one way but then there was one outlier. I just kept this one as simple as possible for the tests I knew I'd be writing.

tests/test_websocket_bootstrap.c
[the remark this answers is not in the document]
graebm: Yes, we have a task to give every library a single init() call ala aws_http_library_init().

If we have each library init its own dependencies, we'll have to worry about diamond patterns (ex: a crt relies on mqtt and http, and each of those tries to init aws-c-io).

But I'm with you! every library_init and library_clean_up() function just needs to no-op if it's called a second time.

tests/test_websocket_bootstrap.c
```diff
@@ -0,0 +1,635 @@
...
+    AWS_FATAL_ASSERT(user_data == &s_tester);
+
+    s_tester.websocket_shutdown_invoked = true;
+    s_tester.websocket_shutdown_error_code = error_code;
+}
+
+/* Calls aws_websocket_client_connect(), and drives the async call to its conclusions.
+ * Reports the reason for the failure via `out_error_code`. */
+static int s_drive_websocket_connect(int *out_error_code) {
+    bool websocket_connect_called_successfully = false;
```
justinboswell: AWS_FATAL_ASSERT(out_error_code)

source/websocket_bootstrap.c
```diff
@@ -37,6 +37,27 @@ static const struct scheme_port s_scheme_ports[] = {
...
+
+void aws_websocket_client_bootstrap_set_function_table(struct aws_websocket_client_bootstrap_function_table *table) {
```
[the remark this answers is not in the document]
graebm: ✅

tests/test_websocket_bootstrap.c
graebm: This only fails in my out-of-memory test when 0 allocations are allowed. There's a `goto finishing_checks` in the next line

tests/test_websocket_bootstrap.c
graebm: Not all the functions being mocked have access to user_data. (ex `aws_http_connection_close(struct aws_http_connection*)`)

but they can access this static singleton

tests/test_websocket_bootstrap.c
graebm: I have a lot I need to get done this week. Can I leave that as an exercise for the *next* person who wants to try this out?

include/aws/http/private/websocket_impl.h
```diff
@@ -72,6 +76,18 @@ struct aws_websocket_handler_options {
     bool is_server;
 };
 
+struct aws_websocket_client_bootstrap_function_table {
```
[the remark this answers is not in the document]
graebm: They're not called vtables because that's for overriding the functions of the object itself. These are for overriding the functions that this object uses to call into external systems.

Not saying "function_table" is the greatest name. etables? xtables? evtable?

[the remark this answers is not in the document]
justinboswell: Hmmm, yeah, that's a fair point. I rescind judgement!

tests/test_websocket_bootstrap.c
justinboswell: That's fair. It's going to cause mayhem I bet when we try to use it, so let's be more intentional about that.

tests/test_websocket_bootstrap.c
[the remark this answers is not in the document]
graebm: I added these comments to the internal "init() for each aws-c library" task

tests/CMakeLists.txt
```diff
@@ -131,6 +131,17 @@ add_test_case(websocket_midchannel_write_huge_message)
...
 add_test_case(websocket_midchannel_read_message)
 add_test_case(websocket_midchannel_read_multiple_messages)
+add_test_case(websocket_boot_sanity_check)
+add_test_case(websocket_boot_golden_path)
+add_test_case(websocket_boot_fail_at_http_connect)
+add_test_case(websocket_boot_fail_at_http_connect_error)
+add_test_case(websocket_boot_fail_at_new_request)
+add_test_case(websocket_boot_fail_at_response_error)
+add_test_case(websocket_boot_fail_at_response_status)
+#add_test_case(websocket_boot_fail_at_response_handshake_key)
```
bretambrose: actually, what's the deal here?  maybe delete if no longer applicable

graebm: It's a test I need to write when I finish implementing the full websocket spec. I'll remove this. I have another TODO in the code about it

PR #84, opened 2019-06-19
Websocket bugfixes and little changes

graebm:
* Bugfix 1: deal aws_io_message completion callback firing before send_message() returns.
* Bugfix2: confused len and capacity.
* Let users get their hands on the websocket's channel. This came in handy when integrating with MQTT.


source/websocket.c
```diff
@@ -65,9 +65,10 @@ struct aws_websocket {
     struct aws_channel_task move_synced_data_to_thread_task;
     struct aws_channel_task shutdown_channel_task;
     struct aws_channel_task increment_read_window_task;
+    struct aws_channel_task waiting_on_payload_stream_task;
     bool is_server;
 
-    struct {
+    struct thread_data {
```
justinboswell: thread_data_t or something, differentiate the name. Old GCC will hate this IIRC

graebm: I'll just use a comment instead. I just wanted something at the top of this giant block saying what it was

PR #90, opened 2019-06-24
Revamp of elasticurl signing

bretambrose:
* Use the new aws_http_request mutable struct/API.  
* Move signer-specific details like region and service to a generic context map.

### DIFF
--- a/bin/elasticurl/main.c
+++ b/bin/elasticurl/main.c
@@ -743,5 +743,7 @@ int main(int argc, char **argv) {
         fclose(app_ctx.input_file);
     }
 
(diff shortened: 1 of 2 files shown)


bin/elasticurl/main.c
```diff
@@ -515,6 +551,14 @@ int main(int argc, char **argv) {
     app_ctx.output = stdout;
     app_ctx.verb = "GET";
     aws_mutex_init(&app_ctx.mutex);
+    aws_hash_table_init(
+        &app_ctx.signing_context,
+        allocator,
+        10,
```
justinboswell: Shouldn't this be 1 or 2? There's only a single kv pair ever stored in here, right?

PR #93, opened 2019-07-08
test aws_http_request OOM errors

graebm:
Test found an actual bug!

### DIFF
--- a/source/request_response.c
+++ b/source/request_response.c
@@ -133,15 +133,17 @@ void aws_http_request_destroy(struct aws_http_request *request) {
     aws_string_destroy(request->method);
     aws_string_destroy(request->path);
 
-    const size_t length = aws_array_list_length(&request->headers);
-    struct aws_http_header_impl *header_impl = NULL;
-    for (size_t i = 0; i < length; ++i) {
-        aws_array_list_get_at_ptr(&request->headers, (void **)&header_impl, i);
-        AWS_ASSERT(header_impl);
-        s_header_impl_clean_up(header_impl);
-    }
+    if (aws_array_list_is_valid(&request->headers)) {
+        const size_t length = aws_array_list_length(&request->headers);
+        struct aws_http_header_impl *header_impl = NULL;
+        for (size_t i = 0; i < length; ++i) {
+            aws_array_list_get_at_ptr(&request->headers, (void **)&header_impl, i);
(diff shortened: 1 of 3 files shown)


source/request_response.c
```diff
@@ -133,15 +133,17 @@ void aws_http_request_destroy(struct aws_http_request *request) {
...
     aws_string_destroy(request->path);
 
-    const size_t length = aws_array_list_length(&request->headers);
-    struct aws_http_header_impl *header_impl = NULL;
-    for (size_t i = 0; i < length; ++i) {
-        aws_array_list_get_at_ptr(&request->headers, (void **)&header_impl, i);
-        AWS_ASSERT(header_impl);
-        s_header_impl_clean_up(header_impl);
-    }
+    if (aws_array_list_is_valid(&request->headers)) {
```
bretambrose: This makes me nervous. If we need to check array lists for validity before cleaning them up (which I know I don't do all over the place) due to construction-OOM failures then if possible I'd prefer that array_list init is all-or-nothing like C++ constructors (if possible).

justinboswell: +1

tests/test_request.c
```diff
@@ -194,3 +199,78 @@ TEST_CASE(request_erase_headers) {
...
+}
+
+TEST_CASE(request_handles_oom) {
+    (void)ctx;
+    struct aws_allocator timebomb_alloc;
+    ASSERT_SUCCESS(aws_timebomb_allocator_init(&timebomb_alloc, allocator, SIZE_MAX));
+
+    bool test_succeeded = false;
+    size_t allocations_until_failure;
+    for (allocations_until_failure = 0; allocations_until_failure < 10000; ++allocations_until_failure) {
```
justinboswell: there's only like 300-ish possible allocations, is 10K necessary?

graebm: DUDE IT"S A MAGIC NUMBER THAT MEANS RIDICULOUS, GIMME A BREAK

source/request_response.c
graebm: This code-change was handled in https://github.com/awslabs/aws-c-http/pull/94

tests/test_request.c
graebm: we bail out of this loop when everything works, so we never actually hit 10000 unless the test is failing

I see your valid concern

PR #94, opened 2019-07-10
don't get_length() unless array_list_is_valid()

graebm:

### DIFF
--- a/source/request_response.c
+++ b/source/request_response.c
@@ -133,14 +133,15 @@ void aws_http_request_destroy(struct aws_http_request *request) {
     aws_string_destroy(request->method);
     aws_string_destroy(request->path);
 
-    const size_t length = aws_array_list_length(&request->headers);
-    struct aws_http_header_impl *header_impl = NULL;
-    for (size_t i = 0; i < length; ++i) {
-        aws_array_list_get_at_ptr(&request->headers, (void **)&header_impl, i);
-        AWS_ASSERT(header_impl);
-        s_header_impl_clean_up(header_impl);
+    if (aws_array_list_is_valid(&request->headers)) {
+        const size_t length = aws_array_list_length(&request->headers);
+        struct aws_http_header_impl *header_impl = NULL;
+        for (size_t i = 0; i < length; ++i) {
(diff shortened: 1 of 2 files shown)


source/request_response.c
```diff
@@ -133,14 +133,15 @@ void aws_http_request_destroy(struct aws_http_request *request) {
     aws_string_destroy(request->method);
     aws_string_destroy(request->path);
 
-    const size_t length = aws_array_list_length(&request->headers);
-    struct aws_http_header_impl *header_impl = NULL;
-    for (size_t i = 0; i < length; ++i) {
-        aws_array_list_get_at_ptr(&request->headers, (void **)&header_impl, i);
-        AWS_ASSERT(header_impl);
-        s_header_impl_clean_up(header_impl);
+    if (aws_array_list_is_valid(&request->headers)) {
```
justinboswell: This is a bit concerning. Why wouldn't this array list be valid? Do we not initialize it as part of initializing the request? I'm not sure how comfortable it feels to start sprinkling ```is_valid``` checks all over...

graebm: So, proof peeps added `AWS_PRECONDITION(aws_array_list_is_valid())` to the `aws_array_list_get_length()` fn.

The contract being: you shouldn't be able to call anything but _clean_up() on something that isn't initialized.

That means we shouldn't be calling _get_length() unless we know the array_list is initialized. We *could* make a special exception for `aws_array_list_get_length()` where it doesn't freak out if the list is uninitialized AND length is 0. We had a quick chat about it up in Seattle and and thought it was better to stick by the simple contract where _clean_up() is the only safe function.

justinboswell: That doesn't really answer the question though? Shouldn't we be initializing this array? That's the part I'm confused about.

graebm: `aws_http_request_new()` will call `aws_http_request_destroy()` to clean things up if something goes wrong mid-setup. So that's why the list may not be valid.

That's a handy pattern when your datastructure has a bunch of members whose setup can fail. This particular datastructure only has 2 points of failure (alloc, and array_list_init) so it wouldn't be the end of the world if init() had its own cleanup code instead of using the _destroy() function

PR #101, opened 2019-07-16
Proposed header changes v1.0 for proxies

bretambrose:

### DIFF
--- a/include/aws/http/connection.h
+++ b/include/aws/http/connection.h
@@ -35,6 +35,94 @@ typedef void(
 typedef void(
     aws_http_on_client_connection_shutdown_fn)(struct aws_http_connection *connection, int error_code, void *user_data);
 
+/*
+ * Connection target properties
+ */
+struct aws_http_endpoint_options {
+
+    /*
+     * Required
+     */
+    struct aws_byte_cursor schema;
+
+    /**
+     * Required.
+     */
+    struct aws_byte_cursor host_name;
+
+    /**
+     * Required.
+     */
+    uint16_t port;
+};
+
+/*
+ * All of the currently supported ways of connecting to an http proxy
+ */
+enum aws_http_proxy_auth_type {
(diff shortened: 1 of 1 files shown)


include/aws/http/connection.h
```diff
@@ -35,6 +35,94 @@ typedef void(
...
+/*
+ * Optional configuration for proxy connection authorization.  At the moment, only supports
+ * standard username/password auth via Proxy-Authorization header.
+ */
+struct aws_http_client_proxy_auth_options {
+
+    enum aws_http_proxy_auth_type type;
+
+    union {
+        struct aws_http_client_proxy_name_and_pass_options name_and_pass_options;
```
graebm: Are there other headers someone might want to add to the CONNECT request?
Would this ever be signed?
Those might be arguments for just accepting a list of headers and not doing anything fancy beyond that

include/aws/http/connection.h
```diff
@@ -35,6 +35,94 @@ typedef void(
...
+
+    /*
+     * Required
+     */
+    struct aws_byte_cursor schema;
+
+    /**
+     * Required.
+     */
+    struct aws_byte_cursor host_name;
```
graebm: It might be simpler to just use `aws_uri`?
Rather than declare a unique struct.
I'm super conflicted on this

include/aws/http/connection.h
```diff
@@ -35,6 +35,94 @@ typedef void(
...
+ * All of the currently supported ways of connecting to an http proxy
+ */
+enum aws_http_proxy_auth_type {
+    AWS_HPAT_NAME_AND_PASS
+};
+
+/*
+ * Configuration for username/password based proxy authentication
+ */
+struct aws_http_client_proxy_name_and_pass_options {
```
graebm: I would just combine all these new structs into a single `aws_http_proxy_options` struct?

PR #103, opened 2019-07-17
H1B: error codes for callbacks

graebm:
All HTTP request/response callbacks (other than on_complete) can return an error code.

In HTTP/1.1, this stops all further decoding of incoming data, ends the stream (with an error), and closes the connection (with an error).


graebm:
Whoops. shouldn't be a merge to master

PR #113, opened 2019-07-24
H1b server destroy

TingDaoK:
*Issue #, if available:*

*Description of changes:*

Server Destroy, which will fail if the IO is not updated.


source/connection.c
```diff
@@ -49,9 +50,17 @@ struct aws_http_server {
     size_t initial_window_size;
     void *user_data;
     aws_http_server_on_incoming_connection_fn *on_incoming_connection;
+    aws_http_server_on_destroy_fn *on_destroy_complete;
 
     struct aws_hash_table channel_to_connection_map;
```
TingDaoK: we do not need to protect this hash table by lock, because if the sever is shutting down, the bool value will protect from more new connections to the hash table. If not, concurrent insertions and deletions will be fine.

source/connection.c
```diff
@@ -385,8 +432,15 @@ struct aws_http_server *aws_http_server_new(const struct aws_http_server_options
     server->initial_window_size = options->initial_window_size;
     server->user_data = options->server_user_data;
     server->on_incoming_connection = options->on_incoming_connection;
-
-    int err = aws_hash_table_init(
+    server->on_destroy_complete = options->on_destroy_complete;
+    server->synced_data.is_shutting_down = false;
```
graebm: trivial/style: you don't need to change this, but we usually don't bother setting values that are 0/false at init time. Since we pretty much always grab the memory with calloc() (always zeroes out the memory). Or in cases where the memory already exists, we zero it out with AWS_STRUCT_ZERO() before setting up the fields

source/connection.c
```diff
@@ -439,38 +493,55 @@ struct aws_http_server *aws_http_server_new(const struct aws_http_server_options
     return server;
 
 error:
-    aws_http_server_destroy(server);
+    aws_http_server_release(server);
```
graebm: Call `s_http_server_clean_up(server);` directly. It's boils to the same thing as server_release(), but it's more obvious that you're just cleaning up memory here.

include/aws/http/server.h
```diff
@@ -163,14 +171,15 @@ AWS_HTTP_API
...
+ * Release the server. It will close the listening socket and all the connections existing in the server.
+ * The on_destroy_complete will be invoked when the destroy operation completes
  *
  * Note: this function should be called by either a user thread (like the main entry point, or from the event-loop the
  * server is assigned to. Otherwise a deadlock is possible. If you call this function from outside the assigned
  * event-loop, this function will block waiting on the assigned event-loop runs the close sequence in its thread.
  */
 AWS_HTTP_API
-void aws_http_server_destroy(struct aws_http_server *server);
+void aws_http_server_release(struct aws_http_server *server);
```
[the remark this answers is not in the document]
TingDaoK: Yes, the shutdown process is async now, but the aws_server_bootstrap_destroy_socket_listener() will only be called from the aws_http_server_release() function and it will only be called from the user or the server event loop, thus I think there is no problem here

source/connection.c
TingDaoK: Because the original code do not protect the hash table, I though it was fine to do w/r concurrently. I was wrong

source/connection.c
TingDaoK: I change it to clean up in different states, because the mutex lock is not safe to clean up, if the init failed, and I can find a way to check if the mutex lock init fail or not in the s_http_server_clean_up(server); function. So do the hash table.

source/connection.c
```diff
@@ -438,39 +513,60 @@ struct aws_http_server *aws_http_server_new(const struct aws_http_server_options
...
+            already_shutting_down = true;
+        } else {
+            server->synced_data.is_shutting_down = true;
+        }
+        s_server_unlock_synced_data(server);
+        /* END CRITICAL SECTION */
 
+        if (already_shutting_down) {
+            /* nothing to do */
+            AWS_LOGF_TRACE(AWS_LS_HTTP_SERVER, "id=%p: The server is already shutting down", (void *)server);
```
JonathanHenson: it might be good to add more, "Why should I care about this log message" info here.

"The service is already shutting down, not doing x"

JonathanHenson:
I meant to say, "Address Mike's feedback, and consider mine, then ship it."

include/aws/http/server.h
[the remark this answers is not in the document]
TingDaoK: I did this in http server, I guess we can also do the same thing in IO

source/connection.c
```diff
@@ -478,6 +474,7 @@ struct aws_http_server *aws_http_server_new(const struct aws_http_server_options
         goto hash_table_error;
     }
 
+    s_server_lock_synced_data(server);
```
graebm: comment `/* Protect against callbacks firing before server->socket is set */`

graebm: I have sad sad news.
While I wrote this, I wondered if this actually did protect us, and it doesn't. The accept callback accesses server->socket->... multiple times for logging before it ever acquires a lock. Let's do the following:
1) Log about a server based on `(void*)server` instead of `server->socket->address`.
2) Put another comment somewhere in the accept callback like `/* Be careful not to access server->socket until lock is acquired to avoid race conditions */`

PR #114, opened 2019-07-24
H1B: User initiates creation of server stream

graebm:
The old pattern (server creates stream, user must immediately configure it) had edge cases. For example:  Should user release() stream if they never called configure() on it? What if user tries to start using the stream the moment we hand it to them, before we've finished "installing" it into the connection's various lists?

New pattern is also weird (user must create the stream at one specific moment), but we can guarantee that, if the stream exists at all, it is fully read for action. And yes, the user must release it, they created it after all.


include/aws/http/server.h
```diff
@@ -103,10 +103,14 @@ struct aws_http_server_options {
...
-    struct aws_http_stream *stream,
-    void *user_data);
+/**
+ * Invoked at the start of an incoming request.
+ * From this callback, the user MUST call aws_http_stream_new_server_request_handler().
+ *
+ * Return AWS_OP_SUCCESS to continue processing the stream.
+ * Return AWS_OP_ERR to indicate failure and cancel the stream.
+ */
+typedef int(aws_http_on_incoming_request_fn)(struct aws_http_connection *connection, void *user_data);
```
ColdenCullen: WHAT IF THIS THING RETURNED THE THING THEY WERE SUPPOSED TO MAKE. Then you could super extra make it clear that this function super double needs to make a request handler.

PR #115, opened 2019-07-25
H1b proxy2

bretambrose:
http connections only

Does connection rerouting and request rewriting.  Ultimately I decided to use a decorate/wrapper approach rather than just dumping everything into connection.  I feel like this gives better separation and is more manageable long term as other functionality like this is added.


source/connection.c
```diff
@@ -26,6 +28,17 @@
 #    pragma warning(disable : 4204) /* non-constant aggregate initializer */
 #endif
 
+static struct aws_http_connection_system_vtable g_default_connection_system_vtable = {
+    .new_socket_channel = aws_client_bootstrap_new_socket_channel,
+    .new_tls_socket_channel = aws_client_bootstrap_new_tls_socket_channel};
+
+static const struct aws_http_connection_system_vtable *g_aws_http_connection_system_vtable_ptr =
+    &g_default_connection_system_vtable;
```
graebm: style nitpick: I like g_ too, but s_ is what we've been doing

also, just s_system_vtable would be much shorter, but that's just, like, my opinion man

source/proxy_connection.c
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    if (aws_base64_encode(&base64_source_cursor, &header_value)) {
+        goto done;
+    }
+
+    struct aws_http_header header = {.name = aws_byte_cursor_from_string(s_proxy_authorization_header_name),
+                                     .value = aws_byte_cursor_from_array(header_value.buffer, header_value.len)};
+
+    if (aws_http_message_add_header(request, header)) {
```
graebm: Is there any chance the authorization header already exists, and we should be appending to it, with like comma separated values or anything weird like that?

source/proxy_connection.c
```diff
@@ -0,0 +1,316 @@
...
+    struct aws_byte_cursor path_cursor;
+    AWS_ZERO_STRUCT(path_cursor);
+
+    if (aws_http_message_get_request_path(request, &path_cursor)) {
+        goto done;
+    }
+
+    /* Pull out the original path/query */
+    struct aws_uri uri;
+    if (aws_uri_init_parse(&uri, proxy_user_data->allocator, &path_cursor)) {
```
graebm: An OPTIONS request can have "*" for the path. What happens then?

source/proxy_connection.c
bretambrose: In that case you'd just use no authentication; I don't think it's worth checking for duplicates

source/proxy_connection.c
bretambrose: Shouldn't cause any problems.  I did a few experiments using curl through a proxy with and without '*' and there was no difference in responses given to the OPTIONS request.

include/aws/http/connection.h
```diff
@@ -35,6 +38,56 @@ typedef void(
...
+};
+
+/*
+ * Tagged union for proxy authentication options
+ */
+struct aws_http_proxy_authentication_options {
+    enum aws_http_proxy_authentication_type type;
+
+    union {
+        struct aws_http_proxy_authentication_basic_options basic_options;
```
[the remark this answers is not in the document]
bretambrose: I did almost all of this, but I kept the auth type enum.

source/proxy_connection.c
graebm: But "*" isn't a valid URI, so this parse() call will fail, so the OPTIONS request will fail to even send

include/aws/http/request_response.h
```diff
@@ -166,10 +165,8 @@ struct aws_http_request_options {
     bool manual_window_management;
 };
 
-typedef int(aws_transform_http_request_fn)(
-    struct aws_http_message *request,
-    struct aws_allocator *allocator,
-    const struct aws_hash_table *context);
+typedef int(
+    aws_http_request_transform_fn)(struct aws_http_message *request, struct aws_allocator *allocator, void *context);
```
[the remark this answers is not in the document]
bretambrose: Will fix in a different CR; need to coordinate with auth

PR #119, opened 2019-07-30
H1b tls proxy2

bretambrose:
TLS-based proxy connections

Includes some commits from Colden/Mike since the rebase got so messy that I just merged from h1b instead.

For a diff that does not include Mike/Colden commits, try:

https://github.com/awslabs/aws-c-http/compare/h1b...h1b-tls-proxy2


source/proxy_connection.c
```diff
@@ -67,6 +87,16 @@ struct aws_http_proxy_user_data *aws_http_proxy_user_data_new(
         }
     }
 
+    if (options->tls_options) {
+        /* clone tls options, but redirect user data to what we're creating */
+        user_data->tls_options = aws_mem_calloc(allocator, 1, sizeof(struct aws_tls_connection_options));
```
JonathanHenson: we gonna check OOM?

source/proxy_connection.c
```diff
@@ -67,6 +97,17 @@ struct aws_http_proxy_user_data *aws_http_proxy_user_data_new(
...
+        if (user_data->tls_options == NULL ||
+            aws_tls_connection_options_copy(user_data->tls_options, options->tls_options)) {
```
graebm: this also can fail due to OOM

PR #123, opened 2019-08-01
H1B: Async message transform callback

graebm:
We know that transforms will need to be async someday, and it will be super painful to make that change, so let's do it now.

Also, make proxy request transformation a private thing. This doesn't need to be async, and supporting generic async transformations would make the `aws_http_connection` class even more complex.


graebm:
@bretambrose I think this is the first time we've let the user insert an optional async step. I don't know of any other way to deal with it though :|

PR #126, opened 2019-08-15
Manager fixes

bretambrose:
* Fixes an issue where a manager member variable (the allocator) was accessed after the manager had been freed.
* Refactors all of the out-of-lock operations (new connections, releasing connections, completion callbacks, etc...) into a unified structure, with appropriate operations.  This centralizes the logic better.


source/connection_manager.c
```diff
@@ -365,6 +365,113 @@ static void s_aws_http_connection_manager_move_front_acquisition(
...
+ *   manager release
+ *   connection release
+ *   connection acquire
+ *   connection_setup
+ *   connection_shutdown
+ *
+ * The work unit is built under the manager's lock, but then executed
+ * outside of it.
+ */
+struct aws_connection_management_work_unit {
```
justinboswell: I'm curious, why not just do this as a function pointer, similar to how https://github.com/facebook/folly/blob/master/folly/docs/Synchronized.md ```withXLock()``` works? Are the operations such that they need to be done together?"

bretambrose: They do need to be done as a unit and it's order sensitive.  Also recall that we're building up a set of orders now (under the lock) so that we can execute them later (not under the lock) because callbacks and notifications while holding the lock are bad.

Some of the buildup is factored into a function call (s_aws_http_connection_manager_build_work_unit) and all of the apply is one as well (s_aws_http_connection_manager_execute_work_unit).  There is just some unique-per-event initialization that needs to be done as well.

justinboswell: Hmmm, after our discussion, what about ```aws_connection_manager_transaction```? Thoughts? If I read work_unit as transaction throughout this commit, everything makes sense.

bretambrose: Yeah, I like that better.

PR #130, opened 2019-08-20
Update to common/io changes, fix clean_up

ColdenCullen:

### DIFF
--- a/bin/elasticurl/main.c
+++ b/bin/elasticurl/main.c
@@ -518,10 +518,8 @@ static bool s_completion_predicate(void *arg) {
 int main(int argc, char **argv) {
     struct aws_allocator *allocator = aws_default_allocator();
 
-    aws_load_error_strings();
-    aws_common_load_log_subject_strings();
-    aws_io_load_error_strings();
-    aws_io_load_log_subject_strings();
+    aws_common_library_init();
+    aws_io_library_init();
     aws_http_library_init(allocator);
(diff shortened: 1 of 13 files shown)


bin/elasticurl/main.c
```diff
@@ -547,8 +545,8 @@ int main(int argc, char **argv) {
     AWS_ZERO_STRUCT(logger);
 
     if (app_ctx.log_level) {
-        aws_common_load_log_subject_strings();
-        aws_io_load_log_subject_strings();
+        aws_common_library_init();
+        aws_io_library_init();
```
graebm: already initialized at line 521

bin/elasticurl/main.c
```diff
@@ -518,10 +518,8 @@ static bool s_completion_predicate(void *arg) {
 int main(int argc, char **argv) {
     struct aws_allocator *allocator = aws_default_allocator();
 
-    aws_load_error_strings();
-    aws_common_load_log_subject_strings();
-    aws_io_load_error_strings();
-    aws_io_load_log_subject_strings();
+    aws_common_library_init();
+    aws_io_library_init();
     aws_http_library_init(allocator);
```
graebm: I was hoping every library_init() function would take an allocator whether it needed it or not. That makes all the init() functions consistent, and if in the future we need to do allocations it won't be a breaking API change

PR #134, opened 2019-08-29
Http/1.1 body less edge cases (head & 304)

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/http.c
```diff
@@ -55,6 +55,9 @@ static struct aws_error_info s_errors[] = {
     AWS_DEFINE_ERROR_INFO_HTTP(
         AWS_ERROR_HTTP_MISSING_BODY_STREAM,
         "Given the provided headers (ex: Content-Length), a body is expected."),
+    AWS_DEFINE_ERROR_INFO_HTTP(
+        AWS_ERROR_HTTP_INVALID_BODY_STREAM,
+        "A body stream provided, but the message do not allow body (ex: response for HEAD Request and 304 response)"),
```
graebm: ...", but the message does not allow a body (ex: "...

source/h1_decoder.c
```diff
@@ -714,6 +721,9 @@ static int s_linestate_response(struct aws_h1_decoder *decoder, struct aws_byte_
         return AWS_OP_ERR;
     }
 
+    /* RFC-7230 section 3.3 Message Body */
+    decoder->expects_no_body |= code_val == 304 || code_val == 204 || code_val / 100 == 1;
```
graebm: 304/204/1xx should handle things differently. Check the language of [rfc7230](https://tools.ietf.org/html/rfc7230#section-3.3.2):

> A server MAY send a Content-Length header field in a 304 (Not Modified) response

> A server MUST NOT send a Content-Length header field in any response with a status code of 1xx (Informational) or 204 (No Content).

So Content-Length in a 204/1xx should be an error

PS: I looked at the rules for "Transfer-Encoding" and they're the same as for "Content-Length": MAY be in HEAD or 304 response. MUST NOT be in 1xx or 204 response.

graebm: Also, add enums to `http_impl.h` for 304 and 204 and use them here

source/h1_encoder.c
```diff
@@ -55,6 +58,13 @@ static int s_scan_outgoing_headers(const struct aws_http_message *message, size_
             return AWS_OP_ERR;
         }
     }
+    if (expects_no_body) {
+        /* no body should follow, no matter what the headers are */
```
graebm: I think we need a more specific word than "expect".
Sometimes we don't expect a body, so we ignore body headers.
Other times we don't expect a body, so body headers are forbidden.
Maybe more specific terms like: "body_headers_forbidden" and "body_headers_ignored"

source/h1_decoder.c
```diff
@@ -514,6 +523,15 @@ static int s_linestate_header(struct aws_h1_decoder *decoder, struct aws_byte_cu
                 return aws_raise_error(AWS_ERROR_HTTP_PARSE);
             }
 
+            if (decoder->body_headers_forbidden) {
+                AWS_LOGF_ERROR(
+                    AWS_LS_HTTP_STREAM,
+                    "id=%p: Incoming headers for content-length received, but incoming status forbids it. This is "
+                    "illegal.",
```
graebm: nitpick: Keep the message vague,  since we might find more reasons in the future to forbid a body.
"Incoming headers for content-length received, but it is illegal for this message to have a body."

source/h1_decoder.c
```diff
@@ -537,6 +555,14 @@ static int s_linestate_header(struct aws_h1_decoder *decoder, struct aws_byte_cu
                 return aws_raise_error(AWS_ERROR_HTTP_PARSE);
             }
 
+            if (decoder->body_headers_forbidden) {
+                AWS_LOGF_ERROR(
+                    AWS_LS_HTTP_STREAM,
+                    "id=%p: Incoming headers for transfer-encoding received, but incoming status forbids it. This is "
+                    "illegal.",
```
graebm: likewise

PR #140, opened 2019-09-06
Http 100 response server side

TingDaoK:
*Issue #, if available:*

*Description of changes:*

Server side support the HTTP 100 response


source/h1_encoder.c
```diff
@@ -103,6 +103,10 @@ static void s_write_headers(struct aws_byte_buf *dst, const struct aws_http_mess
     AWS_ASSERT(wrote_all);
 }
 
+static bool s_check_info_response_status_code(int code_val) {
+    return code_val >= 100 && code_val < 200 && code_val != 101;
```
graebm: trivial: use `AWS_HTTP_HEADER_101_SWITCHING_PROTOCOLS` and add a note that we should probably handle it the same as everything else

source/h1_encoder.c
```diff
@@ -233,35 +271,40 @@ int aws_h1_encoder_message_init_from_response(
...
-    wrote_all &= aws_byte_buf_write_u8(&message->outgoing_head_buf, '\n');
-
-    s_write_headers(&message->outgoing_head_buf, response);
-
-    wrote_all &= aws_byte_buf_write_u8(&message->outgoing_head_buf, '\r');
-    wrote_all &= aws_byte_buf_write_u8(&message->outgoing_head_buf, '\n');
-    (void)wrote_all;
-    AWS_ASSERT(wrote_all);
-
+    err = s_write_response_headers_into_buffer(&message->outgoing_head_buf, response);
```
graebm: trivial: was there a reason to have two separate functions? `s_get_response_header_len()` and `s_write_response_headers_into_buffer()`? Would it be simpler to have 1 function that took a pointer to the buffer it should operate on?

TingDaoK:
too old

PR #151, opened 2019-10-22
Implement H2 decoder

ColdenCullen:
* Implement new H2 decoder vtable
* Implement the H2 decoder state machine
* NOTE: Header blocks are not implemented yet, and as such the HEADERS, CONTINUATION, and PUSH_PROMISE frames are incomplete. These are going to require changes to the HPACK API to allow for streaming decoding.


source/h2_decoder.c
```diff
@@ -15,48 +15,622 @@
...
+DEFINE_STATE(type, 1);
+DEFINE_STATE(flags, 1);
+DEFINE_STATE(stream_id, 4);
+DEFINE_STATE(padding_len, 1);
+DEFINE_STATE(padding, 0);
+
+DEFINE_STATE(priority_block, 5);
+
+/* Frame-specific states */
+DEFINE_STATE(f_data, 0);
```
justinboswell: s/f/frame/ pls

source/h2_decoder.c
```diff
@@ -15,48 +15,622 @@
...
+    AWS_ASSERT(succ);
+    (void)succ;
+
+    /* Write the priority weight */
+    uint8_t weight = 0;
+    succ = aws_byte_cursor_read_u8(input, &weight);
+    AWS_ASSERT(succ);
+    (void)succ;
+
+#if defined(AWS_H2_ENABLE_PRIORITY)
```
justinboswell: We should probably assert if someone defines this, until we actually want to support it. It's considered harmful, ATM.

tests/fuzz/fuzz_h2_decoder.c
```diff
@@ -0,0 +1,70 @@
...
+    aws_http_library_init(allocator);
+
+    /* Create the decoder */
+    struct aws_h2_decoder_params decoder_params = {
+        .alloc = allocator,
+    };
+    struct aws_h2_decoder *decoder = aws_h2_decoder_new(&decoder_params);
+
+    /* Decode whatever we got */
+    aws_h2_decode(decoder, &to_decode);
```
justinboswell: Is there any state on the decoder that we should validate, like maybe that it resets when the packet is bad?

source/h2_decoder.c
ColdenCullen: I mean, this code won't compile with it, so does that count?

source/h2_decoder.c
```diff
@@ -15,48 +15,622 @@
...
+            /* Easy case, there is no scratch and we have enough data, so just send it to the state */
+
+#if DEBUG_BUILD
+            const char *current_state = decoder->state.name;
+            const uint64_t pre_state_data_len = data->len;
+#endif
+
+            err = decoder->state.fn(decoder, data);
+
+#if DEBUG_BUILD
```
JonathanHenson: I'd love to be able to get these logging statements on a production system without having to recompile the application.

source/h2_decoder.c
```diff
@@ -15,48 +15,622 @@
...
+            (void)succ;
+
+            /* If we have the correct number of bytes, call the state */
+            if (will_finish_state) {
+#if DEBUG_BUILD
+                const char *current_state = decoder->state.name;
+#endif
+
+                struct aws_byte_cursor state_data = aws_byte_cursor_from_buf(&decoder->scratch);
+                err = decoder->state.fn(decoder, &state_data);
```
JonathanHenson: you need to memmove scratch if it was advanced and reset the length.

tests/fuzz/fuzz_h2_decoder.c
ColdenCullen: That's a good idea, but I'll have to think about it. I don't know how we would know if a fame is valid or not, but I guess I could make sure that if it returns an error, I can successfully call decode again with the rest of the data.

source/h2_decoder.c
```diff
@@ -15,48 +15,638 @@
...
+        DECODER_CALL_VTABLE(decoder, on_settings_ack);
+        s_decoder_reset_state(decoder);
+        return AWS_OP_SUCCESS;
+    }
+
+    /* Otherwise, we need to process blocks of 6 octets at a time */
+    decoder->state.bytes_required = 6;
+
+    /* Truck through the list until we run out of space */
+    while (input->len >= 6) {
```
JonathanHenson: can we give this "6" a more descriptive name?

source/h2_decoder.c
```diff
@@ -15,48 +15,642 @@
...
+
+    bool succ = aws_byte_cursor_read_be32(input, &last_stream);
+    AWS_ASSERT(succ);
+    (void)succ;
+
+    succ = aws_byte_cursor_read_be32(input, &error_code);
+    AWS_ASSERT(succ);
+    (void)succ;
+
+    struct aws_byte_cursor debug_data = aws_byte_cursor_advance(input, payload_len - 8);
```
graebm: Maybe split GOAWAY into 2 distinct states, where the 2nd one processes the Additional Debug Data similar to how we process DATA?

source/h2_decoder.c
```diff
@@ -15,48 +15,642 @@
...
+}
+static int s_state_fn_frame_goaway(struct aws_h2_decoder *decoder, struct aws_byte_cursor *input) {
+
+    const uint32_t payload_len = decoder->frame_in_progress.payload_len;
+    AWS_FATAL_ASSERT(input->len >= payload_len);
+
+    uint32_t last_stream = 0;
+    uint32_t error_code = AWS_H2_ERR_NO_ERROR;
+
+    bool succ = aws_byte_cursor_read_be32(input, &last_stream);
```
graebm: ignore high bit, right?
```
 +-+-------------------------------------------------------------+
 |R|                  Last-Stream-ID (31)                        |
 +-+-------------------------------------------------------------+
```

source/h2_decoder.c
```diff
@@ -15,48 +15,642 @@
...
+    return AWS_OP_SUCCESS;
+}
+static int s_state_fn_frame_window_update(struct aws_h2_decoder *decoder, struct aws_byte_cursor *input) {
+    (void)decoder;
+    (void)input;
+
+    AWS_FATAL_ASSERT(input->len >= 4);
+
+    uint32_t window_increment = 0;
+    bool succ = aws_byte_cursor_read_be32(input, &window_increment);
```
graebm: ignore high bit, right?
```
 +-+-------------------------------------------------------------+
 |R|              Window Size Increment (31)                     |
 +-+-------------------------------------------------------------+
```

maybe we should have a helper function

source/h2_decoder.c
```diff
@@ -15,48 +15,642 @@
...
+    }
+
+    return AWS_OP_SUCCESS;
+}
+static int s_state_fn_frame_push_promise(struct aws_h2_decoder *decoder, struct aws_byte_cursor *input) {
+    (void)decoder;
+    (void)input;
+    return AWS_OP_ERR;
+}
+static int s_state_fn_frame_ping(struct aws_h2_decoder *decoder, struct aws_byte_cursor *input) {
```
graebm: > PING frames are not associated with any individual stream. If a PING frame is received with a stream identifier field value other than 0x0, the recipient MUST respond with a connection error (Section 5.4.1) of type PROTOCOL_ERROR.

Do any other states need that check as well?
Or should we check in some common place that non-stream frames don't have stream-id set?

source/h2_decoder.c
```diff
@@ -15,48 +15,642 @@
...
+            const uint64_t pre_state_data_len = data->len;
+
+            err = decoder->state.fn(decoder, data);
+
+            const uint64_t data_processed = pre_state_data_len - data->len;
+            if (bytes_required > 0 && data_processed != bytes_required) {
+                DECODER_LOGF(
+                    DEBUG,
+                    decoder,
+                    "Decoder state %s requested %" PRIu64 " bytes of data, but only used %" PRIu64,
```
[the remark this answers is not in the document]
ColdenCullen: If this raises, it's probably a programmer error with the decoder. The states should request the exact number of bytes they need, or `0` if the number of bytes required is indeterminate. If a state completes without processing the bytes it requested, it is likely a bug with the state.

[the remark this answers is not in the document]
ColdenCullen: But yes, if `err` is non-zero it's probably not worth reporting this issue.

source/h2_decoder.c
```diff
@@ -15,48 +15,642 @@
...
+            /* Not enough in this cursor, need to read as much as possible and then come back */
+            if (bytes_to_read > data->len) {
+                bytes_to_read = data->len;
+                will_finish_state = false;
+            }
+
+            /* Read the appropriate number of bytes into scratch */
+            struct aws_byte_cursor to_read = aws_byte_cursor_advance(data, bytes_to_read);
+            bool succ = aws_byte_buf_write_from_whole_cursor(&decoder->scratch, to_read);
+            AWS_ASSERT(succ);
```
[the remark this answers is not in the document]
ColdenCullen: I like the streaming debug-data idea, I'll implement that instead.

source/h2_decoder.c
```diff
@@ -15,48 +15,642 @@
...
+                err = decoder->state.fn(decoder, &state_data);
+
+                const size_t data_processed = decoder->scratch.len - state_data.len;
+                const size_t leftover_scratch = state_data.len;
+
+                if (leftover_scratch) {
+                    /* Move whatever unprocessed scratch we have back to the start of the buffer, and update the len */
+                    memmove(decoder->scratch.buffer, decoder->scratch.buffer + data_processed, leftover_scratch);
+                    decoder->scratch.len = leftover_scratch;
+                }
```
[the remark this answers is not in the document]
ColdenCullen: This is a mitigation in case of a bug (state doesn't consume all requested data).

It could, but I can mitigate it. Fix inbound.

source/h2_decoder.c
```diff
@@ -15,48 +15,642 @@
...
+            bool succ = aws_byte_buf_write_from_whole_cursor(&decoder->scratch, to_read);
+            AWS_ASSERT(succ);
+            (void)succ;
+
+            /* If we have the correct number of bytes, call the state */
+            if (will_finish_state) {
+                const char *current_state = decoder->state.name;
+
+                struct aws_byte_cursor state_data = aws_byte_cursor_from_buf(&decoder->scratch);
+                err = decoder->state.fn(decoder, &state_data);
```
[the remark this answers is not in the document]
ColdenCullen: No, we don't want to early out, because whether it errorred or not, data was still processed, so we need to update.

source/h2_decoder.c
```diff
@@ -15,48 +15,642 @@
...
+
+    /* Read the first 3 bytes */
+    bool succ = aws_byte_cursor_read(input, length_ptr, 3);
+    AWS_ASSERT(succ);
+    (void)succ;
+
+    /* Reverse from network order */
+    payload_len = aws_ntoh24(payload_len);
+    /* Assert top byte isn't set */
+    AWS_FATAL_ASSERT((payload_len & 0xFF000000) == 0);
```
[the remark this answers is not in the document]
ColdenCullen: Hi hello, please read above where I'm only reading 3 bytes. This assert is to validate ntoh24 works doesn't set the top byte. I can add that check for frame size, but since I don't support settings yet, it'll have to be hard coded.

source/h2_decoder.c
```diff
@@ -15,48 +15,642 @@
...
+
+    AWS_FATAL_ASSERT(input->len >= 1);
+
+    bool succ = aws_byte_cursor_read_u8(input, &decoder->frame_in_progress.type);
+    AWS_ASSERT(succ);
+    (void)succ;
+
+    if (AWS_UNLIKELY(decoder->frame_in_progress.type > 0x09)) {
+        return aws_raise_error(AWS_ERROR_HTTP_PROTOCOL_ERROR);
+    }
```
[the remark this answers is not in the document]
ColdenCullen: What a stupid spec. I'm just gonna toss a TODO comment on it and we can re-visit.

source/h2_decoder.c
```diff
@@ -15,48 +15,642 @@
...
+        aws_h2_frame_type_to_str(frame_type),
+        decoder->frame_in_progress.stream_id,
+        decoder->frame_in_progress.payload_len,
+        decoder->frame_in_progress.flags);
+
+    if (decoder->frame_in_progress.flags & AWS_H2_FRAME_F_PADDED) {
+
+        /* Read padding length if necessary */
+        s_decoder_set_state(decoder, &s_state_padding_len);
+    } else if (decoder->frame_in_progress.flags & AWS_H2_FRAME_T_PRIORITY) {
```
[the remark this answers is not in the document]
ColdenCullen: It does need to be an else-if, because we don't want to reset the state twice. What I should do instead is have padding_len also check for priority.

source/h2_decoder.c
```diff
@@ -15,48 +15,642 @@
...
+    if (input->len >= padding_len) {
+        aws_byte_cursor_advance(input, padding_len);
+    } else {
+        AWS_FATAL_ASSERT(input->len <= UINT8_MAX);
+        decoder->frame_in_progress.padding_len -= (uint8_t)input->len;
+        aws_byte_cursor_advance(input, input->len);
+    }
+
+    /* Done with the frame! */
+    s_decoder_reset_state(decoder);
```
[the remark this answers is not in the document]
ColdenCullen: Yes, I realized I screwed up a bunch of these lol

source/h2_decoder.c
ColdenCullen: I can definitely do a validation pass, do you mind if I do that as a separate PR?

source/h2_decoder.c
```diff
@@ -15,48 +15,642 @@
...
+        " flags=%" PRIu8 ")",
+        aws_h2_frame_type_to_str(frame_type),
+        decoder->frame_in_progress.stream_id,
+        decoder->frame_in_progress.payload_len,
+        decoder->frame_in_progress.flags);
+
+    if (decoder->frame_in_progress.flags & AWS_H2_FRAME_F_PADDED) {
+
+        /* Read padding length if necessary */
+        s_decoder_set_state(decoder, &s_state_padding_len);
```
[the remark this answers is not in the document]
ColdenCullen: See below about doing a validation pass.

tests/fuzz/fuzz_h2_decoder.c
graebm: It would be wayyyy simpler to just say that the decoder's contract is:
If decode() ever returns an error, it is illegal to call decode() ever again

source/h2_decoder.c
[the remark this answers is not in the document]
graebm: If this is all to catch programmer error, use an ASSERT to report it, not a log statement.

source/h2_decoder.c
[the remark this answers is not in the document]
graebm: Just ASSERT. Don't write a bunch of code to silently recover from your own internal logic errors

source/h2_decoder.c
graebm: Put a TODO somewhere in the code about it

include/aws/http/private/h2_decoder.h
```diff
@@ -27,10 +27,14 @@ struct aws_h2_decoder_vtable {
 
     int (*on_rst_stream)(uint32_t stream_id, uint32_t error_code, void *userdata);
 
-    int (*on_ping)(bool ack, uint64_t opaque_data, void *userdata);
+    int (*on_custom_frame_begin)(uint32_t stream_id, uint32_t payload_length, void *userdata);
+    int (*on_custom_frame_payload)(uint32_t stream_id, const struct aws_byte_cursor *data, void *userdata);
```
graebm: I would do what the spec says and simply "ignore and discard" any unrecognized data.

It's not for people to introduce their own new custom data frames, they can use DATA for that. Any new frame type would likely be something intended to alter the state of the connection, and an external user can't do that.

source/h2_decoder.c
```diff
@@ -195,52 +201,64 @@ int aws_h2_decode(struct aws_h2_decoder *decoder, struct aws_byte_cursor *data)
...
         } else {
             /* In every other case, we have to copy to scratch */
-            uint64_t bytes_to_read = bytes_required - decoder->scratch.len;
+            size_t bytes_to_read = bytes_required - decoder->scratch.len;
             bool will_finish_state = true;
-            /* Not enough in this cursor, need to read as much as possible and then come back */
-            if (bytes_to_read > data->len) {
+            if (AWS_UNLIKELY(bytes_required == 0 && decoder->scratch.len)) {
+                /* If this is a streaming state, and there's data in scratch, just send it all in now w/o reading */
+                bytes_to_read = 0;
```
graebm: Is this possible? If this can only happen when we program our states wrong, let's turn it into an ASSERT

source/h2_decoder.c
```diff
@@ -308,6 +326,18 @@ static int s_state_fn_length(struct aws_h2_decoder *decoder, struct aws_byte_cur
...
+    /* #TODO handle the SETTINGS_MAX_FRAME_SIZE setting */
+    static const uint32_t MAX_FRAME_SIZE = 16384;
+    if (payload_len > MAX_FRAME_SIZE) {
+        DECODER_LOGF(
+            ERROR,
+            decoder,
+            "Decoder's max frame size is %" PRIu32 ", but frame of size %" PRIu32 " was received.",
+            MAX_FRAME_SIZE,
+            payload_len);
+        return aws_raise_error(AWS_ERROR_HTTP_PROTOCOL_ERROR);
```
graebm: Still, this should be `FRAME_SIZE_ERROR` not PROTOCOL

source/h2_decoder.c
```diff
@@ -419,25 +450,34 @@ static int s_state_fn_padding_len(struct aws_h2_decoder *decoder, struct aws_byt
         aws_h2_frame_type_to_str(frame_type),
         decoder->frame_in_progress.padding_len);
 
-    /* Set the state to the appropriate frame's state */
-    s_decoder_set_state(decoder, s_state_frames[frame_type]);
+    if (decoder->frame_in_progress.flags & AWS_H2_FRAME_T_PRIORITY) {
+        /* Read the stream dependency and weight if PRIORITY is set */
+        s_decoder_set_state(decoder, &s_state_priority_block);
```
graebm: Add a TODO: validate that this frame type uses PRIORITY

source/h2_decoder.c
```diff
@@ -458,11 +498,7 @@ static int s_state_fn_priority_block(struct aws_h2_decoder *decoder, struct aws_
     AWS_ASSERT(succ);
     (void)succ;
 
-#if defined(AWS_H2_ENABLE_PRIORITY)
-    priority->stream_dependency = top_bytes & s_31_bit_mask;
-    priority->stream_dependency_exclusive = top_bytes >> 31;
-    priority->weight = weight;
-#endif
+    /* #NOTE: throw priority data on the GROUND. They make us hecka vulnerable to DDoS and stuff. */
```
graebm: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-9513

source/h2_decoder.c
```diff
@@ -610,21 +674,53 @@ static int s_state_fn_frame_goaway(struct aws_h2_decoder *decoder, struct aws_by
     AWS_ASSERT(succ);
     (void)succ;
 
+    if (last_stream & s_u32_top_bit_mask) {
+        /* Top bit of stream ids is reserved */
+        DECODER_LOG(ERROR, decoder, "Received GOAWAY frame with last_stream_id with top bit set");
+        return aws_raise_error(AWS_ERROR_HTTP_PROTOCOL_ERROR);
```
graebm: > R: A reserved 1-bit field. The semantics of this bit are undefined, and the bit MUST remain unset (0x0) when sending and **MUST be ignored when receiving.**

Just mask it out. Still think it's a good idea to make a helper function

source/h2_decoder.c
```diff
@@ -633,14 +729,19 @@ static int s_state_fn_frame_window_update(struct aws_h2_decoder *decoder, struct
     AWS_ASSERT(succ);
     (void)succ;
 
+    if (window_increment & s_u32_top_bit_mask) {
+        /* Top bit of stream ids is reserved */
+        DECODER_LOG(ERROR, decoder, "Received WINDOW_UPDATE frame with window size increment with top bit set");
+        return aws_raise_error(AWS_ERROR_HTTP_PROTOCOL_ERROR);
```
graebm: The WINDOW_UPDATE docs don't say what the rules are with the reserved bit. I'm going to assume rules are the same as the reserved bit in stream-id

source/h2_decoder.c
```diff
@@ -756,24 +732,14 @@ static int s_state_fn_frame_continuation(struct aws_h2_decoder *decoder, struct
...
     } else {
-        data_to_pass = aws_byte_cursor_advance(input, remaining_length);
-        will_finish_state = true;
-    }
-    AWS_FATAL_ASSERT(data_to_pass.len <= UINT32_MAX);
-
-    DECODER_CALL_VTABLE_STREAM_ARGS(decoder, on_custom_frame_payload, &data_to_pass);
-
-    if (will_finish_state) {
+        aws_byte_cursor_advance(input, remaining_length);
```
graebm: don't we need to decrement `decoder->frame_in_progress.payload_len`?

ColdenCullen: Not here, because we're resetting the state immediately after this, but for the first case, yes, yes we should.

PR #152, opened 2019-10-23
Stat

bretambrose:
Enum range calculations and package id
Socket channel creation options


source/connection_monitor.c
```diff
@@ -0,0 +1,200 @@
...
+    return handler;
+}
+
+bool aws_http_connection_monitoring_options_is_valid(const struct aws_http_connection_monitoring_options *options) {
+    if (options == NULL) {
+        return false;
+    }
+
+    return options->minimum_throughput_failure_threshold_in_seconds > 1 &&
+           options->minimum_throughput_bytes_per_second > 0;
```
graebm: The comments warn that 1 might result in flaky behavior, but here you outright refuse to do monitoring if they use it

source/connection_monitor.c
```diff
@@ -0,0 +1,200 @@
...
+     */
+    impl->consecutive_throughput_failures = current_failure_count + 1;
+    AWS_LOGF_INFO(
+        AWS_LS_IO_CHANNEL,
+        "id=%p: Channel low throughput warning.  Currently %u consecutive failures",
+        (void *)channel,
+        impl->consecutive_throughput_failures);
+
+    if (impl->consecutive_throughput_failures < impl->options.minimum_throughput_failure_threshold_in_seconds) {
+        return;
```
graebm: Seems like it we'd get better behavior if we used the actual time elapsed between ticks.

Kind of like how games **try** to tick at 60FPS, but you get weird behavior if you write your game code assuming it's running at a constant tick-rate, instead of coding against actual `float delta`

instead of just counting 1,2,3 each time one of these intervals failed, you could sum the length of time they'd been failing for. So if you failed on the first tick, but you were only reading/writing a fraction of that time, then you wouldn't exceed the `minimum_throughput_failure_threshold_in_sec` because you'd only been failing for, say, 100ms so far.

source/connection_monitor.c
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    uint64_t bytes_per_second = 0;
+    if (aws_mul_u64_checked(total_bytes, (uint64_t)AWS_TIMESTAMP_NANOS, &bytes_per_second)) {
+        AWS_LOGF_INFO(AWS_LS_IO_CHANNEL, "id=%p: io throughput overflow calculation", (void *)channel);
+        return;
+    }
+
+    bytes_per_second /= pending_io_time_ns;
```
graebm: maybe use `double` or `float` in these calculations so we don't have to worry about integer math chopping off remainders coming up with fishy numbers

source/connection_monitor.c
```diff
@@ -0,0 +1,200 @@
...
+                break;
+        }
+    }
+
+    struct aws_channel *channel = context;
+
+    if (impl->options.minimum_throughput_bytes_per_second == 0 ||
+        impl->options.minimum_throughput_failure_threshold_in_seconds == 0) {
+        return;
+    }
```
graebm: [trivial] this could be moved earlier in the fn to avoid unnecessary work

source/connection_monitor.c
```diff
@@ -0,0 +1,200 @@
...
+
+    /*
+     * All early-out/negative execution paths reset the failure count to zero.  Keep a copy of the current count for the
+     * remaining path.
+     */
+    uint32_t current_failure_count = impl->consecutive_throughput_failures;
+    impl->consecutive_throughput_failures = 0;
+
+    uint64_t pending_io_time_ns = 0;
+    if (aws_add_u64_checked(pending_read_interval_ns, pending_write_interval_ns, &pending_io_time_ns)) {
```
graebm: I don't understand why you're adding together the read and write intervals. If there's a stream that's been sending the entire time, you'll end up with a `pending_io_time = 2 seconds`
and a `total_bytes = N sent bytes` which will get you `bytes_per_second = N/2`

I would think you'd want to report `bytes_per_second = N`

Maybe it would be better to calculate the `read_bytes_per_second` and `write_bytes_per_second` separately, and either ADD those or take the MAX of the two?

source/connection_monitor.c
graebm: An approach like this might make the issues you were seeing with 1 second thresholds go away.

bretambrose: I dunno, I'd rather force multiple samples unconditionally.  Using actual time elapsed, any single long hiccup (1 bad callback for example) will shut down the connection.  In my update I've renamed the config parameter away from a notion of time.

source/connection_monitor.c
bretambrose: Modified the name and meaning a bit.

include/aws/http/connection.h
```diff
@@ -56,6 +56,25 @@ typedef void(
...
+     * there is actual io to perform
+     */
+    uint64_t minimum_throughput_bytes_per_second;
+
+    /*
+     * number of throughput failures that may occur before the monitor shuts down the connection.  If zero, no checks
+     * will be performed.  A value of two or greater is STRONGLY recommended.  A value of one will result in
+     * timing-dependent, unreliable behavior.
+     */
+    uint32_t minimum_throughput_failure_threshold_in_seconds;
```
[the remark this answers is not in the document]
bretambrose: Reworked parameter meaning away from a notion of time.

source/connection_monitor.c
bretambrose: Done.

source/connection_monitor.c
bretambrose: That was just dumb on my part.

source/connection_monitor.c
bretambrose: Entire approach reworked

source/connection_monitor.c
```diff
@@ -0,0 +1,217 @@
...
+        }
+    }
+
+    AWS_LOGF_DEBUG(
+        AWS_LS_IO_CHANNEL,
+        "id=%p: channel throughput - %" PRIu64 " bytes per second",
+        (void *)channel,
+        bytes_per_second);
+
+    /* Check throughput only if at least one stream exists and was observed in that role pregviously */
```
JonathanHenson: typo, pregviously

source/connection_monitor.c
```diff
@@ -0,0 +1,217 @@
...
+    bool check_throughput =
+        (current_incoming_stream_id != 0 && current_incoming_stream_id == impl->last_incoming_stream_id) ||
+        (current_outgoing_stream_id != 0 && current_outgoing_stream_id == impl->last_outgoing_stream_id);
+
+    impl->last_outgoing_stream_id = current_outgoing_stream_id;
```
JonathanHenson: how does this handle parallel streams? wouldn't we need a monitor per stream in that case?

JonathanHenson (approved):

source/connection_monitor.c
bretambrose: When we update this to support H2, this particular logic would only apply to h1 streams.  H2 streams would have their own stats structure which we can configure to contain the data we need to make a similar condition (if that makes sense) or for h2 streams we could always just drop this requirement.  

The fact that the streams have a simple monotonic id mechanism could mean that just keeping track of the current min and max within a connection is sufficient to apply the same conditional filter (ie if the current min id is less than the previous max id then you know at least one stream exists from the previous sample timepoint).

source/h1_connection.c
```diff
@@ -615,6 +627,50 @@ static void s_stream_complete(struct aws_h1_stream *stream, int error_code) {
...
 }
 
+static void s_add_time_measurement_to_stats(uint64_t start_ns, uint64_t end_ns, uint64_t *output_ms) {
+    if (end_ns > start_ns) {
+        *output_ms += aws_timestamp_convert(end_ns - start_ns, AWS_TIMESTAMP_NANOS, AWS_TIMESTAMP_MILLIS, NULL);
+    }
+}
+
+static void s_set_outgoing_stream_ptr(struct h1_connection *connection, struct aws_h1_stream *next_outgoing_stream) {
+    struct aws_h1_stream *prev = connection->thread_data.outgoing_stream;
```
graebm: Just a heads-up that when we change the outgoing_stream_ptr, it doesn't necessarily mean that the stream's data is done writing to the socket. It just means we've copied the stream's last outgoing bytes to a `aws_channel_message`. So for a stream whose outgoing data all fits in 16KB, it's only going to be the "current outgoing stream" for a fleeting moment, even if the socket ends up taking minutes to actually write that full `aws_channel_message`

Maybe what we want to measure is the time from when we shipped the `aws_channel_message` down the channel, to when it finished writing (there is a callback when the write completes, but the H1 code isn't currently hooked into it)

source/connection_monitor.c
```diff
@@ -0,0 +1,222 @@
...
+            2,
+            &handler,
+            sizeof(struct aws_crt_statistics_handler),
+            &impl,
+            sizeof(struct aws_statistics_handler_http_connection_monitor_impl))) {
+        return NULL;
+    }
+
+    AWS_ZERO_STRUCT(*handler);
+    AWS_ZERO_STRUCT(*impl);
```
graebm: style/opinion: you could just embed the `aws_crt_statistics_handler` within the impl struct. Then you can just use one calloc instead of acquire_many and 2 ZERO_STRUCTS

source/connection_monitor.c
```diff
@@ -0,0 +1,222 @@
...
+    uint64_t bytes_per_second = 0;
+    uint64_t max_pending_io_interval_ms = 0;
+
+    if (pending_write_interval_ms > 0) {
+        double fractional_bytes_written_per_second =
+            (double)bytes_written * (double)AWS_TIMESTAMP_MILLIS / (double)pending_write_interval_ms;
+        if (fractional_bytes_written_per_second >= (double)UINT64_MAX) {
+            bytes_per_second = UINT64_MAX;
+        } else {
+            bytes_per_second = (uint64_t)fractional_bytes_written_per_second;
```
graebm: I think you can just do `bytes_per_second = (uint64_t)fractional_bytes_written_per_second;`
I played around in a toy program and `((uint64_t)DBL_MAX == UINT64_MAX)` was true

source/connection_monitor.c
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    struct aws_channel *channel = context;
+
+    uint64_t bytes_per_second = 0;
+    uint64_t max_pending_io_interval_ms = 0;
+
+    if (pending_write_interval_ms > 0) {
+        double fractional_bytes_written_per_second =
+            (double)bytes_written * (double)AWS_TIMESTAMP_MILLIS / (double)pending_write_interval_ms;
```
graebm: I think you should replace `AWS_TIMESTAMP_MILLIS` here with MY_REPORT_INTERVAL_MS?

`AWS_TIMESTAMP_MILLIS` happens to be 1000, and `s_get_report_interval_ms()` has `return 1000;`
but using the same constant/define in both places would be preferable.

Assuming my understanding of this math is correct

Or should we measure the ACTUAL time elapsed since the last sample and use that here, instead of a hardcoded number?

source/connection_monitor.c
```diff
@@ -0,0 +1,222 @@
...
+        max_pending_io_interval_ms = pending_write_interval_ms;
+    }
+
+    if (pending_read_interval_ms > 0) {
+        double fractional_bytes_read_per_second =
+            (double)bytes_read * (double)AWS_TIMESTAMP_MILLIS / (double)pending_read_interval_ms;
+        if (fractional_bytes_read_per_second >= (double)UINT64_MAX) {
+            bytes_per_second = UINT64_MAX;
+        } else {
+            bytes_per_second = aws_add_u64_saturating(bytes_per_second, (uint64_t)fractional_bytes_read_per_second);
```
graebm: trivial: not sure if bytes_per_second should be a SUM or a MAX of the read and write
either works for the purposes of killing a slow connection
but I don't know who else might be using stats and how they might see it

source/connection.c
```diff
@@ -627,6 +633,17 @@ static void s_client_bootstrap_on_channel_setup(
         goto error;
     }
 
+    if (aws_http_connection_monitoring_options_is_valid(&http_bootstrap->monitoring_options)) {
+        struct aws_crt_statistics_handler *http_connection_monitor =
+            aws_crt_statistics_handler_new_http_connection_monitor(
+                http_bootstrap->alloc, &http_bootstrap->monitoring_options);
```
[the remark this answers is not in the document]
bretambrose: Looks like I'm conflating invalid and not-set.  So maybe creation should check for set-and-invalid and log that as an error and continue with not-set.  Everywhere else then assumes not-set or valid.

source/h1_connection.c
```diff
@@ -615,6 +627,50 @@ static void s_stream_complete(struct aws_h1_stream *stream, int error_code) {
...
+
+    uint64_t now_ns = 0;
+    aws_channel_current_clock_time(connection->base.channel_slot->channel, &now_ns);
+    if (prev == NULL && next_outgoing_stream != NULL) {
+        /* transition from nothing to write -> something to write */
+        connection->thread_data.outgoing_stream_timestamp_ns = now_ns;
+    } else if (prev != NULL && next_outgoing_stream == NULL) {
+        /* transition from something to write -> nothing to write */
```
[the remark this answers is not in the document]
bretambrose: This logic is just "start measuring"/"stop measuring"  Transitioning streams should have no effect.

source/connection.c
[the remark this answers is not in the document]
bretambrose: Looking at it closer, I think I'll just track monitoring options by pointer and thus we can assume they're valid after creation time, if non-null.  Will have to cleanup on channel shutdown when the http bootstrap is freed.

source/h1_connection.c
```diff
@@ -1751,3 +1808,51 @@ static size_t s_handler_message_overhead(struct aws_channel_handler *handler) {
...
+    }
+
+    if (connection->thread_data.outgoing_stream) {
+        s_add_time_measurement_to_stats(
+            connection->thread_data.outgoing_stream_timestamp_ns,
+            now_ns,
+            &connection->thread_data.stats.pending_outgoing_stream_ms);
+        connection->thread_data.outgoing_stream_timestamp_ns = now_ns;
+        connection->thread_data.stats.current_outgoing_stream_id =
+            aws_http_stream_get_id(&connection->thread_data.outgoing_stream->base);
```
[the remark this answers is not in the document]
bretambrose: I blame clang format.  Was all one long line before!

include/aws/http/statistics.h
```diff
@@ -0,0 +1,63 @@
...
+
+enum aws_crt_http_statistics_category {
+    AWSCRT_STAT_CAT_HTTP1_CHANNEL = AWS_CRT_STATISTICS_CATEGORY_BEGIN_RANGE(AWS_C_HTTP_PACKAGE_ID),
+};
+
+/**
+ * A statistics struct for http handlers.  Tracks the actual amount of time that incoming and outgoing requests are
+ * waiting for their IO to complete.
+ */
+struct aws_crt_statistics_http1_channel {
```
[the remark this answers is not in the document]
bretambrose: Leaving it public allows people to write their own monitors.

source/h1_connection.c
bretambrose: That would be a precision improvement, but how complicated would its addition be?  Seems like something that could wait.

source/connection_monitor.c
bretambrose: Tbh I find that approach confusing.

source/connection_monitor.c
bretambrose: Digging further, it appears that casting a too-large double to an integer type is undefined behavior:

6.3.1.4 Real floating and integer

When a finite value of real ﬂoating type is converted to an integer type other than _Bool, the fractional part is discarded (i.e., the value is truncated toward zero). If the value of the integral part cannot be represented by the integer type, the behavior is undefined.


From http://blog.frama-c.com/index.php?post/2013/10/09/Overflow-float-integer

source/connection_monitor.c
bretambrose: We're using the actual time we were "trying" to write (with the caveat that, as you pointed out elsewhere, it's not completely accurate to the socket level).  The stats reporting interval is completely independent and not a factor in this calculation.

source/connection_monitor.c
bretambrose: Well this particular calculation result isn't consumable on its own, so nothing else can use it atm.

source/h1_connection.c
```diff
@@ -1751,3 +1808,51 @@ static size_t s_handler_message_overhead(struct aws_channel_handler *handler) {
...
+    }
+
+    if (connection->thread_data.incoming_stream) {
+        s_add_time_measurement_to_stats(
+            connection->thread_data.incoming_stream_timestamp_ns,
+            now_ns,
+            &connection->thread_data.stats.pending_incoming_stream_ms);
+        connection->thread_data.incoming_stream_timestamp_ns = now_ns;
+        connection->thread_data.stats.current_incoming_stream_id =
+            aws_http_stream_get_id(&connection->thread_data.incoming_stream->base);
```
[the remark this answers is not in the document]
bretambrose: We do that right there (in the middle of the hard-to-read part).

PR #158, opened 2019-11-01
Content-Length is the boss

graebm:
- Don't freak out if body stream is set when Content-Length is 0.
  - aws-crt-java unit tests were doing this.
- Obey Content-Length when sending body, rather than trusting end-of-stream.


source/h1_encoder.c
```diff
@@ -66,7 +66,7 @@ static int s_scan_outgoing_headers(
                 }
             } break;
             case AWS_HTTP_HEADER_TRANSFER_ENCODING:
-                has_body_headers = true;
+                /*has_body_headers = true;*/
```
justinboswell: Remove?

source/h1_encoder.c
```diff
@@ -28,12 +30,13 @@
 static int s_scan_outgoing_headers(
```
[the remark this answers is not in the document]
graebm: had the same thought.
changed

source/h1_encoder.c
graebm: sure

PR #162, opened 2019-11-18
Can make new request with existing headers

graebm:

### DIFF
--- a/include/aws/http/request_response.h
+++ b/include/aws/http/request_response.h
@@ -420,6 +420,15 @@ AWS_HTTP_API
 struct aws_http_message *aws_http_message_new_request(struct aws_allocator *allocator);
 
 /**
+ * Like aws_http_message_new_request(), but uses existing aws_http_headers instead of creating a new one.
+ * Acquires a hold on the headers, and releases it when the request is destroyed.
+ */
+AWS_HTTP_API
+struct aws_http_message *aws_http_message_new_request_use_headers(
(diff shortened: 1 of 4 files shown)


include/aws/http/request_response.h
```diff
@@ -419,6 +419,15 @@ void aws_http_headers_clear(struct aws_http_headers *headers);
 AWS_HTTP_API
 struct aws_http_message *aws_http_message_new_request(struct aws_allocator *allocator);
 
+/**
+ * Like aws_http_message_new_request(), but uses existing aws_http_headers instead of creating a new one.
+ * Acquires a hold on the headers, and releases it when the request is destroyed.
+ */
+AWS_HTTP_API
+struct aws_http_message *aws_http_message_new_request_use_headers(
```
justinboswell: Why not just have ```aws_http_message_new_request``` always take an optional ```aws_http_headers*``` and then have it work just like ```s_message_new_request_common``` does? This feels a bit...weird

graebm: because this is a backwards compatible change?

justinboswell: That's fine, but I'm not sure this isn't more awkward than fixing it, that's why I wanted to discuss it.

graebm: because I'm not 100% sure this is even a good idea, and I'd rather not mess up the other constructor which I know is fine?

PR #170, opened 2019-12-13
Rewrite HPACK dyn table to have size be in octets instead of elements

ColdenCullen:
I misread the spec the first time, whoops.

### DIFF
--- a/source/hpack.c
+++ b/source/hpack.c
@@ -12,7 +12,6 @@
  * express or implied. See the License for the specific language governing
  * permissions and limitations under the License.
  */
-
 #include <aws/http/private/hpack.h>
 
 #include <aws/http/request_response.h>
@@ -25,8 +24,12 @@
 
 /* RFC-7540 6.5.2 */
 const size_t s_hpack_dynamic_table_initial_size = 4096;
+const size_t s_hpack_dynamic_table_initial_elements = 512;
 /* TBD */
-const size_t s_hpack_dynamic_table_max_size = 4096;
+const size_t s_hpack_dynamic_table_max_size = 16 * 1024 * 1024;
+
+/* Used for growing the dynamic table buffer when it fills up */
+const float s_hpack_dynamic_table_buffer_growth_rate = 1.5F;
 
 struct aws_huffman_symbol_coder *hpack_get_coder(void);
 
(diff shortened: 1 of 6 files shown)


source/hpack.c
```diff
@@ -538,6 +524,105 @@ int aws_hpack_resize_dynamic_table(struct aws_hpack_context *context, size_t new
...
+    /* Don't move forward if no elements allowed in the dynamic table */
+    if (AWS_UNLIKELY(context->dynamic_table.max_size == 0)) {
+        return AWS_OP_SUCCESS;
+    }
+
+    const size_t header_size = aws_hpack_get_header_size(header);
+
+    /* If for whatever reason this new header is bigger than the total table size, burn everything to the ground. */
+    if (AWS_UNLIKELY(header_size > context->dynamic_table.max_size)) {
+        return s_dynamic_table_shrink(context, 0);
```
graebm: If the other side claims to have fit this into their dynamic table, which should be impossible, it seems like we're in undefined behavior town. Raise an error and fail the connection?

source/hpack.c
```diff
@@ -538,6 +524,105 @@ int aws_hpack_resize_dynamic_table(struct aws_hpack_context *context, size_t new
...
+
+int aws_hpack_resize_dynamic_table(struct aws_hpack_context *context, size_t new_max_size) {
+
+    /* Nothing to see here! */
+    if (new_max_size == context->dynamic_table.max_size) {
+        return AWS_OP_SUCCESS;
+    }
+
+    if (new_max_size > s_hpack_dynamic_table_max_size) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
```
graebm: while this is an accurate error message for this function, if this is the error that bubbles out as the reason that the connection was terminated, we might want to create a more specific error like AWS_ERROR_HTTP_INTERNAL_LIMIT_EXCEEDED

source/hpack.c
```diff
@@ -538,6 +524,105 @@ int aws_hpack_resize_dynamic_table(struct aws_hpack_context *context, size_t new
...
+    }
+
+    /* If downsizing, remove elements until we're within the new size constraints */
+    if (s_dynamic_table_shrink(context, new_max_size)) {
+        goto error;
```
graebm: maybe edit comment. sounds like you don't MEAN to call this unless we're downsizing, but we always call it because it's safe either way

PR #171, opened 2019-12-18
CI wasn't running elasticurl tests.

graebm:
Also, hide elasticurl output unless test fails

### DIFF
--- a/builder.json
+++ b/builder.json
@@ -6,6 +6,12 @@
             "_comment": "al2012 can't currently pull builder.py from github. SSL: CERTIFICATE_VERIFY_FAILED"
         }
     },
+    "targets": {
+        "android": {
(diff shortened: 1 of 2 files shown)


builder.json
```diff
@@ -6,5 +6,9 @@
     ],
     "downstream": [
         { "name": "aws-c-auth" }
+    ],
+    "test": [
+        ["ctest", ".", "--output-on-failure"],
+        ["{python}", "integration-testing/http_client_test.py", "{install_dir}/bin/elasticurl{exe}"]
```
bretambrose: does the path need to go up one level here?

graebm: Yes, good catch
using "{source_dir}/integration-testing/..." now

PR #189, opened 2020-02-27
Read HTTP/2 connection preface

graebm:
Clients need to send a magic string, then a SETTINGS frame, before anything else.
Servers need to send a SETTINGS frame before anything else.


source/h2_decoder.c
```diff
@@ -456,6 +466,19 @@ static int s_state_fn_prefix(struct aws_h2_decoder *decoder, struct aws_byte_cur
...
+            decoder->connection_preface_complete = true;
+        } else {
+            DECODER_LOG(ERROR, decoder, "First frame must be SETTINGS");
+            return aws_raise_error(AWS_ERROR_HTTP_PROTOCOL_ERROR);
+        }
+    }
+
     /* Validate the frame's stream ID.
      * Frame types generally either require a stream-id, or require that it be zero. */
     frame->stream_id &= s_31_bit_mask; /* Mask off reserved bit */
```
bretambrose: Should we fail if the high bit is set?  More generally, when the spec says "MUST NOT DO X" but doesn't explicitly define X as resulting in a (protocol/connection) error, should it be ignored or connection-terminal?

graebm: > R: A reserved 1-bit field. The semantics of this bit are undefined, and the bit MUST remain unset (0x0) when sending and **MUST be ignored when receiving**

I'll add a comment to clarify

PR #190, opened 2020-03-04
HTTP/2 queues outgoing frames

graebm:
**EXPAND LARGE DIFFS**

High-Level algorithm is this:
- `outgoing_frames_queue` is simple FIFO queue, holds all frame-types EXCEPT DATA.
- `outgoing_streams_list` is list of streams that have data to send.
- Connection writes frames from queue until it's empty, then writes data frames from the streams in a round-robin manner.

Details:
- Grab the largest possible `aws_io_message` and stuff as many frames as possible into it.
- Don't send another `aws_io_message` until the previous one finishes writing (same behavior as HTTP/1).

There's still a bunch of work to do, but this is a checkpoint.

### DIFF
--- a/include/aws/http/private/h2_stream.h
+++ b/include/aws/http/private/h2_stream.h
@@ -52,9 +52,9 @@ struct aws_h2_stream {
 
     /* Only the event-loop thread may touch this data */
     struct {
-        bool expects_continuation;
         enum aws_h2_stream_state state;
         uint64_t window_size; /* #TODO try to figure out how this actually works, and then implement it */
+        struct aws_http_message *outgoing_message;
     } thread_data;
 
     /* Any thread may touch this data, but the lock must be held */
@@ -64,18 +64,15 @@ struct aws_h2_stream {
     } synced_data;
 };
 
-struct aws_h2_stream;
-
-AWS_EXTERN_C_BEGIN
-
-AWS_HTTP_API
 const char *aws_h2_stream_state_to_str(enum aws_h2_stream_state state);
 
-AWS_HTTP_API
 struct aws_h2_stream *aws_h2_stream_new_request(
     struct aws_http_connection *client_connection,
     const struct aws_http_make_request_options *options);
 
-AWS_EXTERN_C_END
+enum aws_h2_stream_state aws_h2_stream_get_state(const struct aws_h2_stream *stream);
+
+/* Connection is ready to send frames from stream now */
+int aws_h2_stream_on_activated(struct aws_h2_stream *stream, bool *out_has_outgoing_data);
 
 #endif /* AWS_HTTP_H2_STREAM_H */
--- a/source/h2_connection.c
+++ b/source/h2_connection.c
@@ -144,6 +144,11 @@ static void s_stop(
     }
 }
 
+static void s_shutdown_due_to_write_err(struct aws_h2_connection *connection, int error_code) {
+    AWS_PRECONDITION(error_code);
+    s_stop(connection, false /*stop_reading*/, true /*stop_writing*/, true /*schedule_shutdown*/, error_code);
+}
+
 /* Common new() logic for server & client */
 static struct aws_h2_connection *s_connection_new(
     struct aws_allocator *alloc,
@@ -175,6 +180,9 @@ static struct aws_h2_connection *s_connection_new(
 
     connection->synced_data.is_open = true;
     aws_linked_list_init(&connection->synced_data.pending_stream_list);
+
+    aws_linked_list_init(&connection->thread_data.outgoing_streams_list);
+    aws_linked_list_init(&connection->thread_data.outgoing_frames_queue);
 
     if (aws_mutex_init(&connection->synced_data.lock)) {
         CONNECTION_LOGF(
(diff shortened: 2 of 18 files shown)


include/aws/http/private/h2_stream.h
```diff
@@ -64,18 +64,15 @@ struct aws_h2_stream {
...
 };
 
-struct aws_h2_stream;
-
-AWS_EXTERN_C_BEGIN
-
-AWS_HTTP_API
 const char *aws_h2_stream_state_to_str(enum aws_h2_stream_state state);
 
-AWS_HTTP_API
```
JonathanHenson: should we actively hide these from the symbol exports? or does -fvisibility=hidden do this for us?

source/h2_connection.c
```diff
@@ -256,16 +272,326 @@ static void s_handler_destroy(struct aws_channel_handler *handler) {
...
     aws_h2_decoder_destroy(connection->thread_data.decoder);
     aws_h2_frame_encoder_clean_up(&connection->thread_data.encoder);
     aws_hash_table_clean_up(&connection->thread_data.active_streams_map);
     aws_mutex_clean_up(&connection->synced_data.lock);
     aws_mem_release(connection->base.alloc, connection);
 }
 
-static void s_stream_complete(struct aws_h2_stream *stream, int error_code) {
+void aws_h2_connection_enqueue_outgoing_frame(struct aws_h2_connection *connection, struct aws_h2_frame_base *frame) {
+    /* DATA frames are not queued. We only send DATA frames when the frame queue is empty */
```
JonathanHenson: can you add this to the header file for this function declaration?

source/h2_connection.c
```diff
@@ -256,16 +272,326 @@ static void s_handler_destroy(struct aws_channel_handler *handler) {
...
+                    available_bytes);
+                aws_raise_error(AWS_ERROR_INVALID_STATE);
+                goto error;
+            }
+
+            CONNECTION_LOG(TRACE, connection, "Outgoing frames task filled message, and has more frames to send later");
+            goto done_encoding;
+        }
+
+        /* #TODO some way for frame to say it's not done yet.
```
JonathanHenson: why not track the amount written in the frame and just leave it in the queue until the entire thing is sent? This of course adds some craziness because you now have to make sure no data frames go out until you're done with this part, but that should be fine since you always prefer this queue.

source/h2_connection.c
```diff
@@ -256,16 +272,326 @@ static void s_handler_destroy(struct aws_channel_handler *handler) {
...
+        num_frames_encoded++;
+    }
+
+    /* Write as many DATA frames from outgoing_streams_list as possible.
+     * We simply round-robin through available streams, instead of using stream priority.
+     *
+     * Respecting priority is not required (RFC-7540 5.3), so we're ignoring it for now. This also keeps use safe
+     * from priority DOS attacks: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-9513
+     */
+    while (!aws_linked_list_empty(outgoing_streams_list)) {
```
JonathanHenson: and msg->msg_data.capacity - msg->msg_data.len > 0?

source/h2_stream.c
```diff
@@ -87,15 +90,97 @@ struct aws_h2_stream *aws_h2_stream_new_request(
...
+    struct aws_h2_frame_headers *headers_frame = aws_mem_calloc(alloc, 1, sizeof(struct aws_h2_frame_headers));
+    if (!headers_frame) {
+        goto error_alloc;
+    }
+
+    if (aws_h2_frame_headers_init(headers_frame, alloc)) {
+        goto error_init;
+    }
+
+    const size_t num_headers = aws_http_message_get_header_count(message);
```
JonathanHenson: how are we handling header's size being too large for the frame here?

source/h2_connection.c
graebm: there's currently no way to query the frame's total encoded length, so I can't just compare total length against the amount written. You're right though, that with this design I can split any non-data frame across `aws_io_messages` no problem.

I think my next PR will be overhauling this encoder to work the way I need it to.

source/h2_connection.c
graebm: `outgoing_streams_list` is only for streams that *currently* have data to send (there's another datastructure `active_streams_map` that contains ALL living streams). In a future PR there will be some trigger somewhere that pops streams out of this list when they're done sending data.

source/h2_stream.c
graebm: Not handling it yet. I'll add an explicit TODO.

The way we encode headers needs a complete revamp to account for max frame size, and splitting the contents across CONTINUATION frames. This will probably be the next PR.

include/aws/http/private/h2_stream.h
graebm: I don't know. Linker stuff is not my strong suit.
I know public API stuff needs AWS_HTTP_API
And private-but-called from tests needs AWS_HTTP_API

These are private, not called from tests, but called from other .c files in this project. What *is* the right thing to do here?

JonathanHenson: I'm like 99% sure on all compilers later than gcc 4, we hide all symbols not explicitly marked with the API macros

graebm: We discussed this in chat, and leaving off the exports, like I'm doing here, seems like the right approach

PR #198, opened 2020-03-11
Fix proxies connect and tests

JonathanHenson:

### DIFF
--- a/source/h1_connection.c
+++ b/source/h1_connection.c
@@ -182,6 +182,10 @@ struct h1_connection {
 
         /* If non-zero, then window_update_task is scheduled */
         size_t window_update_size;
+
+        /* for bug for bug compatibility with L7 proxies. They can send connection:close on successful connect
+         * responses. This should be ignored. */
+        bool ignore_response_connection_close_header;
     } synced_data;
 };
 
@@ -441,6 +445,9 @@ struct aws_http_stream *s_make_request(
 
     { /* BEGIN CRITICAL SECTION */
         s_h1_connection_lock_synced_data(connection);
+
+        connection->synced_data.ignore_response_connection_close_header =
+            stream->encoder_message.ignore_response_connection_close_header;
 
         if (connection->synced_data.new_stream_error_code) {
             new_stream_error_code = connection->synced_data.new_stream_error_code;
@@ -985,7 +992,16 @@ static int s_decoder_on_header(const struct aws_h1_decoded_header *header, void 
     /* RFC-7230 section 6.1.
      * "Connection: close" header signals that a connection will not persist after the current request/response */
     if (header->name == AWS_HTTP_HEADER_CONNECTION) {
-        if (aws_byte_cursor_eq_c_str(&header->value_data, "close")) {
+        bool ignore_connection_close = false;
+        {
+            /*BEGIN CRITICAL SECTION*/
+            aws_mutex_lock(&connection->synced_data.lock);
+            ignore_connection_close = connection->synced_data.ignore_response_connection_close_header;
+            connection->synced_data.ignore_response_connection_close_header = false;
+            aws_mutex_unlock(&connection->synced_data.lock);
+            /*END CRITICAL SECTION*/
+        }
+        if (!ignore_connection_close && aws_byte_cursor_eq_c_str_ignore_case(&header->value_data, "close")) {
             AWS_LOGF_TRACE(
                 AWS_LS_HTTP_STREAM,
                 "id=%p: Received 'Connection: close' header. This will be the final stream on this connection.",
--- a/source/h1_connection.c
+++ b/source/h1_connection.c
@@ -182,10 +182,6 @@ struct h1_connection {
 
         /* If non-zero, then window_update_task is scheduled */
         size_t window_update_size;
-
-        /* for bug for bug compatibility with L7 proxies. They can send connection:close on successful connect
-         * responses. This should be ignored. */
-        bool ignore_response_connection_close_header;
     } synced_data;
 };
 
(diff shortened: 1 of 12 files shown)


source/h1_connection.c
```diff
@@ -985,7 +992,16 @@ static int s_decoder_on_header(const struct aws_h1_decoded_header *header, void
     /* RFC-7230 section 6.1.
      * "Connection: close" header signals that a connection will not persist after the current request/response */
     if (header->name == AWS_HTTP_HEADER_CONNECTION) {
-        if (aws_byte_cursor_eq_c_str(&header->value_data, "close")) {
+        bool ignore_connection_close = false;
```
graebm: You can do all the voodoo right here without any extra variables or locking:

```
/* for bug for bug compatibility with L7 proxies. They can send connection:close on successful connect
 * responses. This should be ignored. */
bool ignore_connection_close =
        incoming_stream->base.request_method == AWS_HTTP_METHOD_CONNECT
        && incoming_stream->client_data
        && incoming_stream->client_data->response_status == AWS_HTTP_STATUS_200_OK;
```
Feel free to rewrite that however looks good in your eyes (and in the eyes of clang-format)

You should define AWS_HTTP_METHOD_CONNECT and AWS_HTTP_STATUS_200_OK. I'd been trying to only define enums for things we're actually acting on. There are a few steps to hooking up the method enum, just do all the same steps as the others.

source/h1_connection.c
```diff
@@ -182,6 +182,10 @@ struct h1_connection {
 
         /* If non-zero, then window_update_task is scheduled */
         size_t window_update_size;
+
+        /* for bug for bug compatibility with L7 proxies. They can send connection:close on successful connect
+         * responses. This should be ignored. */
```
graebm: Any specific documentation on this behavior that we can reference? or just the stackoverflow question?

tests/proxy_test_helper.c
```diff
@@ -265,10 +265,10 @@ int proxy_tester_create_testing_channel_connection(struct proxy_tester *tester)
     ASSERT_NOT_NULL(slot);
     ASSERT_SUCCESS(aws_channel_slot_insert_end(tester->testing_channel->channel, slot));
     ASSERT_SUCCESS(aws_channel_slot_set_handler(slot, &connection->channel_handler));
-    connection->vtable->on_channel_handler_installed(&connection->channel_handler, slot);
+    testing_channel_drain_queued_tasks(tester->testing_channel);
 
+    connection->vtable->on_channel_handler_installed(&connection->channel_handler, slot);
     tester->client_connection = connection;
-    testing_channel_drain_queued_tasks(tester->testing_channel);
```
graebm: Why did we need to move `testing_channel_drain_queued_tasks()` between installing the handler and invoking its `on_channel_handler_installed()` callback?

If you needed to drain the tasks exactly here in tests, its not going to work in the real world. The actual function that installs the http-channel-handler outside of tests invokes the callback right away

tests/test_websocket_handler.c
```diff
@@ -487,14 +489,15 @@ static int s_tester_init(struct tester *tester, struct aws_allocator *alloc) {
...
+        .initial_window_size = s_default_initial_window_size,
         .user_data = tester,
         .on_incoming_frame_begin = s_on_incoming_frame_begin,
         .on_incoming_frame_payload = s_on_incoming_frame_payload,
         .on_incoming_frame_complete = s_on_incoming_frame_complete,
         .manual_window_update = s_tester_options.manual_window_update,
     };
     tester->websocket = aws_websocket_handler_new(&ws_options);
+    testing_channel_drain_queued_tasks(&tester->testing_channel);
     ASSERT_NOT_NULL(tester->websocket);
```
graebm: trivial: could you move the `drain_queued_tasks()` after the `ASSERT_NOT_NULL(tester->websocket);`?

It's weird to put an arbitrary function between calling a function and asserting its results

tests/proxy_test_helper.c
```diff
@@ -309,7 +309,7 @@ int proxy_tester_send_connect_response(struct proxy_tester *tester) {
     if (tester->failure_type == PTFT_CONNECT_REQUEST) {
         response_string = "HTTP/1.0 401 Unauthorized\r\n\r\n";
     } else {
-        response_string = "HTTP/1.0 200 Connection established\r\n\r\n";
+        response_string = "HTTP/1.0 200 Connection established\r\nconnection: close\r\n";
```
graebm: Comment that the connection: close header is an edge case

Is it worth having one explicit test where we send this response? instead of sending it in every single test?

tests/test_websocket_handler.c
```diff
@@ -24,6 +24,8 @@
 #    pragma warning(disable : 4204) /* non-constant aggregate initializer */
 #endif
 
+static const size_t s_default_initial_window_size = 256;
```
graebm:
```
/* Use small window so that we can observe it opening in tests.
 * Channel may wait until the window is small before issuing the increment command. */
```

tests/test_h1_client.c
```diff
@@ -71,7 +71,7 @@ static int s_tester_init(struct tester *tester, struct aws_allocator *alloc) {
     struct aws_testing_channel_options test_channel_options = {.clock_fn = aws_high_res_clock_get_ticks};
     ASSERT_SUCCESS(testing_channel_init(&tester->testing_channel, alloc, &test_channel_options));
 
-    tester->connection = aws_http_connection_new_http1_1_client(alloc, SIZE_MAX);
+    tester->connection = aws_http_connection_new_http1_1_client(alloc, 256);
```
graebm:
```
/* Use small window so that we can observe it opening in tests.
 * Channel may wait until the window is small before issuing the increment command. */
```

source/h1_connection.c
JonathanHenson: I updated the comment.

tests/proxy_test_helper.c
JonathanHenson: If these tests were less tangled together in a crazy web of reused functions, I'd say sure. But as they're not, I don't think it's worth the refactor that would entail, and we just want to make sure the behavior is exercised.

Added the comment.

PR #199, opened 2020-03-12
Allow s2n to auto-detect Post Quantum Assembly support

alexw91:
*Issue #, if available:* N/A

*Description of changes:*
s2n's CMake build can now [auto-detect if Post Quantum Assembly is supported](https://github.com/awslabs/s2n/blob/master/CMakeLists.txt#L116) by the Platform/Compiler combination. There's no longer a need to force it to be turned off. 


alexw91:
Looks like this change was included in a different Pull Request: 
 - https://github.com/awslabs/aws-c-http/pull/198/files#diff-bef9937f80b5ebe127d7e131380eb32eL21

PR #200, opened 2020-03-14
Streams now have to be activated. This works around a race condition …

JonathanHenson:
…where the request kicks off before the stream container itself has had a chance to be seated by the user.


source/h1_stream.c
```diff
@@ -131,6 +126,10 @@ struct aws_h1_stream *aws_h1_stream_new_request_handler(const struct aws_http_re
         return NULL;
     }
 
+    /* big assumption, in server mode, the following code can only run from the event-loop thread so don't worry
```
justinboswell: Should we assert on this? Feels like we should?

JonathanHenson: what would I assert on?

JonathanHenson: the other good news for Http 1.1. nothing bad can happen on a 32 bit integer that's only incrementing.

source/connection.c
```diff
@@ -845,13 +845,16 @@ static const uint32_t MAX_STREAM_ID = UINT32_MAX >> 1;
 
 uint32_t aws_http_connection_get_next_stream_id(struct aws_http_connection *connection) {
 
-    uint32_t next_id = (uint32_t)aws_atomic_fetch_add(&connection->next_stream_id, 2);
+    uint32_t next_id = connection->next_stream_id;
+
     /* If next fetch would overflow next_stream_id, set it to 0 */
```
graebm: this comment should be deleted

source/h1_connection.c
```diff
@@ -437,21 +473,12 @@ struct aws_http_stream *s_make_request(
...
-        } else {
-            aws_linked_list_push_back(&connection->synced_data.new_client_stream_list, &stream->node);
-            if (!connection->synced_data.is_outgoing_stream_task_active) {
-                connection->synced_data.is_outgoing_stream_task_active = true;
-                should_schedule_task = true;
-            }
         }
-
         s_h1_connection_unlock_synced_data(connection);
     } /* END CRITICAL SECTION */
```
graebm: Maybe we should move this whole critical section into `activate()`?
That way we only grab the lock once
I could see an argument that you'd rather see a "no new streams allowed" exception sooner than later
I'm really not sure on this one, just bringing it up.

source/h1_stream.c
graebm: Actually, it's not an assumption at all.
This function **can** only be called on the event-loop thread 🥳
The function that calls this one h1_connection.c raises an error if that's not the case.

You could assert again here to drive home the point, or modify the comment to state this as fact

source/h1_connection.c
```diff
@@ -418,6 +418,42 @@ static void s_connection_update_window(struct aws_http_connection *connection_ba
...
+    struct aws_h1_stream *h1_stream = AWS_CONTAINER_OF(stream, struct aws_h1_stream, base);
+
+    struct aws_http_connection *base_connection = stream->owning_connection;
+    struct h1_connection *connection = AWS_CONTAINER_OF(base_connection, struct h1_connection, base);
+
+    bool should_schedule_task = false;
+
+    { /* BEGIN CRITICAL SECTION */
+        s_h1_connection_lock_synced_data(connection);
+        stream->id = aws_http_connection_get_next_stream_id(base_connection);
```
graebm: Check to make sure stream->id isn't already set.
Yes, it's a programmer error, but `.activate()` is going to be public all the way out the the CRTs and they're often not compiled with our asserts turned on

graebm: I guess this could just no-op if it's already activated. Ain't no harm done

source/h1_connection.c
JonathanHenson: I'm not sure on this one either, but I suspect a fast fail if there's a connection error (such as a connection: close header received), we want this to fail sooner rather than later.

PR #201, opened 2020-03-16
Add Http Status Code and Header Utils

singku-china:
*Issue #, if available:*

*Description of changes:*


include/aws/http/request_response.h
```diff
@@ -414,6 +414,18 @@ int aws_http_headers_get(
...
+ * Test if header name exists or not in headers
+ */
+AWS_HTTP_API
+bool aws_http_has_header_in_headers(const struct aws_http_headers *headers, struct aws_byte_cursor name);
+
+/**
+ * Test if header name exists or not in http message
+ */
+AWS_HTTP_API
+bool aws_http_has_header_in_http_message(struct aws_http_message *message, struct aws_byte_cursor name);
```
graebm: Let's not bother adding this function to `aws_http_message`.
Someone can always aws_http_message_get_headers() and then operate on that.

The reason there's overlap between `aws_http_message` and `aws_http_headers` is that the "message" class existed first, and the "headers" class came later, but we kept the existing functions on the "message" class for API compatibility.

include/aws/http/request_response.h
```diff
@@ -414,6 +414,18 @@ int aws_http_headers_get(
     struct aws_byte_cursor name,
     struct aws_byte_cursor *out_value);
 
+/**
+ * Test if header name exists or not in headers
+ */
+AWS_HTTP_API
+bool aws_http_has_header_in_headers(const struct aws_http_headers *headers, struct aws_byte_cursor name);
```
graebm: naming suggestion: `aws_http_headers_has()`

fits in with the existing `aws_http_headers_add()` and `aws_http_headers_get()`

include/aws/http/status_code.h
```diff
@@ -0,0 +1,107 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+/*
+ * Define most of the http response codes we probably will use.
+ * https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
+ */
+enum http_status_code {
```
graebm: There's already a status-code enum defined in `private/http_impl.h`
```
/**
 * Status codes that affect internal processing.
 * This is NOT a definitive list of codes.
 */
enum aws_http_status {
    AWS_HTTP_STATUS_UNKNOWN = -1, /* Invalid status code. Not using 0 because it's technically a legal value */
    AWS_HTTP_STATUS_100_CONTINUE = 100,
    AWS_HTTP_STATUS_101_SWITCHING_PROTOCOLS = 101,
    AWS_HTTP_STATUS_204_NO_CONTENT = 204,
    AWS_HTTP_STATUS_304_NOT_MODIFIED = 304,
};
```
Once upon a time this was a public enum, but I made it private to prevent accidentally building an API where users are required to use the enums, since any numeric value < 1000 is technically allowed. I don't have a huge issue with making it public again, just explaining why it was hidden away.

Couple of points:
1) I'd rather stick to the [official](https://www.iana.org/assignments/http-status-codes/http-status-codes.txt) status codes, rather than this wikipedia list which includes stuff like codes specific to Shopify
2) maybe it's just me? but I liked including the number value in the name of the enum values. `AWS_HTTP_STATUS_100_CONTINUE` vs `AWS_HTTP_STATUS_CONTINUE`?

singku-china: I like the second idea. I had the hesitation when changing all the numbers to names, which makes the numbers implicit.

include/aws/http/status_code.h
```diff
@@ -0,0 +1,92 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+/*
+ * Define most of the http response codes we probably will use.
+ * https://www.iana.org/assignments/http-status-codes/http-status-codes.txt
+ * This is NOT a definitive list of codes.
+ */
+enum http_status_code {
```
graebm: needs to start with "aws_"

PR #202, opened 2020-03-17
Encoder revamp

graebm:
### Summary:
The existing encoder was **not** streaming. It would error if it didn't have enough space, but there was no way to know how much space it would need. We couldn't just try to re-encode the frame again when more space was available, because header-encoding mutates the HPACK state.

Now we have a "partially" streaming encoder. Most frames are pre-encoded, and we stream them into the available buffers. Headers and Data frames, which may need to split their payload across multiple frames, know how to do so.

### Details
- `aws_h2_frame` is a proper base class:
  - Previously, each specific frame type had its own init()/encode()/clean_up().  Now it's heap allocated and has vtable for encode()/destroy()
- Simple frame types (most) make use of the `aws_h2_frame_prebuilt` type.
  - The entire frame is pre-encoded.
  - We stream the pre-encoded frame into the available space of an aws_io_message.
- Remove "continuation" `aws_h2_frame` type.
  - Instead, the "headers" and "push promise" `aws_h2_frame` types will split their payload across CONTINUATION frames if necessary during encoding.
- Remove "data" `aws_h2_frame` type.
  - Instead, the encoder has a special function that takes a body-stream and writes a DATA frame immediately.
  - There are lots of differences in how the connection deals with DATA vs every other frame type, so it didn't make sense for it to be a proper `aws_h2_frame` type.
- HPACK encoding will grow the output buffer if necessary.
  - This was the simpler than building a fully "streaming" encoder.
  - HEADERS frame encoding is therefore 2 stage.
    - First, do HPACK encoding to separate buffer.
    - Then, split that payload across multiple HEADERS/CONTINUATION frames as necessary.
- Moved logic for encoding the header-block into hpack.c. Previously a bunch of that logic was in h2_frames.c. All this code was changing anyway.


source/h2_frames.c
```diff
@@ -332,714 +269,900 @@ void aws_h2_frame_encoder_clean_up(struct aws_h2_frame_encoder *encoder) {
...
+
+static int s_frame_prebuilt_encode(
+    struct aws_h2_frame *frame_base,
+    struct aws_h2_frame_encoder *encoder,
+    struct aws_byte_buf *output,
+    bool *complete) {
+
+    struct aws_h2_frame_prebuilt *frame = AWS_CONTAINER_OF(frame_base, struct aws_h2_frame_prebuilt, base);
+
+    if (frame->send_progress.len == frame->encoded_buf.len) {
```
justinboswell: I'm a little lost, intuitively I thought "wouldn't these both be 0?", but then down below this branch send_progress.len is non-zero, so I don't understand, maybe a comment would help?

I'm sure when it's explained I'm gonna go "OH RIGHT"

source/h2_frames.c
```diff
@@ -332,714 +269,900 @@ void aws_h2_frame_encoder_clean_up(struct aws_h2_frame_encoder *encoder) {
...
+struct aws_h2_frame *aws_h2_frame_new_goaway(
+    struct aws_allocator *allocator,
+    uint32_t last_stream_id,
+    uint32_t error_code,
+    struct aws_byte_cursor debug_data) {
+
+    /* If debug_data is too long, don't sent it.
+     * It's more important that the GOAWAY frame gets sent. */
+    const size_t debug_data_max = s_prebuilt_payload_max() - s_frame_goaway_length_min;
+    if (debug_data.len > debug_data_max) {
```
justinboswell: Would truncation be reasonable here? What would this payload look like?

source/hpack.c
```diff
@@ -47,71 +49,82 @@ static uint8_t s_masked_right_bits_u8(uint8_t num_masked_bits) {
...
-        return num_bytes;
+    /* Capacity must grow to at least this size */
+    size_t required_capacity;
+    if (aws_add_size_checked(output->len, required_space, &required_capacity)) {
+        return AWS_OP_ERR;
     }
+
+    /* Prefer to double capacity, but if that's not enough grow to exactly required_capacity */
+    size_t double_capacity = aws_add_size_saturating(output->capacity, output->capacity);
+    size_t reserve = required_capacity > double_capacity ? required_capacity : double_capacity;
```
justinboswell: aws_size_max()?

source/hpack.c
```diff
@@ -47,71 +49,82 @@ static uint8_t s_masked_right_bits_u8(uint8_t num_masked_bits) {
...
 
-    const struct aws_byte_buf output_backup = *output;
-
-    if (output->len == output->capacity) {
-        return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
-    }
     const uint8_t prefix_mask = s_masked_right_bits_u8(prefix_size);
+    AWS_ASSERT((starting_bits & prefix_mask) == 0);
+
+    const size_t output_len_backup = output->len;
```
justinboswell: Nitpick: ```original_len```?

source/hpack.c
```diff
@@ -611,6 +632,7 @@ int aws_hpack_insert_header(struct aws_hpack_context *context, const struct aws_
 
     /* If for whatever reason this new header is bigger than the total table size, burn everything to the ground. */
     if (AWS_UNLIKELY(header_size > context->dynamic_table.max_size)) {
+        /* #TODO handle this. It's not an error. It should simply result in an empty table RFC-7541 4.4 */
```
justinboswell: Is it hard to clear the table? What makes this a TODO?

source/hpack.c
```diff
@@ -782,79 +804,90 @@ int aws_hpack_decode_integer(
...
-            goto error;
+    /* Encode string data */
+    if (str_length > 0) {
+        if (use_huffman) {
+            /* Huffman encoder doesn't grow buffer, so we ensure it's big enough here */
+            if (s_ensure_space(output, str_length)) {
+                goto error;
+            }
+
+            if (aws_huffman_encode(&context->encoder, &to_encode, output)) {
```
justinboswell: Didn't you just add a huffman_encode that grows dynamically? Am I mis-remembering?

source/hpack.c
```diff
@@ -1174,3 +1207,178 @@ int aws_hpack_decode(
...
+ */
+static const uint8_t s_hpack_entry_starting_bit_pattern[AWS_HPACK_ENTRY_TYPE_COUNT] = {
+    [AWS_HPACK_ENTRY_INDEXED_HEADER_FIELD] = 1 << 7,
+    [AWS_HPACK_ENTRY_LITERAL_HEADER_FIELD_WITH_INCREMENTAL_INDEXING] = 1 << 6,
+    [AWS_HPACK_ENTRY_DYNAMIC_TABLE_RESIZE] = 1 << 5,
+    [AWS_HPACK_ENTRY_LITERAL_HEADER_FIELD_NEVER_INDEXED] = 1 << 4,
+    [AWS_HPACK_ENTRY_LITERAL_HEADER_FIELD_WITHOUT_INDEXING] = 0 << 4,
+};
+
+static const uint8_t s_hpack_entry_num_prefix_bits[AWS_HPACK_ENTRY_TYPE_COUNT] = {
```
justinboswell: Nitpick: If you declared this array first, you could use it to init the other array and not have the values twice :sunglasses:

source/h2_frames.c
```diff
@@ -332,714 +256,829 @@ void aws_h2_frame_encoder_clean_up(struct aws_h2_frame_encoder *encoder) {
...
 
-    return AWS_OP_SUCCESS;
-
-write_error:
-    output->len = output_init_len;
-    return aws_raise_error(AWS_H2_ERR_FRAME_SIZE_ERROR);
+    /*
+     * Write in the other parts of the frame.
+     */
+    bool all_wrote = true;
```
JonathanHenson: I'm ashamed this even bothers me.... but

bool all_written

source/h2_frames.c
```diff
@@ -332,714 +256,829 @@ void aws_h2_frame_encoder_clean_up(struct aws_h2_frame_encoder *encoder) {
...
-
-    /* Write the priority settings */
-    if (s_frame_priority_settings_encode(&frame->priority, output)) {
-        goto write_error;
-    }
+handle_waiting_for_more_space:
+    ENCODER_LOGF(TRACE, encoder, "Insufficient space to encode DATA for stream %" PRIu32 " right now", stream_id);
+    return AWS_OP_SUCCESS;
 
+handle_nothing_to_send_right_now:
```
JonathanHenson: did we ever figure out how we're going to handle this? Or are we just going to keep asking for data on ever event loop tick?

justinboswell:
Pulling out all those checks is NIIIIICE

source/h2_frames.c
```diff
@@ -332,714 +256,829 @@ void aws_h2_frame_encoder_clean_up(struct aws_h2_frame_encoder *encoder) {
...
+/* Represents a HEADERS or PUSH_PROMISE frame (followed by zero or more CONTINUATION frames) */
+struct aws_h2_frame_headers {
+    struct aws_h2_frame base;
+
+    /* Common data */
+    const struct aws_http_headers *headers;
+    uint8_t pad_length; /* Set to 0 to disable AWS_H2_FRAME_F_PADDED */
+
+    /* HEADERS-only data */
+    bool end_stream;   /* AWS_H2_FRAME_F_END_STREAM */
```
JonathanHenson: trivial, maybe take struct packing into account here and group the 1 byte sized members into groups of 2 or 4, or move them to the end

source/h2_frames.c
```diff
@@ -332,714 +256,829 @@ void aws_h2_frame_encoder_clean_up(struct aws_h2_frame_encoder *encoder) {
...
-}
-void aws_h2_frame_push_promise_clean_up(struct aws_h2_frame_push_promise *frame) {
-    AWS_PRECONDITION(frame);
+    const size_t fragment_len = aws_min_size(max_fragment, frame->header_block_cursor.len);
+    if (fragment_len == frame->header_block_cursor.len) {
+        /* This will finish the header-block */
+        flags |= AWS_H2_FRAME_F_END_HEADERS;
+    } else {
+        /* If we're not finishing the header-block, is it even worth trying to send this frame now? */
+        const size_t even_worth_sending_threshold = s_frame_prefix_length + payload_overhead;
```
JonathanHenson: we could probably make this a bit smarter in the future. Currently this would just be, do we at least have enough space to send a valid but empty headers frame?

source/h2_frames.c
```diff
@@ -332,714 +256,829 @@ void aws_h2_frame_encoder_clean_up(struct aws_h2_frame_encoder *encoder) {
...
+        s_frame_priority_settings_encode(priority_settings, output);
     }
 
-    /* Write pad length */
-    if (frame->pad_length) {
-        aws_byte_buf_write_u8(output, frame->pad_length);
+    /* Write promised stream ID */
+    if (promised_stream_id) {
+        AWS_ASSERT(frame_type == AWS_H2_FRAME_T_PUSH_PROMISE);
+        all_wrote &= aws_byte_buf_write_be32(output, *promised_stream_id & s_31_bit_mask);
```
JonathanHenson: if you've already validated the stream_id do you actually need the mask?

source/h2_frames.c
```diff
@@ -332,714 +256,829 @@ void aws_h2_frame_encoder_clean_up(struct aws_h2_frame_encoder *encoder) {
...
-        goto write_error;
-    }
-    if (!aws_byte_buf_write_from_whole_cursor(output, frame->debug_data)) {
-        goto write_error;
-    }
+    bool all_wrote = true;
 
-    return AWS_OP_SUCCESS;
+    /* Write as much of the pre-encoded frame as will fit */
+    size_t chunk_len = aws_min_size(frame->send_progress.len, output->capacity - output->len);
```
JonathanHenson: why isn't this ALWAYS zero? I think Justin and I had the same question... maybe some comments on this somewhere for future posterity?

source/h2_frames.c
JonathanHenson: I think truncation is probably bad. Imagine owning the xml or json parser on the other side?

graebm: Users are free to send whatever they want as the debug data.

Henso and I had a big chat about how to handle debug-data being too long.
We decided it shouldn't be a show-stopping error, it's more important that the GOAWAY frame gets sent.

We were less certain about truncating vs sending nothing when it's too long. Someone might never see their data get too long until the magic day that it does. So do we send partial data, or none at all? Partial XML or JSON would turn into a parse error, but a truncated log ... it might not be clear that it's been truncated and just be misleading. Options I can think of:
1) send no debug-data
1) truncate debug-data
1) truncate and write "TRUNCATED" as the last 9 bytes?

THoughts? We can always change this later. It's not exposed all the way out to a public API yet

source/hpack.c
graebm: YES
written before I added the min/max functions

I've only had them a few days and I love them to death

source/hpack.c
graebm: I wasn't touching the table code in this PR
I just traced my way in here and got spooked, so leaving a todo

source/h2_frames.c
graebm: Jonathan chimed in below and voted again for no data

source/hpack.c
graebm: I modified huffman DECODE to grow dynamically.
Needed that because you don't know how much space the compressed data will expand to unless you run through it all twice, but it's streaming in so you don't have access to it all up front, so it's basically impossible.

For encoding, we have access to all the data before we encode it, so we can just figure out exactly how much space it will need ahead of time.

I actually started modifying the huffman encoder to grow dynamically, then realized I didn't need to and reverted it

source/hpack.c
graebm: actually, you can't when they're global static const 🤷‍♀️
> initializer element is not a compile-time constant`

include/aws/http/private/h2_frames.h
```diff
@@ -94,284 +105,165 @@ struct aws_h2_frame_priority_settings {
...
-void aws_h2_frame_goaway_clean_up(struct aws_h2_frame_goaway *frame);
-AWS_HTTP_API
-int aws_h2_frame_goaway_encode(
-    struct aws_h2_frame_goaway *frame,
-    struct aws_h2_frame_encoder *encoder,
-    struct aws_byte_buf *output);
+struct aws_h2_frame *aws_h2_frame_new_ping(
+    struct aws_allocator *allocator,
+    bool ack,
+    const uint8_t opaque_data[AWS_H2_PING_DATA_SIZE]);
```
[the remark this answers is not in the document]
graebm: It needs to be exactly 8 bytes.
I thought this was a cool way to enforce that.
This isn't a public API anyway

include/aws/http/private/hpack.h
```diff
@@ -43,6 +43,20 @@ struct aws_hpack_decode_result {
...
+/**
+ * Controls whether non-indexed strings will use Huffman encoding.
+ * In SMALLEST mode, strings will only be sent with Huffman encoding if it makes them smaller.
+ *
+ * Note: This does not control compression via "indexing",
+ * for that, see `aws_http_header_compression`.
+ * This only controls how string values are encoded when they're not already in a table.
+ */
+enum aws_hpack_huffman_mode {
+    AWS_HPACK_HUFFMAN_SMALLEST,
```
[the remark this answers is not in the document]
graebm: Yeah, the ALWAYS option is just in there so we can test against some of the samples in [RFC-7541](https://httpwg.org/specs/rfc7541.html#request.examples.with.huffman.coding)

include/aws/http/private/hpack.h
```diff
@@ -75,6 +89,16 @@ int aws_hpack_decode(
     struct aws_byte_cursor *to_decode,
     struct aws_hpack_decode_result *result);
 
+/**
+ * Encode header-block into the output.
+ * This function will mutate the hpack context, so an error means the context can no longer be used.
```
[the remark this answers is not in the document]
graebm: yes
the connection is 2 layers of abstraction away so not mentioning it
oh HTTP/2 you so complicated

source/h2_frames.c
```diff
@@ -20,23 +20,65 @@
...
 
 const struct aws_byte_cursor aws_h2_connection_preface_client_string =
     AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n");
 
+/* Initial values and bounds are from RFC-7540 6.5.2 */
+const uint32_t aws_h2_settings_initial[AWS_H2_SETTINGS_END_RANGE] = {
+    [AWS_H2_SETTINGS_HEADER_TABLE_SIZE] = 4096,
+    [AWS_H2_SETTINGS_ENABLE_PUSH] = 1,
+    [AWS_H2_SETTINGS_MAX_CONCURRENT_STREAMS] = UINT32_MAX, /* "Initially there is no limit to this value" */
+    [AWS_H2_SETTINGS_INITIAL_WINDOW_SIZE] = 65535,
```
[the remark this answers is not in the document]
graebm: That's what I was doing? you'd access them like so:
`aws_h2_settings_initial[AWS_H2_SETTINGS_HEADER_TABLE_SIZE]`

instead of:
`AWS_H2_HEADER_TABLE_SIZE_INITIAL`

I thought it would be more "programmable" to have them in an array.
I'll change it in the future if it's just a pain to read

source/h2_frames.c
```diff
@@ -20,23 +20,65 @@
...
+#define ENCODER_LOG(level, encoder, text) ENCODER_LOGF(level, encoder, "%s", text)
 
 const struct aws_byte_cursor aws_h2_connection_preface_client_string =
     AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n");
 
+/* Initial values and bounds are from RFC-7540 6.5.2 */
+const uint32_t aws_h2_settings_initial[AWS_H2_SETTINGS_END_RANGE] = {
+    [AWS_H2_SETTINGS_HEADER_TABLE_SIZE] = 4096,
+    [AWS_H2_SETTINGS_ENABLE_PUSH] = 1,
+    [AWS_H2_SETTINGS_MAX_CONCURRENT_STREAMS] = UINT32_MAX, /* "Initially there is no limit to this value" */
```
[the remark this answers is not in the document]
graebm: this is the maximum number of CONCURRENT streams, not maximum stream ID.
[Spec](https://httpwg.org/specs/rfc7540.html#SettingFormat) doesn't specify a max for this setting, and it's sent on the wire as an "unsigned 32-bit value"

source/h2_frames.c
```diff
@@ -20,23 +20,65 @@
...
+    [AWS_H2_SETTINGS_INITIAL_WINDOW_SIZE] = 65535,
+    [AWS_H2_SETTINGS_MAX_FRAME_SIZE] = 16384,
+    [AWS_H2_SETTINGS_MAX_HEADER_LIST_SIZE] = UINT32_MAX, /* "The initial value of this setting is unlimited" */
+};
+
+const uint32_t aws_h2_settings_bounds[AWS_H2_SETTINGS_END_RANGE][2] = {
+    [AWS_H2_SETTINGS_HEADER_TABLE_SIZE][0] = 0,
+    [AWS_H2_SETTINGS_HEADER_TABLE_SIZE][1] = UINT32_MAX,
```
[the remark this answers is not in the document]
graebm: Yeah, I did read it super carefully when setting this up.
I also thought it was weird that unsigned 32bit values are allowed here, while things like stream-id are 31bit

source/h2_frames.c
graebm: I was so happy to read this comment, after all the trivial feedback I throw at everyone 🤣

I ended up naming it `writes_ok`
I tried `all_written` and realized it looked like I was tracking whether I was all done encoding (`all_wrote` was also misleading that way)

source/h2_frames.c
graebm: Until we beef up the `aws_input_stream` API to be Reactive-Streams-esque, I plan to just tick every loop

source/h2_frames.c
graebm: I shuffled them a bit,
I didn't want to lose the grouping by category

source/h2_frames.c
graebm: Oh, I can easily do that.
I was trying to avoid sending a 9-byte HEADERS frame that had no actual header-data in it, knowing that the next time we try to encode there will be a fresh new aws_io_message with tons of space.

Do you think it's valuable to send the valid-but-empty frame earlier?

source/h2_frames.c
graebm: guess not anymore. will change

source/h2_frames.c
graebm: I changed some variable names and added comments.
Hope this is less confusing now

source/h2_frames.c
justinboswell: I agree, thanks for clarifying.

source/hpack.c
justinboswell: :ghost:

source/hpack.c
justinboswell: :dog2: :dash: :walking_man:

include/aws/http/private/h2_frames.h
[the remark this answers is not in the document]
justinboswell: :+1:

source/h2_frames.c
```diff
@@ -754,26 +761,47 @@ static int s_frame_prebuilt_encode(
     struct aws_byte_buf *output,
     bool *complete) {
 
+    (void)encoder;
     struct aws_h2_frame_prebuilt *frame = AWS_CONTAINER_OF(frame_base, struct aws_h2_frame_prebuilt, base);
 
-    if (frame->send_progress.len == frame->encoded_buf.len) {
+    if (frame->cursor.len == frame->encoded_buf.len) {
```
justinboswell: ```/* frame->cursor and frame->encoded_buf are initialized to the same storage buffer */```

Maybe some version of that makes this clearer? Also, I see that's how it works (I THINK), but then there's a frame prefix written right after initialization, so how do the cursor and encoded buf ever match again? My brain compiler is not advanced enough to work it out at a glance.

graebm: I'll add more comments explaining it all

PR #207, opened 2020-03-27
Header block validation

graebm:
Decoder validates a bunch of things about headers.
- Check that header-names use only legal characters.
- Ensure correct pseudo-headers (":status", ":method", etc) are present.
- Ensure pseudo-headers come before normal headers

Decoder also reports whether a header-block was "malformed". This means that the stream should get canceled, but it is not fatal to the connection.

There's still plenty more to validate for 100% spec compliance, but this is more than enough to unblock my Stream state-machine work.


tests/CMakeLists.txt
```diff
@@ -237,9 +239,20 @@ add_h2_decoder_test_set(h2_decoder_headers)
...
 add_h2_decoder_test_set(h2_decoder_err_headers_payload_too_small_for_padding)
 add_h2_decoder_test_set(h2_decoder_err_headers_payload_too_small_for_priority)
+add_h2_decoder_test_set(h2_decoder_malformed_headers_blank_name)
+add_h2_decoder_test_set(h2_decoder_malformed_headers_illegal_name)
+add_h2_decoder_test_set(h2_decoder_malformed_headers_response_to_server)
+add_h2_decoder_test_set(h2_decoder_malformed_headers_request_to_client)
+add_h2_decoder_test_set(h2_decoder_malformed_headers_mixed_pseudoheaders)
+add_h2_decoder_test_set(h2_decoder_malformed_headers_late_pseudoheaders)
+add_h2_decoder_test_set(h2_decoder_malformed_headers_trailer_must_end_stream)
+#add_h2_decoder_test_set(h2_decoder_malformed_header_continues_hpack_parsing)
```
graebm: I uncovered a horrible bug in our HPACK code.
I didn't want it to hold up this PR though.

include/aws/http/private/h2_decoder.h
```diff
@@ -34,17 +34,26 @@
 
 struct aws_h2_decoder_vtable {
     /* For HEADERS header-block: _begin() is called, then 0+ _i() calls, then _end().
-     * No other decoder callbacks will occur in this time. */
+     * No other decoder callbacks will occur in this time.
+     * If something is malformed, no further _i() calls occur, and it is reported in _end() */
     int (*on_headers_begin)(uint32_t stream_id, void *userdata);
-    int (*on_headers_i)(uint32_t stream_id, const struct aws_http_header *header, void *userdata);
-    int (*on_headers_end)(uint32_t stream_id, void *userdata);
+    int (*on_headers_i)(
```
justinboswell: I must have missed this, but why isn't this just ```on_header()```? Likewise with ```on_push_promise()```?

source/h2_decoder.c
```diff
@@ -923,6 +979,146 @@ static int s_state_fn_frame_unknown(struct aws_h2_decoder *decoder, struct aws_b
...
+                DECODER_LOG(ERROR, decoder, "Header name contains invalid characters");
+                goto malformed;
+            }
+        }
+
+        /* #TODO Validate characters used in header_field->value */
+    }
+
+    /* #TODO should we hold back pseudo-headers until s_flush_pseudoheaders() is called?
+     * - PRO: will never deliver pseudo-header that we later determine is invalid.
```
justinboswell: Buffer them. They're not that big, and it makes downstream so much more insane if the headers can be revoked after the fact.

tests/fuzz/fuzz_h2_decoder_correct.c
```diff
@@ -29,12 +29,45 @@
...
+    struct aws_byte_cursor *input,
+    enum header_style header_style) {
 
     struct aws_http_headers *headers = aws_http_headers_new(allocator);
 
-    /* Requires 4 bytes: type, size, and then 1 each for name & value */
-    while (input->len >= 4) {
+    /* There are pretty strict requirements about pseudo-headers, no randomness for now */
+    if (header_style == HEADER_STYLE_REQUEST) {
+        struct aws_http_header method = {.name = aws_http_header_method, .value = aws_http_method_get};
```
justinboswell: Would it be reasonable to swap between a list of verbs for method, maybe toggle between http/https for scheme, and generate domains from usable characters?

tests/fuzz/fuzz_h2_decoder_correct.c
```diff
@@ -211,8 +271,24 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
             struct aws_h2_frame_priority_settings priority = s_generate_priority(&input);
             struct aws_h2_frame_priority_settings *priority_ptr = use_priority ? &priority : NULL;
 
+            /* Server can only receive request-style HEADERS, client can only receive respone-style HEADERS.
```
justinboswell: response

include/aws/http/private/h2_decoder.h
graebm: I dunno, wanted the trio of callbacks to look like a for-loop
It's not a public API

include/aws/http/http.h
```diff
@@ -54,6 +54,7 @@ enum aws_http_errors {
     AWS_ERROR_HTTP_STREAM_IDS_EXHAUSTED,
     AWS_ERROR_HTTP_INVALID_FRAME_SIZE,
     AWS_ERROR_HTTP_COMPRESSION,
+    AWS_ERROR_HTTP_FLOW_CONTROL,
```
JonathanHenson: this doesn't seem very .... descriptive?

source/h2_decoder.c
```diff
@@ -38,22 +39,60 @@ static const uint32_t s_31_bit_mask = UINT32_MAX >> 1;
...
+    PSEUDOHEADER_AUTHORITY,
+    PSEUDOHEADER_PATH,
+    /* Response pseudo-headers */
+    PSEUDOHEADER_STATUS,
+
+    PSEUDOHEADER_COUNT, /* Number of enums */
+};
+
+static enum pseudoheader_name s_header_to_pseudoheader_name(enum aws_http_header_name name) {
+    switch (name) {
```
JonathanHenson: nit take it or leave, you could do this as an array and avoid the branch misses here and the code would be just as readable.

include/aws/http/private/http_impl.h
```diff
@@ -49,9 +61,15 @@ AWS_HTTP_API void aws_http_fatal_assert_library_initialized(void);
...
 
+/* case-sensitive */
 AWS_HTTP_API enum aws_http_method aws_http_str_to_method(struct aws_byte_cursor cursor);
+
+/* case-insensitive */
 AWS_HTTP_API enum aws_http_header_name aws_http_str_to_header_name(struct aws_byte_cursor cursor);
 
+/* case-sensitive */
+AWS_HTTP_API enum aws_http_header_name aws_http_lowercase_str_to_header_name(struct aws_byte_cursor cursor);
+
```
JonathanHenson: I'm confused how this behaves with headers that aren't explicitly in the enum, could you document?

source/h2_decoder.c
```diff
@@ -923,6 +979,146 @@ static int s_state_fn_frame_unknown(struct aws_h2_decoder *decoder, struct aws_b
...
+
+    /* #TODO should we hold back pseudo-headers until s_flush_pseudoheaders() is called?
+     * - PRO: will never deliver pseudo-header that we later determine is invalid.
+     * - CON: buffering */
+
+    /* Deliver header-field via callback */
+    if (current_block->is_push_promise) {
+        DECODER_CALL_VTABLE_STREAM_ARGS(decoder, on_push_promise_i, header_field, name_enum);
+    } else {
+        DECODER_CALL_VTABLE_STREAM_ARGS(decoder, on_headers_i, header_field, name_enum);
```
JonathanHenson: are we doing a callback per header?

source/strutil.c
```diff
@@ -112,3 +112,64 @@ static const bool s_http_whitespace_table[256] = {
...
+    ['Y'] = true, ['Z'] = true,
+
+    ['a'] = true, ['b'] = true, ['c'] = true, ['d'] = true, ['e'] = true, ['f'] = true,  ['g'] = true, ['h'] = true,
+    ['i'] = true, ['j'] = true, ['k'] = true, ['l'] = true, ['m'] = true, ['n'] = true,  ['o'] = true, ['p'] = true,
+    ['q'] = true, ['r'] = true, ['s'] = true, ['t'] = true, ['u'] = true, ['v'] = true,  ['w'] = true, ['x'] = true,
+    ['y'] = true, ['z'] = true,
+};
+
+/* Same as above, but with uppercase characters removed */
+static const bool s_http_lowercase_token_table[256] = {
```
JonathanHenson: it's really too bad we can't do 128 here, that fits nicely into memory fetches on the cache lines. If we find this to ever be a problem (I doubt it is), we could check the high-bit before hitting the array.

source/strutil.c
```diff
@@ -112,3 +112,64 @@ static const bool s_http_whitespace_table[256] = {
...
+    ['y'] = true, ['z'] = true,
+};
+
+static bool s_is_token(struct aws_byte_cursor token, const bool token_table[256]) {
+    if (token.len == 0) {
+        return false;
+    }
+
+    for (size_t i = 0; i < token.len; ++i) {
+        const uint8_t c = token.ptr[i];
```
JonathanHenson: /* TODO: SIMD this when the profiler complains about it */

source/strutil.c
justinboswell: Could also be done as 2x128 arrays. branch on high bit.

source/strutil.c
justinboswell: /* TODO: Evaluate the produce ASM from this and ensure it's vectorized already */

:laughing:

JonathanHenson: I'll buy lunch next time if GCC 4 vectorizes this (especially with the 256 byte array)

source/h2_decoder.c
graebm: Compilers are smart:
https://godbolt.org/z/6FpXN4
vs
https://godbolt.org/z/TFCk9A

I can't actually read assembly, but it looks like there are less branches in the switch version.

tests/fuzz/fuzz_h2_decoder_correct.c
graebm: yeah it's reasonable, I was just trying to get this working again as fast as possible.
I'm going to keep it as-is. This doesn't feel like a high value change right now

tests/fuzz/fuzz_h2_decoder_correct.c
graebm: man I do not know why my spell checker things "respone" is a valid word

include/aws/http/http.h
graebm: Agreed.

This maps will map to [FLOW_CONTROL_ERROR](https://httpwg.org/specs/rfc7540.html#FLOW_CONTROL_ERROR) from HTTP/2.

How about:
`AWS_ERROR_HTTP_FLOW_CONTROL_WINDOW`?
`AWS_ERROR_HTTP_FLOW_CONTROL_VIOLATION`?
`AWS_ERROR_HTTP_FLOW_CONTROL_WINDOW_ERROR`?

source/h2_decoder.c
graebm: The public HTTP API is capable of delivering them in batches.

But this internal API is delivering them one at a time.
Though I **just** at Justin's suggestion changed it so we buffer up the pseudo-headers, so it would be theoretically more efficient to deliver all in a single callback.

Honestly everyone's life would be easier if we buffered all headers deep down here in the decoder and always delivered everything all at once. I know that's how we made the callbacks work in (some? most? all?) of the CRT language bindings. The only reason I don't 100% push for that change here in C is that it is less efficient for languages that will need to make a complete copy of the strings in order to pass them to the user

ramble ramble

source/strutil.c
graebm: Are 256 byte fetches common? I honestly don't know

If the fetch is 128 or less, we'll only go into the upper 128 of this table if someone passes in an character > 127 (an illegal token in this case), so it's likely we'll never load that upper 128 bytes, right?

include/aws/http/http.h
graebm: removing for now.
deal with it in another PR

PR #211, opened 2020-03-31
Hpack debug

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/hpack.c
```diff
@@ -668,7 +685,20 @@ int aws_hpack_insert_header(struct aws_hpack_context *context, const struct aws_
 
     /* Put the header at the "front" of the table */
     struct aws_http_header *table_header = s_dynamic_table_get(context, 0);
-    *table_header = *header;
+
+    /* TODO:: We can optimize this with ring buffer. */
+    /* allocate memory for the name and value, which will be deallocated whenever the entry is evicted from the table or
+     * the table is cleaned up. We keep the pointer in the name pointer of each entry */
+    uint8_t *memory_buffer = aws_mem_calloc(context->allocator, header->name.len + header->value.len, sizeof(uint8_t));
```
JonathanHenson: check that memory_buffer isn't NULL

JonathanHenson: In general this could all be more unified using the aws_Byte_buf_append apis.

source/hpack.c
```diff
@@ -236,6 +236,7 @@ struct aws_hpack_context {
     struct aws_huffman_decoder decoder;
 
     struct {
+        /* Array of headers, pointers to memory we alloced, which needs to be clean up whenever we move an entry out */
```
justinboswell: cleaned

tests/test_hpack.c
```diff
@@ -701,3 +701,60 @@ static int test_hpack_dynamic_table_get(struct aws_allocator *allocator, void *c
     aws_http_library_clean_up();
     return AWS_OP_SUCCESS;
 }
+
+static int s_check_header(
```
JonathanHenson: we we already have negative test cases? I see some failure scenarios above (other than for allocation failures) that need testing.

source/hpack.c
```diff
@@ -668,7 +685,20 @@ int aws_hpack_insert_header(struct aws_hpack_context *context, const struct aws_
 
     /* Put the header at the "front" of the table */
     struct aws_http_header *table_header = s_dynamic_table_get(context, 0);
-    *table_header = *header;
+
+    /* TODO:: We can optimize this with ring buffer. */
+    /* allocate memory for the name and value, which will be deallocated whenever the entry is evicted from the table or
```
justinboswell: Why not use aws_byte_buf for this?

source/hpack.c
TingDaoK: I was trying to keep the size of the total buffer we allocate smaller than the max_table_size setting. So, we need to free the buffer when the entry is evicted. If we use byte_buf_append, we may need a "big enough" buffer, which may not meet the setting.

source/hpack.c
TingDaoK: Yeah, I thought about that. But if we use aws_byte_buf, we will keep a list of aws_byte_buf to track which one should be cleaned up, as we just keep all those aws_byte_cursors.

tests/test_hpack.c
TingDaoK: I am not sure... What kind of failure scenarios besides this one?

tests/test_hpack.c
```diff
@@ -701,3 +701,104 @@ static int test_hpack_dynamic_table_get(struct aws_allocator *allocator, void *c
...
+    (void)ctx;
+
+    aws_http_library_init(allocator);
+    struct aws_hpack_context *context = aws_hpack_context_new(allocator, AWS_LS_HTTP_GENERAL, NULL);
+    ASSERT_NOT_NULL(context);
+
+    /* clang-format off */
+    uint8_t input[] = {
+        0x48, 0x03, '3', '0', '2',      /* ":status: 302" - stored to dynamic table */
+        0x40, 0x01, 'a', 0x01, 'b',     /* "a: b" - stored to dynamic table */
```
graebm, line 728: Just FYI: there's another file `test_h2_headers.c` that has a lot of tests that are much more complex than what existed here in `test_hpack.c`. It's actually composed of the samples from [RFC-7541 Section C](https://httpwg.org/specs/rfc7541.html#rfc.section.C.3.1). The problem is, we don't have ALL the samples. If we did we would have picked up this bug already. Looks like we stopped short including any examples that spanned multiple messages, so we didn't catch this bug where we insert something in the table and look it up again later.

In the future we should strongly consider getting the rest of the examples into that other test file

source/hpack.c
```diff
@@ -523,6 +537,9 @@ static int s_dynamic_table_shrink(struct aws_hpack_context *context, size_t max_
                 goto error;
             }
         }
+
+        /* clean-up the memory we allocate for it */
```
graebm, line 541: /* clean up the memory we allocated to hold the name and value string*/

PR #213, opened 2020-04-02
Hpack more debug

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- Add RFC 7541 C 3.2 3.3 4.2 4.3 5.2 5.3 6.2 6.3 as test cases
- Debug the logic for finding index & and the hash table.


source/hpack.c
```diff
@@ -691,16 +700,18 @@ int aws_hpack_insert_header(struct aws_hpack_context *context, const struct aws_
...
-        return AWS_OP_ERR;
+    /* if buf_memory_size is 0, no memory needed, we will insert the empty header into dynamic table */
+    if (buf_memory_size) {
+        uint8_t *buf_memory = aws_mem_acquire(context->allocator, buf_memory_size);
+        if (!buf_memory) {
+            return AWS_OP_ERR;
+        }
+        struct aws_byte_buf buf = aws_byte_buf_from_empty_array(buf_memory, buf_memory_size);
+        /* Copy header, then backup strings into our own allocation */
+        *table_header = *header;
```
graebm, line 711: we're not setting `table_header`'s members to anything in the else-0-memory-size case, and I'm not positive that s_dynamic_table_get() returns zeroed-memory in all possible cases. Let's explicitly set its name/value/compression in an else block

tests/test_hpack.c
```diff
@@ -802,3 +803,48 @@ static int test_hpack_decode_empty_value(struct aws_allocator *allocator, void *
     aws_http_library_clean_up();
     return AWS_OP_SUCCESS;
 }
+
+/* Test header with empty name and value */
+AWS_TEST_CASE(hpack_decode_empty_header, test_hpack_decode_empty_header)
```
graebm, line 808: [style/opinion] A test like this would be much shorter and easier to understand if it used the underlying aws_hpack_insert_header()/aws_hpack_find_index() functions, rather than feeding a binary buffer through the full decode sequence.

tests/test_h2_headers.c
```diff
@@ -269,96 +269,465 @@ static int s_test_ex_2_4_init(struct header_test_fixture *fixture) {
...
+    aws_byte_buf_write(&fixture->expected_encoding_buf[index], encoded_3, sizeof(encoded_3));
+
+    struct aws_http_header last_entry_3 = DEFINE_STATIC_HEADER(":authority", "www.example.com", USE_CACHE);
+    fixture->last_entry_dynamic_table[index] = last_entry_3;
+    fixture->dynamic_table_len[index] = 3;
 
     return AWS_OP_SUCCESS;
 }
-HEADER_TEST(h2_header_ex_3_1, s_test_ex_3_1_init, NULL);
+HEADER_REQUST_RESPONSE_TEST(h2_header_ex_3, s_test_ex_3_init, NULL);
```
graebm, line 516: REQUST -> REQUEST

tests/test_hpack.c
```diff
@@ -759,43 +759,66 @@ static int test_hpack_decode_indexed_from_dynamic_table(struct aws_allocator *al
...
+    So at this point dynamic table should look like:
+        *  INDEX   NAME    VALUE
+        *  62       a        b
+        *  63       ""       ""
+        *  64      :status 302
+    */
+    bool found_value = false;
+    ASSERT_TRUE(aws_hpack_find_index(context, &header1, true, &found_value) == 64);
+    ASSERT_TRUE(aws_hpack_find_index(context, &empty_header, true, &found_value) == 63);
+    ASSERT_TRUE(aws_hpack_find_index(context, &header2, true, &found_value) == 62);
```
graebm, line 821: [trivial] `ASSERT_UINT_EQUALS(64, aws_hpack_find_index(context, &header1, true, &found_value)`

results in printing nicer errors when a test fails

source/hpack.c
```diff
@@ -472,41 +470,44 @@ static const struct aws_http_header *s_get_header_u64(const struct aws_hpack_con
...
-        *found_value = ((const struct aws_http_header *)elem->key)->value.len;
-    } else {
-        /* If not found, check name only table. Don't set found_value, it will be false */
-        aws_hash_table_find(&context->dynamic_table.reverse_lookup_name_only, &header->name, &elem);
+        goto trans_index_from_dynamic_table;
     }
+    return 0;
 
+trans_index_from_dynamic_table:
     if (elem) {
```
graebm, line 510: [style] we only jump here if `elem` is non-null, replace the if-branch with an `AWS_ASSERT(elem);`

reasoning: it's confusing to have branches that shouldn't be possible based on our programming. there are exceptions for checking results of extremely complex machinery, but any error here would come from within this same function.

PR #214, opened 2020-04-02
Settings hook up with hpack

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/h2_frames.c
```diff
@@ -1140,9 +1140,13 @@ int aws_h2_encode_frame(
     return AWS_OP_SUCCESS;
 }
 
-void aws_h2_frame_encoder_set_setting_header_table_size(struct aws_h2_frame_encoder *encoder, uint32_t data) {
-    /* TODO:: pass down to the HPACK context aws_hpack_resize_dynamic_table() */
+int aws_h2_frame_encoder_set_setting_header_table_size(struct aws_h2_frame_encoder *encoder, uint32_t data) {
+    if (aws_hpack_resize_dynamic_table(encoder->hpack, data, true)) {
```
TingDaoK, line 1144: I know, it seems very inefficient, especially when in a case with tons of updates in the same value. But [H2](https://httpwg.org/specs/rfc7540.html#SETTINGS) "Each parameter in a SETTINGS frame replaces any existing value for that parameter." I just don't know... 

If that is what it requests, let's keep it this way. Or, I'll update it with just two resize dynamic table operations.

PR #216, opened 2020-04-03
stream state-machine

graebm:
- Only header callbacks hooked up so far.
- Connection checks for active stream.
  - If stream not found, error raised or frame ignored, according to rules.
- Stream checks that frame is allowed, given current state.
- When END_STREAM is received, stream closes.
- When stream closes, connection marks it complete.
- Stream errors result stream sending RST_STREAM frame before closing.
- Connection errors shut down channel


source/h2_stream.c
```diff
@@ -58,6 +59,123 @@ static struct aws_h2_connection *s_get_h2_connection(const struct aws_h2_stream
...
+};
+
+static int s_check_state_allows_frame_type(const struct aws_h2_stream *stream, enum aws_h2_frame_type frame_type) {
+    AWS_PRECONDITION_ON_CHANNEL_THREAD(stream);
+
+    const enum aws_h2_stream_state state = stream->thread_data.state;
+
+    bool allowed;
+    if (stream->base.server_data) {
+        allowed = s_server_state_allows_frame_type[state][frame_type];
```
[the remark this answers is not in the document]
graebm: state is just an enum, and these arrays are `AWS_H2_STREAM_STATE_COUNT` big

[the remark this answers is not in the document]
JonathanHenson: State is an int. Question here is how much control you have over it. Assuming complete control I think this is fine. Maybe a precondition check would be sufficient.

source/h2_connection.c
```diff
@@ -873,6 +1116,19 @@ static struct aws_http_stream *s_connection_make_request(
     return NULL;
 }
 
+static bool s_connection_is_open(const struct aws_http_connection *connection_base) {
+    struct aws_h2_connection *connection = AWS_CONTAINER_OF(connection_base, struct aws_h2_connection, base);
+    bool is_open;
+
+    { /* BEGIN CRITICAL SECTION */
+        s_lock_synced_data(connection);
+        is_open = connection->synced_data.is_open;
```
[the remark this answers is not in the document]
graebm: done

source/h2_connection.c
```diff
@@ -527,7 +560,178 @@ static void s_try_write_outgoing_frames(struct aws_h2_connection *connection) {
...
+int s_decoder_on_end_stream(uint32_t stream_id, void *userdata) {
+    struct aws_h2_connection *connection = userdata;
+
+    /* Not calling s_get_active_stream_for_incoming_frame() here because END_STREAM
+     * isn't an actual frame type. It's a flag on DATA or HEADERS frames, and we
+     * already checked the legality of those frames in their respective callbacks. */
+
+    struct aws_hash_element *found = NULL;
+    aws_hash_table_find(&connection->thread_data.active_streams_map, (void *)(size_t)stream_id, &found);
+    if (found) {
```
justinboswell, line 725: If this is not found, is that a problem? Seems like that should be impossible?

source/h2_connection.c
```diff
@@ -715,6 +919,44 @@ static void s_stream_complete(struct aws_h2_connection *connection, struct aws_h
...
+                &connection->thread_data.closed_streams_where_frames_might_trickle_in,
+                (void *)(size_t)stream_id,
+                (void *)(size_t)closed_when,
+                NULL)) {
+
+            AWS_H2_STREAM_LOG(ERROR, stream, "Failed inserting ID into map of recently closed streams");
+            return AWS_OP_ERR;
+        }
+
+        /* #TODO remove entries from closed_streams_where_frames_might_trickle_in after some period of time */
```
justinboswell, line 953: Wait so are they in there indefinitely right now?

source/h2_stream.c
```diff
@@ -58,6 +59,123 @@ static struct aws_h2_connection *s_get_h2_connection(const struct aws_h2_stream
 #define AWS_PRECONDITION_ON_CHANNEL_THREAD(STREAM)                                                                     \
     AWS_PRECONDITION(aws_channel_thread_is_callers_thread(s_get_h2_connection(STREAM)->base.channel_slot->channel))
 
+static bool s_client_state_allows_frame_type[AWS_H2_STREAM_STATE_COUNT][AWS_H2_FRAME_T_UNKNOWN] = {
```
justinboswell, line 62: Nitpick: Could we add an ```AWS_H2_FRAME_T_COUNT = AWS_H2_FRAME_T_UNKNOWN``` to that enum to make usages like this less unintuitive?

tests/test_h2_client.c
```diff
@@ -22,6 +23,9 @@
     AWS_TEST_CASE(NAME, s_test_##NAME);                                                                                \
     static int s_test_##NAME(struct aws_allocator *allocator, void *ctx)
 
+#define DEFINE_HEADER(NAME, VALUE)                                                                                     \
```
justinboswell, line 26: I almost wonder if we should put this in a public header. It's going to be everywhere in our code using this.

include/aws/http/private/h2_connection.h
```diff
@@ -62,26 +63,41 @@ struct aws_h2_connection {
...
 
-    /* Any thread may touch this data, but the lock must be held */
+    /* Any thread may touch this data, but the lock must be held (unless it's an atomic) */
     struct {
+        /* For checking status from outside the event-loop thread. */
+        struct aws_atomic_var is_open;
+
+        /* If non-zero, reason to immediately reject new streams. (ex: closing) */
+        struct aws_atomic_var new_stream_error_code;
+
```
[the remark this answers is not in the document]
graebm: 🤷‍♂️
they'll probably flip back to being non-atomic when we need to do something transactional with them. I think it's fine

include/aws/http/private/h2_connection.h
```diff
@@ -62,26 +63,41 @@ struct aws_h2_connection {
...
-        /* For checking status from outside the event-loop thread. */
-        bool is_open;
-
     } synced_data;
 };
 
+/**
+ * The action which caused the stream to close.
+ */
+enum aws_h2_stream_closed_when {
```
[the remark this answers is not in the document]
graebm: 🤷‍♂️
fun fact: You know I've spent at least 5 minutes trying out different words for this Since the values all sounded good with "when" I stuck with it for the type name.
It's not a public enum, I think it's fine

include/aws/http/private/h2_frames.h
```diff
@@ -162,6 +162,25 @@ AWS_EXTERN_C_BEGIN
...
+const char *aws_h2_error_code_to_str(enum aws_h2_error_code h2_error_code);
+
+/**
+ * Translate an AWS_ERROR_* into the appropriate HTTP/2 error-code to sending in a RST_STREAM or GOAWAY frame.
+ *
+ * Ex: AWS_ERROR_HTTP_PROTOCOL_ERROR -> AWS_H2_ERR_PROTOCOL_ERROR (0x1)
+ *
+ * AWS_H2_ERR_INTERNAL_ERROR is returned if nothing else matches.
+ *
+ * #TODO - is this really how we want to handle things? It seems weird to tell OUR user
```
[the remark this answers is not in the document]
graebm: I would like to discuss this more face to face

source/h2_connection.c
graebm: yes
I was trying to keep down the PR size so didn't tackle it yet

source/h2_stream.c
```diff
@@ -143,3 +303,158 @@ int aws_h2_stream_on_activated(struct aws_h2_stream *stream, bool *out_has_outgo
...
+        if (aws_h2_connection_on_stream_closed(
+                s_get_h2_connection(stream),
+                stream,
+                AWS_H2_STREAM_CLOSED_WHEN_BOTH_SIDES_END_STREAM,
+                AWS_ERROR_SUCCESS)) {
+            return AWS_OP_ERR;
+        }
+
+    } else {
+        /* Else can't close until our side sends END_STREAM */
```
TingDaoK, line 454: I was going to say why we don't check if the state is valid for receiving an "end stream". Then, I found out it will only be here if the stream is in the active_stream_map.

source/h2_connection.c
```diff
@@ -216,6 +234,20 @@ static struct aws_h2_connection *s_connection_new(
         goto error;
     }
 
+    if (aws_hash_table_init(
+            &connection->thread_data.closed_streams_where_frames_might_trickle_in,
```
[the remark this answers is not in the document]
graebm: Yeah, I purposefully made this waaaay too explicit because it's so stupidly nuanced.

`recently_closed_streams` seemed like a good name until I realized that I didn't need to store the IDs of streams that closed in certain ways.

source/h2_connection.c
graebm: We're just reading frames off the wire. Nothing stopping a bad implementation or attacker from sending us frames for a stream that doesn't exist

source/h2_stream.c
graebm: Yeah, I'll add that

tests/test_h2_client.c
graebm: aws_http_header_from_c_str(const char *name, const char *value, enum aws_http_header_compression compression);

struct aws_http_header r

source/h2_stream.c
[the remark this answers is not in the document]
graebm: Yeah, the decoder completely hides the existence of anything with an invalid type. I'll add the precondition and a comment

PR #219, opened 2020-04-03
H2 cookie header field

TingDaoK:
*Issue #, if available:*

*Description of changes:*


include/aws/http/private/http_impl.h
```diff
@@ -46,6 +46,9 @@ enum aws_http_header_name {
     /* Response pseudo-headers */
     AWS_HTTP_HEADER_STATUS,
 
+    /* Cookie header */
+    AWS_HTTP_HEADER_COOKIE,
+
     /* Regular headers */
```
graebm, line 52: trivial: this can just go with the rest of the /* Regular headers */ it doesn't need a comment of its own

source/h2_decoder.c
```diff
@@ -245,6 +248,12 @@ struct aws_h2_decoder {
          * A malformed header-block is not a connection error, it's a Stream Error (RFC-7540 5.4.2).
          * We continue decoding and report that it's malformed in on_headers_end(). */
         bool malformed;
+
+        /* Buffer up cookie header fields to concatenate separate ones */
+        struct aws_byte_buf cookies;
+        /* Compression type for the concatenated cookie header. Let's say if any separate one is not cached, the last
+         * one will not be cached */
+        enum aws_http_header_compression cookie_header_compression_type;
```
graebm, line 256: /* If separate cookie fields have different compression types, the concatenated cookie uses the strictest type. */

source/h2_decoder.c
```diff
@@ -314,10 +323,16 @@ struct aws_h2_decoder *aws_h2_decoder_new(struct aws_h2_decoder_params *params)
...
+
     return decoder;
 
-failed_new_hpack:
+buffer_init_failed:
 array_list_failed:
+failed_new_hpack:
     aws_mem_release(params->alloc, allocation);
 failed_alloc:
     return NULL;
```
graebm, line 338: if we jump to `array_list_failed:` or `buffer_init_failed:` then there's stuff we fail to clean up.
maybe make this less error-prone by just having one `error:` label, like:
```
error:
if (decoder) {
  aws_hpack_context_destroy(...)
  aws_array_list_clean_up(...)
  aws_byte_buf_clean_up(...)
}
aws_mem_release(params->alloc, allocation)
```

We've tried to make all of our clean_up()/release()/destroy() functions safe to call on something whose initialization failed, or never happened (as long as we zeroed it all out to start)

source/h2_decoder.c
```diff
@@ -1263,6 +1282,30 @@ static int s_process_header_field(struct aws_h2_decoder *decoder, const struct a
...
+                struct aws_byte_cursor delimiter = aws_byte_cursor_from_c_str("; ");
+                if (aws_byte_buf_append_dynamic(&current_block->cookies, &delimiter)) {
+                    DECODER_LOG(ERROR, decoder, "Store cookie delimiter to buffer failed");
+                    return AWS_OP_ERR;
+                }
+            }
+            if (aws_byte_buf_append_dynamic(&current_block->cookies, &header_field->value)) {
+                DECODER_LOG(ERROR, decoder, "Store the current cookie value to buffer failed");
+                DECODER_LOGF(
+                    DEBUG, decoder, "Value of cookie is '" PRInSTR "'", AWS_BYTE_CURSOR_PRI(header_field->value));
```
graebm, line 1303: super trivial: don't bother logging errors that can only fail due to out-of-memory. They're so incredibly unlikely it's barely worth the effort of writing the commands. And if a header actually runs us out of memory, probably not a good idea to try and log its value

source/h2_decoder.c
```diff
@@ -1298,6 +1341,23 @@ static int s_state_fn_header_block_loop(struct aws_h2_decoder *decoder, struct a
                 return AWS_OP_ERR;
             }
 
+            if (decoder->header_block_in_progress.cookies.len) {
```
graebm, line 1344: don't flush the cookie if the header block is malformed. Once we detect the header block is malformed we stop delivering individual headers.

That behavior would be more clear if we also moved the `s_flush_pseudoheaders()` call inside the `if (!malformed)` block. I thought it was simpler-looking to have that function do its own check, but it's obviously more confusing then helpful

source/h2_decoder.c
```diff
@@ -1263,6 +1282,30 @@ static int s_process_header_field(struct aws_h2_decoder *decoder, const struct a
 
         /* #TODO Validate characters used in header_field->value */
 
+        if (name_enum == AWS_HTTP_HEADER_COOKIE) {
```
graebm, line 1285: trivial/efficiency: `switch(name_enum)`
we'll probably add special code for more headers in the future, and compilers can sometimes do cool tricks with switch statements

source/h2_decoder.c
```diff
@@ -1282,6 +1327,28 @@ static int s_process_header_field(struct aws_h2_decoder *decoder, const struct a
...
+    struct aws_header_block_in_progress *current_block = &decoder->header_block_in_progress;
+    if (current_block->malformed) {
+        return AWS_OP_SUCCESS;
+    }
+    if (current_block->cookies.len == 0) {
+        /* Nothing to flush */
+        return AWS_OP_SUCCESS;
+    }
+    struct aws_http_header concatenated_cookie;
+    concatenated_cookie.name = aws_byte_cursor_from_c_str("cookie");
```
graebm, line 1340: trivial: ` = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("cookie");`
compile-time strlen()

PR #221, opened 2020-04-06
H2 stream receives DATA

graebm:
Hook up DATA decoder callbacks so they go through connection and stream to the end-user.


source/h2_stream.c
```diff
@@ -430,6 +430,32 @@ int aws_h2_stream_on_decoder_headers_end(
...
+        return s_send_rst_and_close_stream(stream, aws_last_error());
+    }
+
+    if (!stream->thread_data.received_main_headers) {
+        /* #TODO Not 100% sure whether this is Stream Error or Connection Error. */
+        AWS_H2_STREAM_LOG(ERROR, stream, "Malformed message, received DATA before main HEADERS");
+        return s_send_rst_and_close_stream(stream, AWS_ERROR_HTTP_PROTOCOL_ERROR);
+    }
+
+    /* #TODO Update connection's flow-control window */
```
JonathanHenson, line 446: you have this comment in two places.... you probably don't want to handle twice tho.

graebm: oh yeah, one of those should be "Update **stream's** flow-control window"

PR #222, opened 2020-04-08
H2 send data

graebm:
H2 streams can now send data!

Algorithm is:

Once all other frames have been written, try to fill the remainder of an aws_io_message with DATA frames. Be "fair" to multiple streams by letting each one write 1 DATA frame in a round-robin manner. If a stream is "stalled" (the body didn't stream as much data as we asked it to) then stop trying to write that stream's data until the next event-loop tick.


source/h2_connection.c
```diff
@@ -433,120 +435,172 @@ static void s_outgoing_frames_task(struct aws_channel_task *task, void *arg, enu
...
+    }
+
+    /* If outgoing_frames_queue emptied, then write as many DATA frames from outgoing_streams_list as possible. */
+    if (aws_linked_list_empty(outgoing_frames_queue)) {
+        if (s_encode_data_from_outgoing_streams(connection, &msg->message_data)) {
+            goto error;
+        }
+    }
+
+    if (msg->message_data.len) {
```
graebm, line 450: it's not really "new code" from here to the end of the function. I moved big chunks of code out into their own functions (`s_encode_outgoing_frames_queue()` and `s_encode_data_from_outgoing_streams()`) and it's messing up the diffs

include/aws/http/private/h2_frames.h
```diff
@@ -220,6 +221,7 @@ int aws_h2_encode_frame(
  * Attempt to encode a DATA frame into the output buffer.
  * AWS_OP_ERR is returned if encoder encounters an unrecoverable error.
  * body_complete will be set true if encoder reaches the end of the body_stream.
+ * body_stalled will be set true if body_stream didn't provide as much data as we would have liked.
```
TingDaoK, line 224: (Personal opinion) When I read this, I am actually a little bit confused about "we would have like". After reading the implementation, I know what it means. It might be helpful to clarify this a little bit more.

graebm:
```
* body_stalled will be true if aws_input_stream_read() stopped early (didn't
 * complete, though more space was available).
 ```

graebm:
> ... we can revisit this later.

Totally, this will evolve 🦠🐟🦕🚀

PR #225, opened 2020-04-08
Transform request to h2

TingDaoK:
*Issue #, if available:*

*Description of changes:*

Transform request message to h2 style. 

- If pseudo headers appear at the beginning of the headers list in aws-http-message. We basically just copy everything from the original one
- Else, we create 4 pseudo headers for it. 
- For regular headers, we convert the name of headers to lowercase. And handled some special case of header fields in h2.

TODO:
- For aws_http_message, we still set all the pseudo headers from the headers list of it, which may need to be changed. 
- For changing h1 request message to h2 style, :scheme is set as "https" all the time, which can be set from the request line(URI) instead. We will need to change aws_http_message to do it. 
- For :authority, We set it from "host" header field. I am not sure what it will be, if "host" is not set. From my understanding, it can also get from URI, which leads to the API change again.
- Elasticurl can create h2 connection now. As long as we send "h2" through TLS. So, we may want to update elasticurl to support that. I can do it in this PR, if needed!


include/aws/http/private/h2_connection.h
```diff
@@ -141,4 +141,9 @@ int aws_h2_connection_on_stream_closed(
     enum aws_h2_stream_closed_when closed_when,
     int aws_error_code);
 
+/* Transform the request to h2 style headers */
+struct aws_http_headers *aws_h2_create_headers_from_request(
+    struct aws_http_message *request,
```
JonathanHenson, line 146: does request already have an allocator object on it?

tests/test_h2_client.c
```diff
@@ -301,6 +301,54 @@ TEST_CASE(h2_client_stream_complete) {
     return s_tester_clean_up();
 }
 
+/* Test that h2 stream can take a h1 request massega and transfrom it to h2 style to send it. */
+TEST_CASE(h2_client_stream_with_h1_request_message) {
```
JonathanHenson, line 305: Do we need a test for if the headers were explicitly set on the request?

include/aws/http/private/h2_connection.h
TingDaoK: Yeah, but aws_http_message is not visible from here, we need another AWS_HTTP_API to get the allocator from it. Do I just add one to get the allocator?

tests/test_h2_client.c
TingDaoK: The other tests of the stream set pseudo headers explicitly on the request. Maybe we just let those to test that situation.

include/aws/http/private/h2_connection.h
TingDaoK: I thought about it before, but I am not sure about the API change. So I just let the caller give me an allocator, instead of taking the allocator from request.

graebm: this seems fine

source/h2_connection.c
```diff
@@ -941,6 +942,112 @@ static void s_stream_complete(struct aws_h2_connection *connection, struct aws_h
...
+        enum aws_http_header_name name_enum = aws_http_lowercase_str_to_header_name(lower_name_cursor);
+        switch (name_enum) {
+            case AWS_HTTP_HEADER_COOKIE:
+                /* split cookie if USE CACHE */
+                if (header_iter.compression == AWS_HTTP_HEADER_COMPRESSION_USE_CACHE) {
+                    struct aws_array_list cookie_chunks;
+                    if (aws_array_list_init_dynamic(&cookie_chunks, alloc, 4, sizeof(struct aws_byte_cursor))) {
+                        goto error;
+                    }
+                    if (aws_byte_cursor_split_on_char(&header_iter.value, ';', &cookie_chunks)) {
```
graebm, line 1015: `aws_byte_cursor_next_split()` would let you iterate the cookies with 0 allocations.
```
struct aws_byte_cursor cookie_chunk = {0};
while (aws_byte_cursor_next_split(&header_iter.value, ';', &cookie_chunk)) {
    ...
}
```

why do both these split functions exist? the allocating one came first, the zero-allocation one came second

source/h2_connection.c
```diff
@@ -941,6 +942,112 @@ static void s_stream_complete(struct aws_h2_connection *connection, struct aws_h
...
+                        goto error;
+                    }
+                    for (size_t i = 0; i < aws_array_list_length(&cookie_chunks); i++) {
+                        struct aws_byte_cursor cookie_chunk;
+                        AWS_ZERO_STRUCT(cookie_chunk);
+                        if (aws_array_list_get_at(&cookie_chunks, &cookie_chunk, i)) {
+                            aws_array_list_clean_up(&cookie_chunks);
+                            goto error;
+                        }
+                        aws_http_headers_add(result, lower_name_cursor, aws_strutil_trim_http_whitespace(cookie_chunk));
```
graebm, line 1026: need to check if `aws_http_headers_add()` failed

PR #228, opened 2020-04-11
Automatically reject PUSH_PROMISE streams

graebm:
Until we have a need for it, PUSH_PROMISE is not a fully supported feature. Therefore, promised streams are automatically rejected in a manner compliant with RFC-7540

Why, you ask, don't we simply send SETTINGS_ENABLE_PUSH=0 in the initial SETTINGS frame and call it a day? Because it's theoretically possible for a server to start sending PUSH_PROMISE frames in the initial response, before sending the ACK to the initial SETTINGS.

### DIFF
--- a/source/h2_connection.c
+++ b/source/h2_connection.c
@@ -64,6 +64,10 @@ static void s_cross_thread_work_task(struct aws_channel_task *task, void *arg, e
 static void s_outgoing_frames_task(struct aws_channel_task *task, void *arg, enum aws_task_status status);
 static int s_encode_outgoing_frames_queue(struct aws_h2_connection *connection, struct aws_byte_buf *output);
 static int s_encode_data_from_outgoing_streams(struct aws_h2_connection *connection, struct aws_byte_buf *output);
+static int s_record_closed_stream(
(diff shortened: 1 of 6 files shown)


source/h2_connection.c
```diff
@@ -772,6 +778,41 @@ int s_decoder_on_headers_end(
     return AWS_OP_SUCCESS;
 }
 
+int s_decoder_on_push_promise(uint32_t stream_id, uint32_t promised_stream_id, void *userdata) {
+    struct aws_h2_connection *connection = userdata;
+    AWS_ASSERT(connection->base.client_data); /* decoder enforces that server cannot receive PUSH_PROMISE */
```
JonathanHenson, line 783: so we're going to crash a server process because of a misbehaving client? It seems better behavior here would be to process the error and quick kill the stream/connection.

graebm: That particular bit of error-handling is done in h2_decoder.c, before we get out here to the connection. We do as much error checking as we possibly can in the decoder.

I was hoping the assert and comment would make that clear, but I guess it was confusing. I'll rephrase the comment to be more clear

PR #230, opened 2020-04-15
Send window_update

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- Send stream and connection window_update back whenever we receive a flow controlled frame


include/aws/http/private/h2_connection.h
```diff
@@ -81,6 +81,11 @@ struct aws_h2_connection {
          * Reduce the space after sending a flow-controlled frame. Increment after receiving WINDOW_UPDATE for
          * connection */
         size_t window_size_peer;
+
+        /* Flow-control of connection for this side. Indicating the buffer capacity of our peer.
+         * Reduce the space after receiving a flow-controlled frame. Increment after sending WINDOW_UPDATE for
+         * connection */
+        size_t window_size_self;
```
graebm, line 88: ~~Indicating the buffer capacity of our peer~~

include/aws/http/private/h2_connection.h
```diff
@@ -117,7 +122,10 @@ enum aws_h2_data_encode_status {
     AWS_H2_DATA_ENCODE_ONGOING_WINDOW_STALLED,
 };
 
+/* When window size is too small to fit the possible padding into it, we stop sending data and wait for WINDOW_UPDATE */
 #define AWS_H2_MIN_WINDOW_SIZE (256)
+/* Initial connection flow-control window with 65535 bytes (RFC 6.9.2) */
+#define AWS_H2_INITIAL_WINDOW_SIZE (65535)
```
graebm, line 128: why did you add `AWS_H2_INITIAL_WINDOW_SIZE`? we already had `aws_h2_settings_initial[AWS_H2_SETTINGS_INITIAL_WINDOW_SIZE]`?

include/aws/http/private/h2_decoder.h
```diff
@@ -56,9 +56,11 @@ struct aws_h2_decoder_vtable {
         void *userdata);
     int (*on_push_promise_end)(uint32_t stream_id, bool malformed, void *userdata);
 
-    /* Called repeatedly as DATA frames are processed.
-     * This may fire multiple times per actual DATA frame. */
-    int (*on_data)(uint32_t stream_id, struct aws_byte_cursor data, void *userdata);
+    /* For DATA header-block: _begin() is called, then 0+ _i() calls, then _end().
```
graebm, line 59: /* For DATA ~~header-block~~ frame: _begin() is called...

source/h2_connection.c
```diff
@@ -845,10 +846,58 @@ int s_decoder_on_push_promise(uint32_t stream_id, uint32_t promised_stream_id, v
...
-int s_decoder_on_data(uint32_t stream_id, struct aws_byte_cursor data, void *userdata) {
+int s_decoder_on_data_begin(uint32_t stream_id, uint32_t payload_len, void *userdata) {
     struct aws_h2_connection *connection = userdata;
 
-    /* #TODO Send a window update frame back. */
+    /* A receiver that receives a flow-controlled frame MUST always account for its contribution against the connection
+     * flow-control window, unless the receiver treats this as a connection error */
+    if (connection->thread_data.window_size_self < payload_len) {
+        /* TODO: Figure out what kind error this is, I failed to find it from spec, and it is possible to happen */
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
```
graebm, line 856: A receiver MAY respond with a stream error (Section 5.4.2) or connection error (Section 5.4.1) of type FLOW_CONTROL_ERROR if it is unable to accept a frame.

source/h2_connection.c
```diff
@@ -845,10 +846,58 @@ int s_decoder_on_push_promise(uint32_t stream_id, uint32_t promised_stream_id, v
...
+    }
+    connection->thread_data.window_size_self -= payload_len;
```
graebm, line 858: use `aws_sub_size_checked()` instead of checking and subtracting yourself

source/h2_connection.c
```diff
@@ -845,10 +846,58 @@ int s_decoder_on_push_promise(uint32_t stream_id, uint32_t promised_stream_id, v
...
+    if (s_get_active_stream_for_incoming_frame(connection, stream_id, AWS_H2_FRAME_T_DATA, &stream)) {
+        return AWS_OP_ERR;
+    }
+
+    if (stream) {
+        if (aws_h2_stream_on_decoder_data_begin(stream, payload_len)) {
+            return AWS_OP_ERR;
+        }
+        /* send a stream window_update frame to automatically maintain the stream self window size */
+        struct aws_h2_frame *stream_window_update_frame =
```
graebm, line 869: [trivial] let's send the stream's WINDOW_UPDATE from stream.c

At first I wrote "Just for the sake of maintaining the logical split between stream & connection"

But later I realized the stream might decide to send RST_STREAM and close itself from within this call, so it would be a crime to send a WINDOW_UPDATE for the stream, and also the `stream` ptr itself could be invalid at this point

source/h2_stream.c
```diff
@@ -535,9 +536,10 @@ int aws_h2_stream_on_decoder_push_promise(struct aws_h2_stream *stream, uint32_t
         s_get_h2_connection(stream), promised_stream_id, AWS_H2_ERR_REFUSED_STREAM);
 }
 
-int aws_h2_stream_on_decoder_data(struct aws_h2_stream *stream, struct aws_byte_cursor data) {
+int aws_h2_stream_on_decoder_data_begin(struct aws_h2_stream *stream, uint32_t data_payload_len) {
     AWS_PRECONDITION_ON_CHANNEL_THREAD(stream);
 
+    stream->thread_data.window_size_self -= data_payload_len;
     if (s_check_state_allows_frame_type(stream, AWS_H2_FRAME_T_DATA)) {
```
graebm, line 543: Don't alter the stream until we've checked whether it's even legal to receive the DATA frame

Also, if stream's window_size goes below 0 we need to RST_STREAM with FLOW_CONTROL_ERROR

tests/CMakeLists.txt
```diff
@@ -353,6 +353,7 @@ add_test_case(h2_client_stream_send_data_controlled_by_stream_window_size)
 add_test_case(h2_client_stream_send_data_controlled_by_negative_stream_window_size)
 add_test_case(h2_client_stream_send_data_controlled_by_connection_window_size)
 add_test_case(h2_client_stream_send_data_controlled_by_connection_and_stream_window_size)
+add_test_case(h2_client_stream_send_window_update)
```
graebm, line 356: add these todos please

#TODO add_test_case(h2_client_stream_err_flow_control)
#TODO add_test_case(h2_client_conn_err_flow_control)

we can't actually test this right now, since we always auto-update and can't even tweak INITIAL_WINDOW_SIZE and MAX_FRAME_SIZE yet to make it possible 🙃

tests/test_h2_client.c
```diff
@@ -639,7 +639,9 @@ TEST_CASE(h2_client_stream_err_receive_data_before_headers) {
 
     /* validate that stream sent RST_STREAM */
     ASSERT_SUCCESS(h2_fake_peer_decode_messages_from_testing_channel(&s_tester.peer));
-    struct h2_decoded_frame *rst_stream_frame = h2_decode_tester_latest_frame(&s_tester.peer.decode);
+    /* last two frames will be window updates */
+    struct h2_decoded_frame *rst_stream_frame =
+        h2_decode_tester_get_frame(&s_tester.peer.decode, h2_decode_tester_frame_count(&s_tester.peer.decode) - 3);
     ASSERT_INT_EQUALS(AWS_H2_FRAME_T_RST_STREAM, rst_stream_frame->type);
```
graebm, line 645: Whoa, couple of things here:
First, it's a crime to send anything to a stream after RST_STREAM
> An endpoint that receives any frame other than PRIORITY after receiving a RST_STREAM MUST treat that as a stream error (Section 5.4.2) of type STREAM_CLOSED

Second, we should try if possible not to make the tests suuper dependent on the order of frames. Since as we see here, the connection can now send WINDOW_UPDATE frames that shouldn't affect stream frames.

There are some helper functions like `h2_decode_tester_find_stream_frame()` you should use to find the RST_STREAM, instead of looking at an absolute index. There's another find() function that finds a frame of the specified type, but doesn't care about stream_id. We could add another that finds frames of any type, but with the specified stream_id, something like `h2_decode_tester_find_stream_frame_any_type()`

tests/h2_test_helper.c
```diff
@@ -458,7 +458,7 @@ static struct aws_h2_decoder_vtable s_decoder_vtable = {
     .on_push_promise_begin = s_decoder_on_push_promise_begin,
     .on_push_promise_i = s_decoder_on_push_promise_i,
     .on_push_promise_end = s_decoder_on_push_promise_end,
-    .on_data = s_decoder_on_data,
+    .on_data_i = s_decoder_on_data_i,
```
graebm, line 474: Could you please modify the decode_tester to do testing on this API change? In addition to storing info about the callbacks that happen, this code also validates that the callbacks happen in the order they're supposed to.

Check out how we do things with the other _begin()/i()/end() callbacks work

Instead of creating a new `h2_decoded_frame` per on_data_i() call, create a new frame in on_data_begin() and append to its data as on_data_i() callbacks happen. End the frame when on_data_end() is called.

TingDaoK: Oh, I just realized it after I made the PR. And it's actually already there. Sorry for the confusion.

include/aws/http/private/h2_connection.h
TingDaoK: I was trying to send the window_update when the window size is too small, so that value will be used much more frequently. And the implementation is changed now, which makes this seems useless now. I'll remove it. 😅

source/h2_decoder.c
```diff
@@ -653,6 +653,10 @@ static int s_state_fn_prefix(struct aws_h2_decoder *decoder, struct aws_byte_cur
         frame->stream_id,
         frame->payload_len);
 
+    if (decoder->frame_in_progress.type == AWS_H2_FRAME_T_DATA) {
+        /* We invoke the on_data_begin here to report the whole payload size */
+        DECODER_CALL_VTABLE_STREAM_ARGS(decoder, on_data_begin, frame->payload_len);
+    }
```
graebm, line 659: [trivial] Let's handle this like we handle other states that have a begin() callback. Declare a new state that calls this function. Rather than having the common prefix() function call this DATA-specific callback

source/h2_stream.c
```diff
@@ -535,20 +536,46 @@ int aws_h2_stream_on_decoder_push_promise(struct aws_h2_stream *stream, uint32_t
...
 
-int aws_h2_stream_on_decoder_data(struct aws_h2_stream *stream, struct aws_byte_cursor data) {
+int aws_h2_stream_on_decoder_data_begin(struct aws_h2_stream *stream, uint32_t data_payload_len) {
     AWS_PRECONDITION_ON_CHANNEL_THREAD(stream);
 
+    if (stream->thread_data.window_size_self < 0 ||
+        (stream->thread_data.window_size_self == 0 && data_payload_len > 0)) {
+        /* We should not treat when the window size and data payload len are both 0 as an error */
+        s_send_rst_and_close_stream(stream, AWS_ERROR_HTTP_FLOW_CONTROL_ERROR);
+    }
```
graebm, line 546: **return** s_send_rst_and_close_stream(stream, AWS_ERROR_HTTP_FLOW_CONTROL_ERROR);

source/h2_stream.c
```diff
@@ -535,20 +536,46 @@ int aws_h2_stream_on_decoder_push_promise(struct aws_h2_stream *stream, uint32_t
         s_get_h2_connection(stream), promised_stream_id, AWS_H2_ERR_REFUSED_STREAM);
 }
 
-int aws_h2_stream_on_decoder_data(struct aws_h2_stream *stream, struct aws_byte_cursor data) {
+int aws_h2_stream_on_decoder_data_begin(struct aws_h2_stream *stream, uint32_t data_payload_len) {
     AWS_PRECONDITION_ON_CHANNEL_THREAD(stream);
 
+    if (stream->thread_data.window_size_self < 0 ||
+        (stream->thread_data.window_size_self == 0 && data_payload_len > 0)) {
```
graebm, line 543: It's probably OK to get a DATA frame with 0 payload_len even if the stream's window is < 0.
right?

source/h2_stream.c
```diff
@@ -535,20 +536,46 @@ int aws_h2_stream_on_decoder_push_promise(struct aws_h2_stream *stream, uint32_t
...
+    stream->thread_data.window_size_self += data_payload_len;
+    return AWS_OP_SUCCESS;
+}
+
+int aws_h2_stream_on_decoder_data_i(struct aws_h2_stream *stream, struct aws_byte_cursor data) {
+    AWS_PRECONDITION_ON_CHANNEL_THREAD(stream);
+
+    if (s_check_state_allows_frame_type(stream, AWS_H2_FRAME_T_DATA)) {
+        return s_send_rst_and_close_stream(stream, aws_last_error());
+    }
```
graebm, line 578: [trivial] remove the state check here. It's harmless in this case, but it's consistent with other _i() callbacks where it's not actually safe to call because the _begin() callback might have changed the stream's state.

/* Not calling s_check_state_allows_frame_type() here because we already checked
 * at start of DATA frame in aws_h2_stream_on_decoder_data_begin() */

tests/h2_test_helper.c
```diff
@@ -290,16 +300,28 @@ static int s_decoder_on_push_promise_end(uint32_t stream_id, bool malformed, voi
...
-
-    /* Pretend each on_data callback is a full DATA frame for the purposes of these tests */
     ASSERT_SUCCESS(s_begin_new_frame(decode_tester, AWS_H2_FRAME_T_DATA, stream_id, &frame));
+    return AWS_OP_SUCCESS;
+}
+
+static int s_decoder_on_data_i(uint32_t stream_id, struct aws_byte_cursor data, void *userdata) {
+    (void)stream_id;
+    struct h2_decode_tester *decode_tester = userdata;
+    struct h2_decoded_frame *frame = h2_decode_tester_latest_frame(decode_tester);
```
graebm, line 314: [trivial] assert that the latest frame is type DATA

source/h2_stream.c
graebm: nevermind, we talked and the other way is too confusing to be worth it

source/h2_decoder.c
graebm: nevermind, we talked and the other way is too confusing to be worth it

source/h2_stream.c
TingDaoK: I don't really know... But spec says "A sender MUST track the negative flow-control window and MUST NOT send new flow-controlled frames until it receives WINDOW_UPDATE frames that cause the flow-control window to become positive." MUST NOT send new flow-controlled frames

PR #234, opened 2020-04-20
Receive GOAWAY

TingDaoK:
*Issue #, if available:*

*Description of changes:*

Receive GOAWAY frame
- Stop creating and sending new streams.
- Mark the streams with higher ID than the last_stream_id indicated by the GOAWAY frame as completed with error to inform the user to retry them on a new connection.

TODO:
- Inform our user about the debug information in GOAWAY.


source/h2_connection.c
```diff
@@ -1111,6 +1120,51 @@ static struct aws_h2err s_decoder_on_window_update(uint32_t stream_id, uint32_t
...
+            ERROR,
+            connection,
+            "Peer GOAWAY frame has invalid last stream id %" PRIu32 ", which is higher than the one we received",
+            last_stream);
+        return aws_h2err_from_h2_code(AWS_H2_ERR_PROTOCOL_ERROR);
+    }
+    /* stop sending any new stream and making new request */
+    aws_atomic_store_int(&connection->synced_data.new_stream_error_code, AWS_ERROR_HTTP_GOAWAY_RECEIVED);
+    connection->thread_data.goaway_received = true;
+    connection->thread_data.goaway_received_error_code = error_code;
```
TingDaoK, line 1142: I am not reporting the error at s_handler_shutdown(), since we will report it when we receive the GOAWAY. So I just feel maybe we can just report it here, and it seems more clear.

source/h2_connection.c
```diff
@@ -1111,6 +1120,51 @@ static struct aws_h2err s_decoder_on_window_update(uint32_t stream_id, uint32_t
...
+    connection->thread_data.goaway_received_last_stream_id = last_stream;
+    CONNECTION_LOGF(
+        DEBUG,
+        connection,
+        "Peer sent a GOAWAY frame, with error code %" PRIu32 ". New streams will not be accepted.",
+        error_code);
+    /* Complete activated streams whose id is higher than last_stream, since they will not process by peer. We should
+     * treat them as they had never been created at all */
+    struct aws_hash_iter stream_iter = aws_hash_iter_begin(&connection->thread_data.active_streams_map);
+    while (!aws_hash_iter_done(&stream_iter)) {
```
TingDaoK, line 1152: Another way will be storing the highest ID we processed, iterating the ID instead of the hash map. I'm not sure which one will be better for real life.

tests/test_h2_client.c
```diff
@@ -1906,3 +1906,125 @@ TEST_CASE(h2_client_push_promise_automatically_rejected) {
     client_stream_tester_clean_up(&stream_tester);
     return s_tester_clean_up();
 }
+
+/* Test client receives the GOAWAY frame, stop creating new stream and complete the streams whose id are higher than the
+ * last stream id included in GOAWAY frame */
+TEST_CASE(h2_client_conn_receive_goaway) {
```
graebm, line 1912: Nice test!

source/h2_connection.c
TingDaoK: So, I feel like the connection->thread_data.goaway_received_error_code seems serving no purpose now. Should I delete it?

PR #235, opened 2020-04-21
Integration test

TingDaoK:
*Issue #, if available:*

-  elasticurl uses "h2;http/1.1" alpn string by default
        --http2 and --http1.1 flags force "h2" or "http/1.1" and error if they don't get it
-  add h2 tests to integration-testing/http_client_test.py
-  add add_net_test_case(tls_download_medium_file_http2)


integration-testing/http_client_test.py
```diff
@@ -77,10 +77,27 @@ def test_simple_post(self):
     def test_simple_download(self):
```
graebm, line 77: let's make the existing tests be explicitly "_h1" tests
modify the test names and pass "--http1.1"

bin/elasticurl/main.c
```diff
@@ -453,6 +461,10 @@ static void s_on_signing_complete(struct aws_http_message *request, int error_co
 
 static void s_on_client_connection_setup(struct aws_http_connection *connection, int error_code, void *user_data) {
     struct elasticurl_ctx *app_ctx = user_data;
+    if (aws_http_connection_get_version(connection) != app_ctx->required_http_version) {
+        fprintf(stderr, "Error. Connection we got is different from the required version.");
```
JonathanHenson, line 465: Error. The requested http version, %s, is not supported by the peer., ctx->alpn

bin/elasticurl/main.c
```diff
@@ -453,6 +461,10 @@ static void s_on_signing_complete(struct aws_http_message *request, int error_co
 
 static void s_on_client_connection_setup(struct aws_http_connection *connection, int error_code, void *user_data) {
     struct elasticurl_ctx *app_ctx = user_data;
+    if (aws_http_connection_get_version(connection) != app_ctx->required_http_version) {
```
JonathanHenson, line 464: how does required_http_version get set if user doesn't specify one explicitly?

TingDaoK: I forgot about this case.🤦‍ Fixed.

PR #236, opened 2020-04-21
Fix refcounting bug in h2_connection.c

graebm:
- aws_stream_activate() should be increment stream refcount, not the function that moves it to a thread.
    - rename `s_stream_activate()` -> `s_move_stream_to_thread()` so it's not confused with `aws_h2_stream_activate()`
- reduce number of times we load atomic `new_stream_error_code`.
    - moved atomics out of `synced_data` to reduce confusion

### DIFF
--- a/source/h2_connection.c
+++ b/source/h2_connection.c
@@ -183,8 +183,8 @@ static void s_stop(
 
     /* Even if we're not scheduling shutdown just yet (ex: sent final request but waiting to read final response)
      * we don't consider the connection "open" anymore so user can't create more streams */
-    aws_atomic_store_int(&connection->synced_data.new_stream_error_code, AWS_ERROR_HTTP_CONNECTION_CLOSED);
-    aws_atomic_store_int(&connection->synced_data.is_open, 0);
+    aws_atomic_store_int(&connection->atomic.new_stream_error_code, AWS_ERROR_HTTP_CONNECTION_CLOSED);
+    aws_atomic_store_int(&connection->atomic.is_open, 0);
 
     if (schedule_shutdown) {
         AWS_LOGF_INFO(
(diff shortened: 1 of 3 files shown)


source/h2_connection.c
```diff
@@ -1635,10 +1635,13 @@ static void s_cross_thread_work_task(struct aws_channel_task *task, void *arg, e
     } /* END CRITICAL SECTION */
 
     /* Process new pending_streams */
-    while (!aws_linked_list_empty(&pending_streams)) {
-        struct aws_linked_list_node *node = aws_linked_list_pop_front(&pending_streams);
-        struct aws_h2_stream *stream = AWS_CONTAINER_OF(node, struct aws_h2_stream, node);
-        s_activate_stream(connection, stream);
+    if (!aws_linked_list_empty(&pending_streams)) {
+        int new_stream_error_code = (int)aws_atomic_load_int(&connection->atomic.new_stream_error_code);
```
TingDaoK, line 1639: Question: Why we fetch the error code here? So when the connection closes while we are iterating the pending_streams, it will not be updated, right? To improve performance?

graebm: Yes, to improve performance. It's 1 atomic_load for the N streams being activated, vs N atomic_loads.

It's ok if we're not perfectly in sync here. Eventually, the connection shuts down and any streams remaining in `synced_data.pending_streams` or `thread_data.active_streams_map` are completed with `AWS_ERROR_HTTP_CONNECTION_CLOSED`. This function is running on the event-loop thread, and so are (almost) all the other things that might set the `atomic.new_stream_error_code`.

The **only other** thing that can set it is a user calling `aws_http_connection_close()`. So, if a user calls close() in the middle of this loop running, then the streams will still activate successfully, but most likely they'll fail when the connection shuts down over the next few ticks of the event-loop.

PR #237, opened 2020-04-24
implement "Transfer-Encoding: chunked" for PUT and POST requests on H…

czakian:
*Issue #, if available:*

*Description of changes:*
This change implements transfer encoding chunked for PUT and POST requests using HTTP/1.1. 
Note: this change depends on https://github.com/awslabs/aws-c-io/pull/272. 


include/aws/http/request_response.h
```diff
@@ -296,6 +296,60 @@ struct aws_http_request_handler_options {
...
+
+    /**
+     * The number of elements defined in the extensions array.
+     */
+    size_t num_extensions;
+
+    /**
+     * Holds the number of extensions already sent over the connection.
+     */
+    size_t sent_extensions;
```
graebm, line 333: rather keep implementation details like `sent_headers` out of this public struct.
Maybe move it to `aws_http1_stream_chunk` since that has other internal details like `node`.

include/aws/http/request_response.h
```diff
@@ -296,6 +296,60 @@ struct aws_http_request_handler_options {
     aws_http_on_stream_complete_fn *on_complete;
 };
 
+struct aws_http1_stream_chunk;
+
+/**
+ * Signature definition of the function which will be invoked, if provided by the caller, after a chunked transfer
+ *  encoding chunk has been written to an HTTP/1.1 connection.
+ */
+typedef void aws_http1_stream_write_chunk_complete_fn(struct aws_http1_stream_chunk *chunk, void *user_data);
```
graebm, line 305: I'm on the fence about passing `struct aws_http1_stream_chunk` here.
The only thing a completion callback ever *really* needs to pass is user_data, where the user can store whatever they need to know when the call completes.

The `chunk` has everything in it the user might want to know, but also internal detail stuff like `node`. I'd kind of rather just hide that struct in a private header as an implementation detail.

include/aws/http/request_response.h
```diff
@@ -296,6 +296,60 @@ struct aws_http_request_handler_options {
...
+    /**
+     * User provided data passed to the on_complete callback on its invocation.
+     */
+    void *user_data;
+};
+
+struct aws_http1_stream_chunk {
+    struct aws_input_stream *stream;
+    struct aws_http1_chunk_options options;
+    struct aws_linked_list_node node;
```
graebm, line 350: If you end up keeping `aws_http_1_stream_chunk` part of the public API, consider prefixing some variables with "internal_" or "private_" or "_detail" or grouping them like
```
    struct {
        size_t sent_extensions;
        struct aws_linked_list node;
    } private;
};
```

include/aws/http/request_response.h
```diff
@@ -546,6 +600,28 @@ struct aws_input_stream *aws_http_message_get_body_stream(const struct aws_http_
...
 
+/**
+ * Submit a chunk of data to a stream for writing to an outbound HTTP/1.1 connection using chunked transfer encoding.
+ * Note: The stream does not take ownership of allocated resources in the chunk, if any. The caller is responsible
+ * for managing these resources. The data inside the chunk is expected not to be modified by the caller until
+ * the on_complete callback inside the chunk structure is invoked (if provided by the caller). Upon invocation of
+ * on_complete, the caller may release or modify structures in the chunk.
+ * On successful submission of the chunk to the stream, AWS_OP_SUCCESS is returned and AWS_OP_ERR otherwise.
+ */
+AWS_HTTP_API int aws_http1_stream_write_chunk(struct aws_http_stream *stream, struct aws_http1_stream_chunk *chunk);
```
graebm, line 611: The usual approach we've taken is for the public function to take the `options`, and then we allocate internal book-keeping stuff (ex: `struct aws_http1_stream_chunk`) based on that. It's a little weird for us to ask the user to give us the stuff we'll need for our internal book-keeping.

It makes it challenging to change the internal implementation when too much is public

include/aws/http/private/h1_encoder.h
```diff
@@ -37,10 +49,23 @@ enum aws_h1_encoder_state {
     AWS_H1_ENCODER_STATE_DONE,
 };
 
+enum aws_h1_encoder_body_stream_state {
+    AWS_H1_ENCODER_STATE_CHUNK_INIT,
+    AWS_H1_ENCODER_STATE_CHUNK_SIZE,
+    AWS_H1_ENCODER_STATE_CHUNK_EXTENSION_SEMICOLON_TOKEN,
+    AWS_H1_ENCODER_STATE_CHUNK_EXTENSION_KEY,
+    AWS_H1_ENCODER_STATE_CHUNK_EXTENSION_EQUAL_TOKEN,
+    AWS_H1_ENCODER_STATE_CHUNK_EXTENSION_VALUE,
```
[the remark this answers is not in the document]
czakian: I was trying to copy as little memory as possible, but the size of  chunk extensions should normally be none or quite small. However, I do agree it will make the state machinery simpler and easier for the caller to. I will make this change.

include/aws/http/request_response.h
czakian: I agree. Will remove. If they caller needs anything from that struct, they can save the pointer to it themselves.

include/aws/http/request_response.h
czakian: gotcha.

source/h1_encoder.c
```diff
@@ -61,9 +63,20 @@ static int s_scan_outgoing_headers(
...
-            case AWS_HTTP_HEADER_TRANSFER_ENCODING:
-                AWS_LOGF_ERROR(AWS_LS_HTTP_STREAM, "id=static: Sending of chunked messages not yet implemented");
-                return aws_raise_error(AWS_ERROR_UNIMPLEMENTED);
+            case AWS_HTTP_HEADER_TRANSFER_ENCODING: {
+                if (0 == header.value.len) {
+                    AWS_LOGF_ERROR(AWS_LS_HTTP_STREAM, "id=static: Transfer-Encoding must include a valid value");
+                    return aws_raise_error(AWS_ERROR_HTTP_INVALID_HEADER_VALUE);
+                }
+                struct aws_byte_cursor trimmed_value = aws_strutil_trim_http_whitespace(header.value);
+                struct aws_byte_cursor chunked = aws_byte_cursor_from_c_str("chunked");
```
JonathanHenson, line 72: This can be predefined at static file scope.

source/h1_encoder.c
```diff
@@ -61,9 +63,20 @@ static int s_scan_outgoing_headers(
...
+                struct aws_byte_cursor chunked_match = {0};
```
JonathanHenson, line 73: this won't compile everywhere.

AWS_ZERO_STRUCT(chunked_match)

source/h1_encoder.c
```diff
@@ -319,6 +343,347 @@ int aws_h1_encoder_start_message(
     return AWS_OP_SUCCESS;
 }
 
+static int s_h1_encoder_process_content_length_body(struct aws_byte_buf *dst, struct aws_h1_encoder *encoder) {
+    while (true) {
```
JonathanHenson, line 347: struct aws_byte_cursor cursor = aws_byte_cursor_from_buf(dst)

while (cursor->len) {

     //do stuff to seek the cursor, e.g. call the read functions or aws_byte_cursor_advance())
}

source/h1_encoder.c
```diff
@@ -319,6 +343,347 @@ int aws_h1_encoder_start_message(
...
+}
+
+static bool s_write_crlf(struct aws_byte_buf *dst) {
+    char crlf[2] = {'\r', '\n'};
+    struct aws_byte_cursor crlf_cursor = aws_byte_cursor_from_array(&crlf, AWS_ARRAY_SIZE(crlf));
+    return aws_byte_buf_write_from_whole_cursor(dst, crlf_cursor);
+}
+
+static bool s_chunk_size_state(struct aws_byte_buf *dst, struct aws_h1_encoder *encoder, int *aws_op_result) {
+    AWS_ASSERT(dst);
```
JonathanHenson, line 431: Let's make these formal verification friendly by using AWS_PRECONDITION instead of asserts.

source/h1_encoder.c
```diff
@@ -319,6 +343,347 @@ int aws_h1_encoder_start_message(
...
+    sub_cursor.len = write_size;
+    sub_cursor.ptr = src->ptr;
+    if (AWS_UNLIKELY(AWS_OP_SUCCESS != aws_byte_buf_append(dst, &sub_cursor))) {
+        return false;
+    }
+    aws_byte_cursor_advance(src, write_size);
+    return 0 == src->len;
+}
+
+#define fall_through __attribute__((fallthrough))
```
JonathanHenson, line 465: quite likely, gcc and clang only, might have to move this to aws-c-common's platform macros.

include/aws/http/request_response.h
czakian: I have removed it from the public API.

include/aws/http/request_response.h
czakian: I now manage the internal book keeping entirely in h1_stream.c and h1_encoder.c. The user is now no longer aware of the chunk.

source/h1_encoder.c
czakian: I actually lifted this part from the code directly from what was previously there and didn't look at it. Will update.

source/h1_encoder.c
czakian: will do throughout.

source/h1_encoder.c
czakian: How would you feel about a goto label instead of the fallthrough for a portable solution? 

e.g.
...
switch(integer) {
    case 1:
       /*... stuff */
      goto two;
    case 2:
    two:
      /* ... more stuff */
    default:
        break;
}
...

source/h1_encoder.c
czakian: FIxed. I should mention that I used it because I saw its usage in h2_connection.c.

source/h1_encoder.c
czakian: I have updated this to use the remaining buffer len in the while.

include/aws/http/request_response.h
```diff
@@ -546,6 +587,20 @@ struct aws_input_stream *aws_http_message_get_body_stream(const struct aws_http_
 AWS_HTTP_API
 void aws_http_message_set_body_stream(struct aws_http_message *message, struct aws_input_stream *body_stream);
 
+/**
+ * Submit a chunk of data to a stream for writing to an outbound HTTP/1.1 connection using chunked transfer encoding.
+ * Note: The stream does not take ownership of allocated resources in the chunk, if any. The caller is responsible
+ * for managing these resources. The data inside the chunk is expected not to be modified by the caller until
+ * the on_complete callback inside the chunk structure is invoked (if provided by the caller). Upon invocation of
+ * on_complete, the caller may release or modify structures in the chunk.
```
graebm, line 595: update comment. Caller only needs to keep the `aws_input_stream` alive until on_complete

include/aws/http/request_response.h
```diff
@@ -546,6 +587,20 @@ struct aws_input_stream *aws_http_message_get_body_stream(const struct aws_http_
...
+ * On successful submission of the chunk to the stream, AWS_OP_SUCCESS is returned and AWS_OP_ERR otherwise.
+ */
+AWS_HTTP_API int aws_http1_stream_write_chunk(
+    struct aws_allocator *allocator,
```
graebm, line 599: don't need to pass `allocator`, just use the stream's existing allocator

include/aws/http/request_response.h
```diff
@@ -546,6 +587,20 @@ struct aws_input_stream *aws_http_message_get_body_stream(const struct aws_http_
...
+    struct aws_http_stream *stream,
+    struct aws_input_stream *chunk_data,
+    struct aws_http1_chunk_options *options);
```
graebm, line 602: [style] move the `aws_input_stream` into the options struct. In functions where we use an options struct, we put all the arguments in there, except the 1st argument (object that function is being called on).

source/h1_encoder.c
```diff
@@ -319,6 +344,309 @@ int aws_h1_encoder_start_message(
...
+    return aws_byte_buf_write_from_whole_cursor(dst, crlf_cursor);
+}
+
+bool write_chunk_size(struct aws_byte_buf *dst, struct aws_input_stream *chunk) {
+    AWS_PRECONDITION(dst);
+    AWS_PRECONDITION(aws_byte_buf_is_valid(dst));
+    AWS_PRECONDITION(chunk);
+    AWS_PRECONDITION((dst->capacity - dst->len) >= MAX_ASCII_HEX_CHUNK_STR_SIZE);
+    int64_t chunk_size = 0;
+    if (AWS_OP_SUCCESS != aws_input_stream_get_length(chunk, &chunk_size)) {
```
graebm, line 438: not all types of `aws_input_stream` support seeking, which the get_length() function relies on.
> a valid stream may return an error instead when there is not a good answer (socket stream, for example).

It would be more flexible to let the user pass the chunk size in through the options struct

source/h1_encoder.c
```diff
@@ -319,6 +344,309 @@ int aws_h1_encoder_start_message(
...
+static bool s_chunk_line_state(struct aws_byte_buf *dst, struct aws_h1_encoder *encoder) {
+    AWS_PRECONDITION(dst);
+    AWS_PRECONDITION(encoder);
+    AWS_PRECONDITION(encoder->message);
+    AWS_PRECONDITION(encoder->message->body_chunk);
+    AWS_PRECONDITION(aws_byte_buf_is_valid(dst));
+    AWS_PRECONDITION(aws_byte_buf_is_valid(&encoder->message->body_chunk->chunk_line));
+    bool completely_written = s_fill_byte_buffer(dst, &encoder->message->body_chunk->chunk_line_cursor);
+    if (completely_written) {
+        aws_byte_buf_clean_up(&encoder->message->body_chunk->chunk_line);
```
graebm, line 479: Not 100% sure: Are chunk_line, and anything else chunk-related, getting cleaned upeven if the connection comes to a halt and stuff never gets sent? We also need to be sure completion callbacks fire so users know that it's safe to clean up the aws_input_streams. You could add a test that queues up a bunch of chunks and then just closes the connection.

source/h1_encoder.c
```diff
@@ -319,6 +344,309 @@ int aws_h1_encoder_start_message(
...
+    AWS_PRECONDITION(dst);
+    AWS_PRECONDITION(aws_byte_buf_is_valid(dst));
+    AWS_PRECONDITION(encoder);
+    AWS_PRECONDITION(encoder->message);
+    AWS_PRECONDITION(encoder->message->body_chunk);
+    AWS_PRECONDITION(encoder->message->body_chunk->data);
+    AWS_PRECONDITION(aws_op_result);
+
+    size_t prev_len = dst->len;
+    if (AWS_UNLIKELY(aws_input_stream_read(encoder->message->body_chunk->data, dst))) {
```
graebm, line 495: are we checking that the we don't stream more or less than the size we wrote at the start of the chunk?

fwiw, I did like the aws_input_stream_read_n() function you'd added once upon a time

source/h1_encoder.c
```diff
@@ -319,6 +344,309 @@ int aws_h1_encoder_start_message(
...
+            encoder,
+            "Failed to query body stream status, error %d (%s)",
+            aws_last_error(),
+            aws_error_name(aws_last_error()));
+        *aws_op_result = AWS_OP_ERR;
+        return false;
+    }
+
+    if (status.is_end_of_stream) {
+        return write_crlf(dst);
```
graebm, line 523: will we be ok if the stream ends, and there's just 1 byte of capacity left in the buffer?
It seems ok to me, just want to here you say it's cool too

we'll call `write_crlf()` which will do an all-or-nothing write and return false
then we'll come back into this state in the future and hopefully aws_input_stream_read() will do nothing and we'll succeed writing CRLF. right?

source/h1_encoder.c
czakian: done.

include/aws/http/request_response.h
czakian: done.

include/aws/http/request_response.h
czakian: done.

source/h1_encoder.c
czakian: done. I added tests for incorrect stream length now that the user pass in an incorrect value.

include/aws/http/private/h1_encoder.h
```diff
@@ -82,6 +113,15 @@ int aws_h1_encoder_process(struct aws_h1_encoder *encoder, struct aws_byte_buf *
...
 
+void aws_h1_lock_stream_list(struct aws_http1_chunks *body_chunks);
+
+void aws_h1_unlock_stream_list(struct aws_http1_chunks *body_chunks);
+
+bool write_chunk_size(struct aws_byte_buf *dst, struct aws_input_stream *chunk);
+
+bool write_crlf(struct aws_byte_buf *dst);
+
+bool write_chunk_extension(struct aws_byte_buf *dst, struct aws_http1_chunk_extension *chunk_extension);
```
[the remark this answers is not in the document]
czakian: fixed.

source/h1_connection.c
```diff
@@ -835,6 +846,30 @@ static void s_on_channel_write_complete(
...
+static bool s_should_process_body(
+    const struct h1_connection *connection,
+    const struct aws_io_message *msg,
+    struct aws_h1_stream *outgoing_stream) {
+    AWS_LOGF_TRACE(AWS_LS_HTTP_CONNECTION, "id=%p: checking if should process body", (void *)&connection->base);
+    return aws_h1_encoder_is_message_in_progress(&connection->thread_data.encoder) &&
+           connection->thread_data.encoder.message->has_chunked_encoding_header &&
+           msg->message_data.capacity > msg->message_data.len &&
+           s_populate_outgoing_buffer(connection, outgoing_stream);
+}
```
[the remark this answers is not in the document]
czakian: I agree, it does properly belong there. I was handling the header include ordering between h1_stream.h and h1_encoder.h awkwardly, which is how I ended up in that place. I have have moved the majority of this logic into h1_encoder.c  and done some refactoring to make everything link.

source/h1_stream.c
```diff
@@ -15,27 +15,149 @@
...
+    struct aws_input_stream *chunk_data,
+    struct aws_http1_chunk_options *options) {
+    AWS_PRECONDITION(allocator);
+    AWS_PRECONDITION(stream_base);
+    AWS_PRECONDITION(chunk_data);
+    AWS_PRECONDITION(options);
+
+    struct aws_h1_stream *stream = AWS_CONTAINER_OF(stream_base, struct aws_h1_stream, base);
+    struct aws_http1_stream_chunk *chunk = aws_mem_acquire(allocator, sizeof(struct aws_http1_stream_chunk));
+    chunk->allocator = allocator;
```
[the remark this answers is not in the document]
czakian: done.

include/aws/http/request_response.h
czakian: Done.

include/aws/http/private/h1_encoder.h
```diff
@@ -82,6 +113,15 @@ int aws_h1_encoder_process(struct aws_h1_encoder *encoder, struct aws_byte_buf *
 AWS_HTTP_API
 bool aws_h1_encoder_is_message_in_progress(const struct aws_h1_encoder *encoder);
 
+void aws_h1_lock_stream_list(struct aws_http1_chunks *body_chunks);
+
+void aws_h1_unlock_stream_list(struct aws_http1_chunks *body_chunks);
```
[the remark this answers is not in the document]
czakian: Done.

source/h1_encoder.c
```diff
@@ -319,6 +344,309 @@ int aws_h1_encoder_start_message(
...
+    AWS_ZERO_ARRAY(ascii_hex_chunk_size_str);
+    snprintf(ascii_hex_chunk_size_str, sizeof(ascii_hex_chunk_size_str), "%zX", chunk_size);
+    return aws_byte_buf_write_from_whole_cursor(dst, aws_byte_cursor_from_c_str(ascii_hex_chunk_size_str));
+}
+
+bool write_chunk_extension(struct aws_byte_buf *dst, struct aws_http1_chunk_extension *chunk_extension) {
+    AWS_PRECONDITION(chunk_extension);
+    AWS_PRECONDITION(aws_byte_buf_is_valid(dst));
+    bool wrote_all = true;
+    wrote_all &= aws_byte_buf_write_u8(dst, ';');
```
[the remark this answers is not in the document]
czakian: yeah, so much easier to write the generator!

source/h1_encoder.c
czakian: Yep. I have added a test. Since we added internal allocation, there's several paths which now need to be cleaned up on error cases. I will have remedied these in the next commit update.

source/h1_encoder.c
czakian: I had not been checking it as the chunk size was taken from the input stream itself and would have been an internal API bug. Once the API is updated to expose an explicit size call, I am checking using the progress_bytes in the encoder to calculate the actual payload size written (excluding chunked encoding overhead and anything on the chunk line). Let me know if progress_bytes is the wrong place to keep that information.

source/h1_encoder.c
czakian: Correct, if the call to s_chunk_payload_state returns false, the state of the state machine is not advanced and the output stream task will reschedule the call on a new channel message. As long as the writecrlf() transactionally succeeds in writing everything or nothing, all will be well.

include/aws/http/private/h1_encoder.h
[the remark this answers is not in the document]
czakian: Done.

source/h1_encoder.c
```diff
@@ -319,6 +351,404 @@ int aws_h1_encoder_start_message(
...
+}
+
+int aws_chunk_line_from_options(struct aws_http1_chunk_options *options, struct aws_byte_buf *chunk_line) {
+    size_t chunk_line_size = s_calculate_chunk_line_size(options);
+    if (AWS_OP_SUCCESS != aws_byte_buf_init(chunk_line, options->chunk_data->allocator, chunk_line_size)) {
+        return AWS_OP_ERR;
+    }
+    if (!s_populate_chunk_line_buffer(chunk_line, options)) {
+        /* The caller will free the chunk, which will free the chunk line in the process */
+        return AWS_OP_ERR;
```
graebm, line 696: returning AWS_OP_ERR without any aws_raise_error() in the callstack.
Seems like this can only go wrong if s_calculate_chunk_line_size() miscalculated. If you are extremely confident in that calculation, you could simply AWS_ASSERT() that it wrote all. If you think it might have edge-cases you could log the error and `return aws_raise_error(AWS_ERROR_INVALID_STATE)`

source/h1_connection.c
```diff
@@ -859,6 +870,11 @@ static void s_outgoing_stream_task(struct aws_channel_task *task, void *arg, enu
         return;
     }
 
+    if (aws_h1_stream_is_paused(outgoing_stream)) {
+        AWS_LOGF_TRACE(AWS_LS_HTTP_STREAM, "id=%p: Pausing outgoing stream task", (void *)outgoing_stream);
```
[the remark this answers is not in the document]
czakian: done.

source/h1_encoder.c
```diff
@@ -319,6 +351,404 @@ int aws_h1_encoder_start_message(
...
+    AWS_PRECONDITION(chunk_out);
+    bool has_next_chunk = true;
+    /* Begin critical section */
+    aws_h1_lock_chunked_list(body_chunks);
+    if (!aws_linked_list_is_valid(&body_chunks->chunk_list)) {
+        has_next_chunk = false;
+        goto release_lock;
+    }
+    if (aws_linked_list_empty(&body_chunks->chunk_list)) {
+        *chunk_out = NULL;
```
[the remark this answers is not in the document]
czakian: It's purely a sanity check. As discussed offline, it is ok to do the ASSERT inside a mutex as they get compiled out in release. The reason I wrote it that way is I didn't want to hold the lock on error.

source/h1_encoder.c
```diff
@@ -61,9 +63,26 @@ static int s_scan_outgoing_headers(
...
+                    AWS_LOGF_ERROR(AWS_LS_HTTP_STREAM, "id=static: Transfer-Encoding must include a valid value");
+                    return aws_raise_error(AWS_ERROR_HTTP_INVALID_HEADER_VALUE);
+                }
+                struct aws_byte_cursor trimmed_value = aws_strutil_trim_http_whitespace(header.value);
+                struct aws_byte_cursor chunked = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("chunked");
+                if (trimmed_value.len >= chunked.len) {
+                    size_t expected_chunked_value_location = trimmed_value.len - chunked.len;
+                    aws_byte_cursor_advance(&trimmed_value, expected_chunked_value_location);
+                    encoder_message->has_chunked_encoding_header =
+                        aws_byte_cursor_eq_ignore_case(&chunked, &trimmed_value);
```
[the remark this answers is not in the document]
czakian: Good point. done.

source/h1_connection.c
```diff
@@ -923,6 +940,13 @@ static void s_outgoing_stream_task(struct aws_channel_task *task, void *arg, enu
         aws_mem_release(msg->allocator, msg);
     }
 
+    /* The rest of the pending chunks will be taken care of when the http stream is destroyed.
+     * Only need to clean up the current one being processed here.*/
+    struct aws_http1_stream_chunk *cur_chunk = connection->thread_data.encoder.message->body_chunk;
+    if (cur_chunk) {
+        aws_h1_stream_release_chunk(cur_chunk);
```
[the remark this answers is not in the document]
czakian: Good point - I hadn't considered the case where the connection was paused, the socket timed out, and then the clean up of the stream was invoked. I have moved all memory clean up into two places: 
1. in the happy case, the current chunk is cleaned up when the chunk transitions into into the "end" or "terminated" state. 
2. In the unhappy case, when the stream is cleaned up. 

I have added asserts to aws_h1_populate_current_stream_chunk to ensure that the caller has nulled out the current chunk before invoking that call which should help catch a programming error later on where the writer forgets to clean up and null out the current chunk before trying to get the next one.

I considered just letting the clean up happen in aws_h1_populate_current_stream_chunk(), but I decided not to do that as it leaves the chunk memory hanging around until the stream clean up or the next chunk is called. On units with memory pressure, this could cause the memory to hang around until the stream is cleaned up on the final terminate call, which doesn't seem ideal. The trade off is that the could would only have two places where memory gets freed instead of three at the expense of keeping some memory around longer.

source/h1_encoder.c
```diff
@@ -319,6 +351,404 @@ int aws_h1_encoder_start_message(
...
+            "Failed to write body stream, error %d (%s)",
+            aws_last_error(),
+            aws_error_name(aws_last_error()));
+        *aws_op_result = AWS_OP_ERR;
+        return false;
+    }
+
+    const int64_t amount_read = dst->len - prev_len;
+    encoder->progress_bytes += amount_read;
+    ENCODER_LOGF(TRACE, encoder, "Wrote %zu body bytes to message", amount_read);
```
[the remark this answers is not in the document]
czakian: Done.

source/h1_stream.c
```diff
@@ -32,10 +33,86 @@ static void s_stream_update_window(struct aws_http_stream *stream, size_t increm
     aws_http_connection_update_window(stream->owning_connection, increment_size);
 }
 
+static bool s_body_chunks_init(struct aws_h1_stream *stream) {
```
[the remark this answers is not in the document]
czakian: done.

source/h1_stream.c
```diff
@@ -32,10 +33,86 @@ static void s_stream_update_window(struct aws_http_stream *stream, size_t increm
...
+}
+
+static int s_aws_h1_stream_write_chunk(struct aws_http_stream *stream_base, struct aws_http1_chunk_options *options) {
+    AWS_PRECONDITION(stream_base);
+    AWS_PRECONDITION(options);
+    struct aws_h1_stream *stream = AWS_CONTAINER_OF(stream_base, struct aws_h1_stream, base);
+    const size_t chunk_alloc_size = sizeof(struct aws_http1_stream_chunk);
+    struct aws_http1_stream_chunk *chunk = aws_mem_acquire(options->chunk_data->allocator, chunk_alloc_size);
+    AWS_ASSERT(chunk);
+    AWS_ZERO_STRUCT(*chunk);
```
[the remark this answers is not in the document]
czakian: done.

source/h1_encoder.c
czakian: I'll move it to an assert. The size is calculated internally and the buffer is allocated off of it, so it would be a programming error to miscalculate the actual written size.

source/h1_encoder.c
```diff
@@ -77,6 +96,15 @@ static int s_scan_outgoing_headers(
...
     }
+
+    /* Per RFC 2656 (https://tools.ietf.org/html/rfc2616#section-4.4), if both the Content-Length and Transfer-Encoding
+     * header are defined, the client should not send the request. */
+    if (encoder_message->has_chunked_encoding_header && has_content_length_header) {
+        AWS_LOGF_ERROR(
+            AWS_LS_HTTP_STREAM, "id=static: Both Content-Length and Transfer-Encoding are set. Only one may be used");
+        return aws_raise_error(AWS_ERROR_HTTP_INVALID_HEADER_VALUE);
+    }
+
```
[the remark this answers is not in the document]
czakian: done.

source/h1_encoder.c
```diff
@@ -77,6 +91,15 @@ static int s_scan_outgoing_headers(
             return AWS_OP_ERR;
         }
     }
+
+    /* Per RFC 2656 (https://tools.ietf.org/html/rfc2616#section-4.4), if both the Content-Length and Transfer-Encoding
+     * header are defined, the client should not send the request. */
+    if (encoder_message->has_chunked_encoding_header && has_content_length_header) {
```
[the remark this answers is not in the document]
czakian: fixed.

source/h1_connection.c
```diff
@@ -870,6 +870,9 @@ static void s_outgoing_stream_task(struct aws_channel_task *task, void *arg, enu
         return;
     }
 
+    /* The s_outgoing_stream_task() will be resumed when a new chunk is added and the stream is paused.
```
graebm, line 873: trivial: "unpaused"

source/h1_encoder.c
```diff
@@ -691,17 +720,13 @@ int aws_chunk_line_from_options(struct aws_http1_chunk_options *options, struct
     if (AWS_OP_SUCCESS != aws_byte_buf_init(chunk_line, options->chunk_data->allocator, chunk_line_size)) {
         return AWS_OP_ERR;
     }
-    if (!s_populate_chunk_line_buffer(chunk_line, options)) {
-        /* The caller will free the chunk, which will free the chunk line in the process */
-        return AWS_OP_ERR;
-    }
+    AWS_ASSERT(s_populate_chunk_line_buffer(chunk_line, options));
```
graebm, line 723: This whole thing will get compiled out in release. You'll want to do one of these annoying things:
```
bool populate_chunk_line_success = s_populate_chunk_line_buffer(chunk_line, options);
AWS_ASSERT(populate_chunk_line_success);
(void)populate_chunk_line_success; <-- avoid compiler warnings about unused variables
```
or
```
if (!s_populate_chunk_line_buffer(chunk_line, options)) {
    AWS_ASSERT(0);
}
```

source/h1_stream.c
```diff
@@ -32,10 +33,94 @@ static void s_stream_update_window(struct aws_http_stream *stream, size_t increm
...
+        return AWS_OP_ERR;
+    }
+    chunk->data = options->chunk_data;
+    chunk->data_size = options->chunk_data_size;
+    chunk->on_complete = options->on_complete;
+    chunk->user_data = options->user_data;
+    aws_linked_list_node_reset(&chunk->node);
+    if (AWS_OP_SUCCESS != aws_chunk_line_from_options(options, &chunk->chunk_line)) {
+        s_clean_up_body_chunk(&chunk);
+        return AWS_OP_ERR;
```
graebm, line 98: set `on_complete` after calling aws_chunk_line_from_options().

That way, if something goes wrong with aws_chunk_line_from_options(), we're not firing the on-complete callback. Contract is: on_complete only fires if we successfully create the chunk

source/h1_connection.c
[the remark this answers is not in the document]
graebm: 👍

source/h1_connection.c
czakian: done.

source/h1_encoder.c
czakian: done.

source/h1_stream.c
czakian: yep, gotcha. my thought here was that is simplified the client handling because they could write one path to free resources, but this is cool too.

czakian:
Will follow instructions and submit a new PR

PR #238, opened 2020-04-27
Switched headers to using a small block allocator

justinboswell:


source/request_response.c
```diff
@@ -60,7 +61,11 @@ struct aws_http_headers *aws_http_headers_new(struct aws_allocator *allocator) {
         goto alloc_failed;
     }
 
-    headers->alloc = allocator;
+    headers->parent_alloc = allocator;
+    headers->alloc = aws_small_block_allocator_new(allocator, false);
```
graebm, line 65: [trivial] maybe keep the existing `headers->alloc` as it was, and name the new one `headers->string_alloc` or `headers->small_block_alloc`.

justinboswell: Then I have to re-write everything, I was trying to do this quickly and non-intrusively :)

justinboswell: Now that we've proven it works, I'll fix it!

PR #239, opened 2020-04-30
Connection setup callback from connection manager is now guaranteed t…

JonathanHenson:
…o run from the connection's thread, deleted some tests that made invalid assumptions about data access patterns.


source/connection_manager.c
```diff
@@ -310,14 +312,43 @@ static bool s_aws_http_connection_manager_should_destroy(struct aws_http_connect
...
+    void *arg,
+    enum aws_task_status status) {
+    (void)channel_task;
+
+    struct aws_http_connection_acquisition *pending_acquisition = arg;
+
+    if (status != AWS_TASK_STATUS_RUN_READY) {
+        AWS_LOGF_WARN(
+            AWS_LS_HTTP_CONNECTION_MANAGER,
+            "id=%p: Failed to completed connection acquisition because the connection was closed",
```
singku-china, line 336: Failed to `complete`

source/connection_manager.c
```diff
@@ -335,25 +366,40 @@ static void s_aws_http_connection_manager_complete_acquisitions(
         struct aws_http_connection_acquisition *pending_acquisition =
             AWS_CONTAINER_OF(node, struct aws_http_connection_acquisition, node);
 
-        pending_acquisition->callback(
-            pending_acquisition->connection, pending_acquisition->error_code, pending_acquisition->user_data);
-
         if (pending_acquisition->error_code != 0) {
             AWS_LOGF_WARN(
                 AWS_LS_HTTP_CONNECTION_MANAGER,
                 "id=%p: Failed to completed connection acquisition with error_code %d(%s)",
```
singku-china, line 372: Here typo as well

source/connection_manager.c
```diff
@@ -335,25 +366,40 @@ static void s_aws_http_connection_manager_complete_acquisitions(
...
                 (void *)pending_acquisition->manager,
                 pending_acquisition->error_code,
                 aws_error_str(pending_acquisition->error_code));
+            pending_acquisition->callback(
```
singku-china, line 376: Can you guarantee here that `aws_channel_thread_is_callers_thread`?

JonathanHenson: we don't actually care in this case.

source/connection_manager.c
```diff
@@ -335,25 +366,40 @@ static void s_aws_http_connection_manager_complete_acquisitions(
         struct aws_http_connection_acquisition *pending_acquisition =
             AWS_CONTAINER_OF(node, struct aws_http_connection_acquisition, node);
 
-        pending_acquisition->callback(
-            pending_acquisition->connection, pending_acquisition->error_code, pending_acquisition->user_data);
-
         if (pending_acquisition->error_code != 0) {
```
bretambrose, line 369: Could we restructure this a little to save a little duplication:

```
if (error_code == 0 && connection channel slot all valid && aws_channel_thread_is_callers_thread()) {
   submit task
   return
}

do callback
mem release
```

source/connection_manager.c
```diff
@@ -310,14 +312,43 @@ static bool s_aws_http_connection_manager_should_destroy(struct aws_http_connect
...
+    (void)channel_task;
+
+    struct aws_http_connection_acquisition *pending_acquisition = arg;
+
+    if (status != AWS_TASK_STATUS_RUN_READY) {
+        AWS_LOGF_WARN(
+            AWS_LS_HTTP_CONNECTION_MANAGER,
+            "id=%p: Failed to completed connection acquisition because the connection was closed",
+            (void *)pending_acquisition->manager);
+        pending_acquisition->callback(NULL, AWS_ERROR_HTTP_CONNECTION_CLOSED, pending_acquisition->user_data);
```
bretambrose, line 338: What are the conditions when this would happen?  I can only see an event loop shutdown being the trigger and if that's happening (while the connection manager is still alive and it is alive because this is a vended connection at this point, preventing connection manager shutdown), then we're in a world of hurt.  

I'm not able to see the link between the task being in a cancelled state and the connection being closed.

I am also worried that this leaks the connection.  The connection was good before the task was scheduled and the connection manager has already incremented its vended count at that time.  If we don't actually end up vending the connection then we need to perform the proper fix up here and possibly release the connection back to the pool.

JonathanHenson: It’s a channel task though

JonathanHenson: This would only happen if the channel was shutdown, which simply means, the connection closed on you. These are channel tasks, not event-loop tasks.

Good catch on releasing the connection back to the pool though.

PR #240, opened 2020-05-01
Rhel5 build fixes

bretambrose:


source/hpack.c
```diff
@@ -1261,9 +1261,7 @@ int aws_hpack_decode(
                 goto handle_complete;
             } break;
 
-            default: {
```
TingDaoK, line 1264: When I run Format Document at windows machine, this will happen. Undo this change will pass the Lint for format...

bretambrose: Yeah, for some reason, the clang format on my machine is slightly out of sync with what's in CI

PR #244, opened 2020-05-06
potential bug for ping_ack

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/h2_connection.c
```diff
@@ -636,17 +636,19 @@ static int s_encode_outgoing_frames_queue(struct aws_h2_connection *connection,
 
     /* Write as many frames from outgoing_frames_queue as possible. */
     while (!aws_linked_list_empty(outgoing_frames_queue)) {
-        struct aws_linked_list_node *frame_node = aws_linked_list_front(outgoing_frames_queue);
-        struct aws_h2_frame *frame = AWS_CONTAINER_OF(frame_node, struct aws_h2_frame, node);
-
+        if (!connection->thread_data.on_going_frame) {
+            struct aws_linked_list_node *frame_node = aws_linked_list_pop_front(outgoing_frames_queue);
+            connection->thread_data.on_going_frame = AWS_CONTAINER_OF(frame_node, struct aws_h2_frame, node);
```
graebm, line 641: If we're going to pop from the queue upon setting the current frame, then we need to change more logic:
1) This loop needs to be `while(!outgoing_frames_queue.empty() || on_going_frame)`. There are other places we check if outgoing_frames_queue is empty, those would need to change too. We probably want a helper function s_has_outgoing_frames().
2) During shutdown/destroy when we clean up all the frames on outgoing_frames_queue we also need to clean up the current frame

OR 

We keep the current behavior and don't pop the frame until we're done with it. We just always have the current frame point at the frame at the front of the queue. The insertion logic would have to change, to check current frame, though

PR #246, opened 2020-05-13
User HTTP/2 change settings API

TingDaoK:
*Issue #, if available:*

*Description of changes:*


include/aws/http/connection.h
```diff
@@ -221,11 +228,44 @@ struct aws_http_client_connection_options {
     bool manual_window_management;
 };
 
+/**
+ * Options for changing the settings for an HTTP/2 connection.
+ */
+struct aws_http2_change_settings_options {
```
TingDaoK, line 234: I don't know we really need this structure or not... Will this make the logic too complicated? Without this, the parameters for aws_http2_connection_change_settings will be complicated...

source/h2_connection.c
```diff
@@ -310,6 +313,31 @@ static struct aws_h2_connection *s_connection_new(
...
+    /* Initial settings with initial window size and disable push promise for the client */
+    struct aws_h2_frame_setting initial_settings[2];
+    size_t iter = 0;
+    if (!server) {
+        initial_settings[iter].id = AWS_H2_SETTINGS_ENABLE_PUSH;
+        initial_settings[iter++].value = 0;
+    }
+    if (initial_window_size != aws_h2_settings_initial[AWS_H2_SETTINGS_INITIAL_WINDOW_SIZE]) {
+        initial_settings[iter].id = AWS_H2_SETTINGS_INITIAL_WINDOW_SIZE;
+        initial_settings[iter++].value = initial_window_size;
```
TingDaoK, line 325: We will not correct this initial_window_size for user, if the user gives us a number that is too big. Since I think user will need that value for manually flow control. But, we can probably assert/precondition to make sure this value will be among the required bounds.

include/aws/http/private/connection_impl.h
```diff
@@ -52,6 +52,9 @@ struct aws_http_connection_vtable {
     void (*close)(struct aws_http_connection *connection);
     bool (*is_open)(const struct aws_http_connection *connection);
     void (*update_window)(struct aws_http_connection *connection, size_t increment_size);
+
+    /* HTTP/2 specific functions */
+    int (*change_settings)(struct aws_http_connection *connection, const struct aws_http2_change_settings_options *opt);
```
TingDaoK, line 57: We use *base approach instead of *impl... So from my understanding, this will be the "way".

include/aws/http/connection.h
graebm: Yeah, these should probably just be function arguments.
`_options` structs are good for functions where you think it's very likely that we'll add more arguments in the future. This seems like it has all that it needs

include/aws/http/connection.h
```diff
@@ -56,6 +56,13 @@ typedef void(
 typedef void(
     aws_http_on_client_connection_shutdown_fn)(struct aws_http_connection *connection, int error_code, void *user_data);
 
+/**
+ * Invoked when the HTTP/2 connection settings ack frame received.
+ * It can be not invoked if we never received the settings ack frame for that settings.
+ * This is always invoked on connetion's event-loop thread.
+ */
+typedef int(aws_http2_on_settings_ack_received)(void *user_data);
```
graebm, line 64: Suggestion:
`typedef void (aws_http2_on_change_settings_complete)(struct aws_http_connection *connection, int error_code, void *user_data)`

Removed `int` return type. Most callbacks, and especially "completion" callbacks, shouldn't affect the thing calling them. I know that *most* of the stream callbacks have an `int` return type, but these are a special case. A long series of stream callbacks that all must succeed : an exception from the user while receiving headers, generally means the user is not in a position to receive the body either. The stream's completion callback, which fires last, and is guaranteed to fire, has a `void` return type.

Also, I would ALWAYS fire the callback. It's way easier to hook these up in bindings when we know they will fire exactly once: we can give the user some "future" type that they can wait on. If we never receive the ACK, we can fire it with an `AWS_ERROR_HTTP_CONNECTION_CLOSED` error_code when tearing things down and cleaning up the pending_settings list.

source/h2_connection.c
```diff
@@ -1727,6 +1742,22 @@ static void s_cross_thread_work_task(struct aws_channel_task *task, void *arg, e
...
+    }
+
+    /* Move pending settings to thread data */
+    if (!aws_linked_list_empty(&pending_settings)) {
+        while (!aws_linked_list_empty(&pending_settings)) {
+            aws_linked_list_push_back(
+                &connection->thread_data.pending_settings_queue, aws_linked_list_pop_front(&pending_settings));
+        }
+    }
     /* #TODO: process stuff from other API calls (ex: window-updates) */
```
graebm, line 1761: trivial: you can remove this TODO, we're living the dream now

include/aws/http/private/h2_connection.h
```diff
@@ -128,6 +134,15 @@ struct aws_h2_connection {
     } atomic;
 };
 
+struct aws_h2_pending_settings {
+    struct aws_h2_frame_setting *settings_array;
+    size_t num_settings;
```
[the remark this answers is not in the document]
graebm: nevermind, you need this

include/aws/http/connection.h
```diff
@@ -221,11 +228,44 @@ struct aws_http_client_connection_options {
...
 
+/**
+ * Options for changing the settings for an HTTP/2 connection.
+ */
+struct aws_http2_change_settings_options {
+    /**
+     * Required
+     * The data of settings to chagne.
+     */
+    const struct aws_h2_frame_setting *settings_array;
```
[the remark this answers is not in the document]
TingDaoK: Yeah... A sequence of change I worried about when I asked you about it... But, your suggestion seems good.

source/h2_connection.c
```diff
@@ -310,6 +322,31 @@ static struct aws_h2_connection *s_connection_new(
...
+    }
+    if (initial_window_size != aws_h2_settings_initial[AWS_H2_SETTINGS_INITIAL_WINDOW_SIZE]) {
+        initial_settings[iter].id = AWS_H2_SETTINGS_INITIAL_WINDOW_SIZE;
+        initial_settings[iter++].value = (uint32_t)initial_window_size;
+    }
+    struct aws_http2_change_settings_options options;
+    AWS_ZERO_STRUCT(options);
+    options.num_settings = iter;
+    options.settings_array = initial_settings;
+    if (aws_http2_connection_change_settings(&connection->base, &options)) {
```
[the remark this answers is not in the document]
TingDaoK: Yeah, I was thinking about "thread_data will only be touched at eventloop thread, but this is the main thread". But, we just initialized it, it's fine to just enqueue it.

include/aws/http/connection.h
```diff
@@ -286,6 +315,18 @@ enum aws_http_version aws_http_connection_get_version(const struct aws_http_conn
 AWS_HTTP_API
 struct aws_channel *aws_http_connection_get_channel(struct aws_http_connection *connection);
 
+/**
+ * HTTP/2 specific. Change the HTTP/2 conenction settings.
+ * Settings frame will be sent. If set, on_completed callback will always be invoked.
+ */
```
graebm, line 321: If set, on_completed callback will be always be invoked, if no error is reported then the peer has acknowledged the settings.

source/h2_connection.c
```diff
@@ -1259,21 +1283,36 @@ static struct aws_h2err s_decoder_on_settings_ack(void *userdata) {
...
                 aws_h2_decoder_set_setting_max_frame_size(decoder, settings_array[i].value);
             } break;
+            default:
+                break;
         }
         connection->thread_data.settings_self[settings_array[i].id] = settings_array[i].value;
     }
+    /* invoke the change settings compeleted user callback */
+    if (pending_settings->on_completed) {
+        pending_settings->on_completed(&connection->base, AWS_H2_ERR_NO_ERROR, pending_settings->user_data);
```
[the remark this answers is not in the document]
TingDaoK: Oh, I was trying to find that AWS error code... We got error_code defined in several files... Anyway, I was a little bit lost...

include/aws/http/connection.h
```diff
@@ -56,6 +56,13 @@ typedef void(
 typedef void(
     aws_http_on_client_connection_shutdown_fn)(struct aws_http_connection *connection, int error_code, void *user_data);
 
+/**
+ * Invoked when the HTTP/2 settings change completes, whether successful or unsuccessful.
+ * This is always invoked on connection's event-loop thread.
+ */
+typedef void(
```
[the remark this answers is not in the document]
TingDaoK: It's unlikely but possible that error happens when we actually apply the settings. Like: 
```
change to SETTINGS_INITIAL_WINDOW_SIZE from internal caused a stream's flow-control window to exceed the maximum size
```
So, probably:
```
/**
 * Invoked when the HTTP/2 settings change is complete, whether successful or unsuccessful.
 * If error_code is 0, then the peer has acknowledged the settings change.
 * If error_code is non-zero, then an error occurred before the settings change applied.
 * This is always invoked on the connection's event-loop thread.
 */
```
Oh man, my POOR English...

include/aws/http/connection.h
TingDaoK: Same as above, `If set, on_completed callback will be always be invoked, if no error is reported then the peer has acknowledged the settings and the change successfully applied.`

include/aws/http/connection.h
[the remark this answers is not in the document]
graebm:
```
/**
 * Invoked when the HTTP/2 settings change is complete, whether successful or unsuccessful.
 * If error_code is AWS_ERROR_SUCCESS (0), then the peer has acknowledged
 * the settings and the change has been applied.
 * If error_code is non-zero, then a connection error occurred before the
 * settings could be fully acknowledged and applied.
 * This is always invoked on the connection's event-loop thread.
 */

```

PR #249, opened 2020-05-15
User ping api

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- User can request PING now


source/h2_connection.c
```diff
@@ -1170,6 +1213,57 @@ static struct aws_h2err s_decoder_on_rst_stream(uint32_t stream_id, uint32_t h2_
...
+        return aws_h2err_from_h2_code(AWS_H2_ERR_PROTOCOL_ERROR);
+    }
+    struct aws_h2err err;
+    struct aws_linked_list_node *node = aws_linked_list_pop_front(&connection->thread_data.pending_ping_queue);
+    struct aws_h2_pending_ping *pending_ping = AWS_CONTAINER_OF(node, struct aws_h2_pending_ping, node);
+    /* Check the payload */
+    if (memcmp(opaque_data, pending_ping->opaque_data, AWS_H2_PING_DATA_SIZE) != 0) {
+        CONNECTION_LOG(
+            ERROR, connection, "Connection error, received PING ACK has different opaque data with our expectation.");
+        err = aws_h2err_from_h2_code(AWS_H2_ERR_PROTOCOL_ERROR);
```
TingDaoK, line 1229: This error and the extra PING ACK are not defined in spec? OR I missed some part of spec again...

source/h2_connection.c
```diff
@@ -1777,6 +1871,13 @@ static void s_cross_thread_work_task(struct aws_channel_task *task, void *arg, e
         }
     }
 
+    /* Move pending PING to thread data */
+    if (!aws_linked_list_empty(&pending_ping)) {
+        while (!aws_linked_list_empty(&pending_ping)) {
```
TingDaoK, line 1876: We can probably just push the front of pending_ping to the end of the pending_ping_queue, right? It will be more efficient.

source/h2_connection.c
```diff
@@ -1926,6 +2027,55 @@ static int s_connection_change_settings(
...
 
+static int s_connection_ping(
+    struct aws_http_connection *connection_base,
+    const uint8_t *opaque_data,
+    void *user_data,
+    aws_http2_on_ping_complete_fn *on_completed) {
+
+    struct aws_h2_connection *connection = AWS_CONTAINER_OF(connection_base, struct aws_h2_connection, base);
+    uint64_t time_stamp;
+    if (aws_sys_clock_get_ticks(&time_stamp)) {
```
TingDaoK, line 2038: This is discussable, I put it here just for simplicity.

tests/CMakeLists.txt
```diff
@@ -399,6 +399,8 @@ add_test_case(h2_client_conn_err_invalid_last_stream_id_goaway)
 #TODO add_test_case(h2_client_send_goaway_with_push_promises) id of 1st should be in GOAWAY 2nd should be ignored
 add_test_case(h2_client_change_settings_succeed)
 add_test_case(h2_client_change_settings_failed_no_ack_received)
+add_test_case(h2_client_send_ping_successfully_receive_ack)
+add_test_case(h2_client_send_ping_no_ack_received)
```
graebm, line 403: implement, or `#TODO`
h2_client_conn_err_extraneous_ping_ack_received
h2_client_conn_err_mismatched_ping_ack_received

source/h2_connection.c
```diff
@@ -1777,6 +1869,13 @@ static void s_cross_thread_work_task(struct aws_channel_task *task, void *arg, e
         }
     }
 
+    /* Move pending PING to thread data */
+    if (!aws_linked_list_empty(&pending_ping)) {
+        while (!aws_linked_list_empty(&pending_ping)) {
```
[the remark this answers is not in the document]
TingDaoK: Ugh, I probably add this accidentally while I made the list-move comments. Will remove.

source/h2_connection.c
TingDaoK: Lol, I did the same thing. Searching the whole spec for "ping".

include/aws/http/connection.h
```diff
@@ -330,6 +343,19 @@ int aws_http2_connection_change_settings(
...
+/**
+ * HTTP/2 specific. Send a PING frame with opaque_data as payload.
+ * Opaque_data has to be 8 bytes data or NULL. If opaque_data is NULL, we will send zeroed 8 bytes data.
+ * If set, on_completed callback will always be invoked.
+ */
+AWS_HTTP_API
+int aws_http2_connection_ping(
+    struct aws_http_connection *connection,
+    const uint8_t *opaque_data,
+    size_t data_len,
```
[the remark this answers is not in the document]
TingDaoK: Oh I thought about `const uint8_t opaque_data[AWS_H2_PING_DATA_SIZE]`. But I don't know how it is different from just a pointer as a parameter... Like the function can still accept a pointer, right?

source/h2_connection.c
TingDaoK: If performance will not matter for our use case, let's go with the existing one for now, I may add those helpers after the window_update.

include/aws/http/connection.h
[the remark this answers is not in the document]
graebm: yeah, passing it like that is the same thing at runtime as passing a pointer. But at compile-time it enforces that the user is passing an array of the exact size. It's cute, but not often used.

include/aws/http/connection.h
```diff
@@ -327,8 +340,20 @@ int aws_http2_connection_change_settings(
...
     const struct aws_http2_setting *settings_array,
     size_t num_settings,
-    void *user_data,
-    aws_http2_on_change_settings_complete_fn *on_completed);
+    aws_http2_on_change_settings_complete_fn *on_completed,
+    void *user_data);
+
+/**
+ * HTTP/2 specific. Send a PING frame with opaque_data as payload.
+ * Opaque_data has to be 8 bytes data or NULL. If opaque_data is NULL, we will send zeroed 8 bytes data.
```
graebm, line 348: trivial: let's start doing doxygen-style. Makes it easy to document the specific details of individual parameters. Some day, if we ever publish docs for aws-c-*** these will become doxygen style

```
/**
 * Send a PING frame (HTTP/2 only).
 * Round-trip-time is calculated when PING ACK is received from peer.
 *
 * @param http2_connection HTTP/2 connection.
 * @param optional_opaque_data Optional payload for PING frame.
 *      Must be NULL, or exactly 8 bytes (AWS_HTTP2_PING_DATA_SIZE).
 *      If NULL, the 8 byte payload will be all zeroes.
 * @param on_completed Optional callback, invoked when PING ACK is received from peer,
 *      or when a connection error prevents the PING ACK from being received.
 *      Callback always fires on the connection's event-loop thread.
 * @param user_data User-data pass to on_completed callback.
 */
```

graebm: also, let's move `AWS_H2_PING_DATA_SIZE` from h2_frames.h to connection.h and rename it "HTTP2"
it's a useful constant

also, let's rename `connection` here to `http2_connection` to drive home how this is http2 only

PR #250, opened 2020-05-19
H2 manual window update

TingDaoK:
*Issue #, if available:*

*Description of changes:*

1. Maintain the existing API for window updates. Stream update window and connection update window will result in sending stream/connection window update frame and update the self window size
2. Created the stream cross-thread task, create a lock for every stream.
 Pro:
 (1) I think it's more clear to let stream handle the stream-control frames.
 (2) I feel if the user shares the same lock among all the h2 APIs, it still will block the thread. If each stream has its own lock, it will be better.
Con:
 (1) As Mike reminds me, if every stream sends its own control frames (right now we just send the window update frame), the outgoing_frames_task will take a poorly packed message and stall other frames from sending  

Comments about this are very welcome! When it comes to multiple threads, I am not sure about my decision...

TODOs:
1. We need to split the h1/h2 flow-control window. They are totally different, we should not share the same generic API. We may change the generic API or take different protocol sepcfic APIs. Suggestion:
`void aws_http1_connection_update_window(struct aws_http_connection *connection, size_t increment_size)`
`int aws_http2_connection_update_window(struct aws_http_connection *connection, size_t increment_size)`
`int aws_http2_stream_update_window(struct aws_http_stream *stream, size_t increment_size)`
2. We may change the outgoing frame logic to that the streams maintain their own frames queue and send their messages for control frames to unblock every control frame.  


source/h2_stream.c
```diff
@@ -211,22 +225,141 @@ struct aws_h2_stream *aws_h2_stream_new_request(
...
+    } /* END CRITICAL SECTION */
+
+    while (!aws_linked_list_empty(&pending_frames)) {
+        struct aws_linked_list_node *node = aws_linked_list_pop_front(&pending_frames);
+        struct aws_h2_frame *frame = AWS_CONTAINER_OF(node, struct aws_h2_frame, node);
+        aws_h2_connection_enqueue_outgoing_frame(connection, frame);
+    }
+    /* We already enqueued the window_update frame, just apply the change and let our peer check this value. No matter
+     * overflow happens or not, peer will dectect it for us. */
+    if (aws_h2err_failed(aws_h2_stream_window_size_change(stream, (int32_t)window_update_size, true /*self*/))) {
```
graebm, line 275: It's possible that the peer does NOT detect this.
If the peer sends some DATA before this WINDOW_UPDATE is received, then it might **not** cause overflow on the peer's side, and then the 2 sides are just out of sync.

Probably best to just kill the connection if we overflow locally.
Detecting the overflow from within the public update_window() fn would require that we store the current window in synced_data, which seems like an unnecessary amount of synchronization.
We could add a new AWS_ERROR like AWS_ERROR_HTTP_FLOW_CONTROL_WINDOW_OVERFLOW (we'd send AWS_H2_INTERNAL_ERROR to the peer)

TingDaoK: you are right, but let's take another approach ensures the sync between remote and local to address this problem.

source/h2_stream.c
```diff
@@ -270,11 +270,9 @@ static void s_stream_cross_thread_work_task(struct aws_channel_task *task, void
...
         aws_h2_connection_enqueue_outgoing_frame(connection, frame);
     }
-    /* We already enqueued the window_update frame, just apply the change and let our peer check this value. No matter
-     * overflow happens or not, peer will detect it for us. */
-    if (aws_h2err_failed(aws_h2_stream_window_size_change(stream, (int32_t)window_update_size, true /*self*/))) {
-        stream->thread_data.window_size_self = INT32_MAX;
-    }
+    /* We already enqueued the window_update frame, just apply the change and let our peer check this value, no matter
+     * overflow happens or not. Peer will detect it for us. */
+    stream->thread_data.window_size_self += window_update_size;
```
TingDaoK, line 275: The largest legal value will be 3*(2^31-1), which is way less than INT64_MAX, so if the window_size_self overflows, there is no reason peer cannot find it out. So just apply the change and ignore the possible overflow. Or tell me that I am too naive.

TingDaoK: Oh no, it should be 2*(2^31-1). When the self-window-size is 2^31 -1, peer sends the data frame. Before we process the data frame, we send WINDOW_UPDATE frame to increment the window by 2^31 - 1. Now the peer stay in a valid state, but we can get  2*(2^31-1) in the self-window-size.

PR #251, opened 2020-05-20
Http2 options

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- http2_connection_options for initial settings in HTTP/2 connection, reuse the same generic API, but it will be ignored by HTTP/1 connection. This will lead to redundant structures for create new connection. 
- We will add more HTTP/2 specific callbacks to http2_connection_options soon.


include/aws/http/connection.h
```diff
@@ -206,6 +236,8 @@ struct aws_http_client_connection_options {
 
     /**
      * Optional.
+     * The initial connection flow-control window size for HTTP/1 connection.
+     * Ignored by HTTP/2 connection, since the initial connection flow-control window in HTTP/2 is not configurable.
      * A default size is set by AWS_HTTP_CLIENT_CONNECTION_OPTIONS_INIT.
      */
     size_t initial_window_size;
```
TingDaoK, line 243: We probably want to rename this. When we resolve the WINDOW HTTP/1 & 2 problems.

include/aws/http/connection.h
```diff
@@ -241,6 +275,28 @@ struct aws_http_client_connection_options {
...
+     * HTTP/2 specific options.
+     * Optional for HTTP/2 connection. Ignored for HTTP/1 connection.
+     */
+    struct aws_http2_connection_options {
+        /**
+         * Required.
+         * Settings to change as the initial local settings.
+         * If set, these settings will be sent to remote peer once the connection set up.
+         * It can be zero out, no change will made to the default initial settings defined by from RFC-7540 6.5.2.
+         * Note: you may want to disable PUSH here, since we don't support server push for now.
```
TingDaoK, line 289: Or we can initialize it by AWS_HTTP_CLIENT_CONNECTION_OPTIONS_INIT. We probably want an AWS_HTTP2_CLIENT_CONNECTION_OPTIONS_INIT.

include/aws/http/connection.h
```diff
@@ -241,6 +275,28 @@ struct aws_http_client_connection_options {
      * reaches 0, no further data will be received.
      **/
     bool manual_window_management;
+
+    /**
+     * HTTP/2 specific options.
+     * Optional for HTTP/2 connection. Ignored for HTTP/1 connection.
+     */
+    struct aws_http2_connection_options {
```
TingDaoK, line 283: If we support HTTP/2 server in the future, this structure may need to be moved?

graebm: we'll worry about it in the future
server.h could just include connection.h

include/aws/http/connection.h
graebm: 👍yeah, we'll worry about this when we resolve the other problems

include/aws/http/connection.h
```diff
@@ -144,6 +144,38 @@ struct aws_http_proxy_options {
...
+ * Options for changing the settings for an HTTP/2 connection.
+ */
+struct aws_http2_change_settings_options {
+    /**
+     * Optional
+     * The data of settings to change.
+     * Note: each setting has its boundary. If settings_array is not set, num_settings has to be 0 to send an empty
+     * SETTINGS frame.
+     */
+    struct aws_http2_setting *settings_array;
```
graebm, line 157: bluhhhhh I know I told you to make the `aws_http2_change_settings_options` struct so we could re-use it between the function, and the connection_options. Now I think maybe we should put the function back to the way it was, and just add
`initial_settings_array`, `num_initial_settings`, and `on_initial_settings_ack` to the connection options.

I just have a sense I that the bindings will end up handling the callback differently for the change_settings() function vs the initial_settings. And if we're using the same struct, the bindings will end up making some class that represents that struct, and the binding APIs will end up weird ... or something

Also, I don't think we'll fire the callback if the connection fails to setup at all, so the documentation about "will always fire" won't be completely correct.

source/h2_connection.c
```diff
@@ -341,25 +328,16 @@ static struct aws_h2_connection *s_connection_new(
...
     if (!init_pending_settings) {
         goto error;
     }
-    struct aws_h2_frame *init_settings_frame =
-        aws_h2_frame_new_settings(connection->base.alloc, initial_settings, iter, false /*ACK*/);
+    struct aws_h2_frame *init_settings_frame = aws_h2_frame_new_settings(
+        connection->base.alloc,
+        http2_options->initial_settings.settings_array,
+        http2_options->initial_settings.num_settings,
+        false /*ACK*/);
```
graebm, line 340: Ugh I just realized something nasty:

We're cleaning up `thread_data.pending_settings_queue` (and also `thread_data.pending_ping_queue`) from `s_handler_destroy()`. That doesn't run until all the refcounts on the connection go away. That's why things like `thread_data.active_streams_map` and `synced_data.pending_stream_list` are cleared from `s_finish_shutdown()`. It lets their completion callbacks fire when the connection class is still "valid". If we're not firing all the completion callbacks from `s_handler_destroy()`, then the bindings probably won't release all their refcounts.

But here we are in `s_connection_new()` adding stuff to the `outgoing_frames_queue` and `pending_settings_queue`. If the connection fails to ever set up, then `s_handler_destroy()` never runs.

Two options I can think of:
1) Don't put anything in these lists until s_handler_installed() is called.
2) Clear `pending_settings_queue` lists in both `s_finish_shutdown()` and `s_handler_destroy()`, but don't fire any callbacks from `s_handler_destroy()`

TingDaoK: Yeah, you are right, I never actually think through how the connection shuts down. The difference between shutdown and destroy. Thank you for pointing it out! Taking the second option here. Since the first one will need h2_connection to keep the initial settings somewhere.

TingDaoK: Taking the first option instead, since the user data is not valid until the handler_installed...

bin/elasticurl/main.c
```diff
@@ -721,20 +721,30 @@ int main(int argc, char **argv) {
...
 
+    struct aws_http2_setting settings_array[] = {
+        {.id = AWS_HTTP2_SETTINGS_ENABLE_PUSH, .value = 0},
+    };
+
+    struct aws_http2_connection_options http2_options = {
+        .initial_settings_array = settings_array,
+        .num_initial_settings = AWS_ARRAY_SIZE(settings_array),
+        .max_closed_streams = AWS_HTTP2_DEFAULT_MAX_CLOSED_STREAMS,
+    };
```
graebm, line 732: trivial/arbitrary: let's just leave this all out and keep elasticurl simple, don't even bother setting `aws_http_client_connection_options.http2_options`

include/aws/http/connection.h
```diff
@@ -334,9 +393,15 @@ AWS_HTTP_API
...
- * HTTP/2 specific. Change the HTTP/2 conenction settings.
- * Settings frame will be sent. If set, on_completed callback will be always be invoked, if no error is reported then
- * the peer has acknowledged the settings and the change has been applied.
+ * Send a SETTINGS frame (HTTP/2 only).
+ * SETTINGS will be applied locally when SETTINGS ACK is received from peer.
+ *
+ * @param http2_connection HTTP/2 connection.
+ * @param settings_array The array of settings to change. Note: each setting has its boundary.
+ * @param num_settings The num of settings to change in settings_array.
+ *      If it's zero, we will just do nothing, no frame will be sent. And callback will be ignored.
```
graebm, line 402: regarding num_settings=0: It's weird to have the function return AWS_OP_SUCCESS, but then the callback doesn't fire. 

Let's do one of these things:
1) Let's just send the empty SETTINGS frame. It's not against the law. That's what they asked us to do.
2) It's an error to call this with num_settings=0

source/h2_connection.c
```diff
@@ -341,37 +334,17 @@ static struct aws_h2_connection *s_connection_new(
...
-            "Failed to create the initial settings frame, error %s",
-            aws_error_name(aws_last_error()));
-        aws_mem_release(connection->base.alloc, init_pending_settings);
+    /* User data from connection base is not ready until the handler installed */
+    connection->thread_data.init_pending_settings = s_new_pending_settings(
+        connection->base.alloc,
+        http2_options->initial_settings_array,
+        http2_options->num_initial_settings,
+        http2_options->on_initial_settings_completed,
+        NULL /*user_data*/);
```
graebm, line 343: /* user_data is set later... */

source/h2_connection.c
```diff
@@ -1563,9 +1530,28 @@ static void s_handler_installed(struct aws_channel_handler *handler, struct aws_
             goto error;
         }
     }
-    /* Initial settings are already enqueued */
-    aws_h2_try_write_outgoing_frames(connection);
+    struct aws_h2_pending_settings *init_pending_settings = connection->thread_data.init_pending_settings;
+    /* Set user_data here, the user_data is valid now */
+    init_pending_settings->user_data = connection->base.user_data;
```
graebm, line 1535: We could do the following here:
```
aws_linked_list_push_back(&connection->thread_data.pending_settings_queue, init_pending_settings);
connection->thread_data.init_pending_settings = NULL;
```

and then we could remove the special-case code for `connection->thread_data.init_pending_settings` from `s_decoder_on_settings_ack()` and `s_finish_shutdown()`

bin/elasticurl/main.c
TingDaoK: I don't get it, actually. Will it be clearer to show the use case of http2_options in elasticurl?

source/h2_connection.c
```diff
@@ -435,25 +408,9 @@ static void s_handler_destroy(struct aws_channel_handler *handler) {
...
-        struct aws_linked_list_node *node = aws_linked_list_pop_front(&connection->thread_data.pending_ping_queue);
-        struct aws_h2_pending_ping *pending_ping = AWS_CONTAINER_OF(node, struct aws_h2_pending_ping, node);
-        /* fire the user callback with error */
-        if (pending_ping->on_completed) {
-            pending_ping->on_completed(
-                &connection->base, 0 /*fake rtt*/, AWS_ERROR_HTTP_CONNECTION_CLOSED, pending_ping->user_data);
-        }
-        aws_mem_release(connection->base.alloc, pending_ping);
+    if (connection->thread_data.init_pending_settings) {
+        /* if initial settings is not ACKed by peer, we need to clear the memory here */
```
graebm, line 412: trivial: /* if initial settings were never sent, we need to clear the memory here */

PR #258, opened 2020-06-04
User goaway api

TingDaoK:
*Issue #, if available:*

*Description of changes:*
1. aws_http2_connection_send_goaway. Asynchronously send goaway
2. aws_http2_connection_get_sent_goaway, which is designed to get the latest sent goaway.
3. Rule out adding into sync_data after s_stop() called

It's arguable that if aws_http2_connection_send_goaway called, but aws_http2_connection_get_sent_goaway cannot get the request goaway until the task runs and actually send/enqueue the goaway. But if user request to send a goaway, then another goaway sent due to internal reason, it will be confusing.

TODO:
Instead of aws_http2_connection_send_goaway(conn, uint32_t http2_error_code, bool allow_more_streams, debug_data), we probably need functions that provide various ways of shutting down. As Mike and me discussed:
```
aws_http_connection_close_immediately(conn)//not sending anything just free the memory as fast as you can
aws_http_connection_close_normal(conn)//send "real" GOAWAY and shutdown gracefully
aws_http_connection_close_gracefully(conn, time?)//send graceful shutdown warning, begin to shutdown after some period of time? 
```


include/aws/http/connection.h
```diff
@@ -478,6 +478,41 @@ int aws_http2_connection_ping(
...
     void *user_data);
 
+/**
+ * Asynchronously request to send a GOAWAY frame. (HTTP/2 only).
+ *
+ * @param http2_connection HTTP/2 connection.
+ * @param http2_error The HTTP/2 error code (RFC-7540 section 7) sent by peer.
+ *      `enum aws_http2_error_code` lists official codes.
+ * @param allow_more_streams If set, the graceful shutdown warning will be sent, new streams from peer will be allowed
+ *      until the connection shuts down. Note: If set, the sent http2_error will be set to AWS_HTTP2_ERR_NO_ERROR
```
TingDaoK, line 488: Alternative: AWS_INVALID_ARGUMENT for allow_more_streams=true && http2_error!=AWS_HTTP2_ERR_NO_ERROR

source/h2_connection.c
```diff
@@ -2154,17 +2263,82 @@ static int s_connection_ping(
...
+}
+
+/* Send a GOAWAY with the lowest possible last-stream-id or graceful shutdown warning */
+static void s_send_goaway(
+    struct aws_h2_connection *connection,
+    uint32_t h2_error_code,
+    bool allow_more_streams,
+    const struct aws_byte_cursor *optional_debug_data) {
     AWS_PRECONDITION(aws_channel_thread_is_callers_thread(connection->base.channel_slot->channel));
+    AWS_PRECONDITION(!allow_more_streams || (h2_error_code == AWS_HTTP2_ERR_NO_ERROR));
```
graebm, line 2322: I'm not sure this should be an assert.
It's WEIRD, but it's not against the law.
We could log about it at DEBUG or WARN level?

TingDaoK: mmm, that's the internal function. If this is called with this kind of error, probably something is wrong internally... In this case, would it be better to assert?

TingDaoK: And if user calls the API with this kind of error, I just ignore this kind of error when we create the pending_goaway structure. It's probably better to log it.
'''
pending_goaway->http2_error = allow_more_streams ? AWS_HTTP2_ERR_NO_ERROR : http2_error;
'''

source/h1_connection.c
```diff
@@ -439,11 +441,18 @@ int aws_h1_stream_activate(struct aws_http_stream *stream) {
...
 
+    bool connection_open;
     bool should_schedule_task = false;
 
     { /* BEGIN CRITICAL SECTION */
         s_h1_connection_lock_synced_data(connection);
 
+        connection_open = connection->synced_data.is_open;
+        if (!connection_open) {
+            s_h1_connection_unlock_synced_data(connection);
```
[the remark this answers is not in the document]
TingDaoK: Yep, you are right. I was thinking that the cross thread task will check for the `new_stream_error_code` anyway, but h1 will not. And, it's a little bit weird that we checked the `new_stream_error_code` when we make the request and check it again when we activate the stream. But as I found the PR for activate stream and saw the discussion between you and Jonathon, I guess it's fine...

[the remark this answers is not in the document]
graebm: I saw your changes, good work.

man, in retrospect I regret adding the activate() function
wish we'd changed it instead so the stream pointer was delivered via some on_setup callback, instead of being returned in a "partially initialized" state from make_request() :|
it would have been a bigger change at the time, but we wouldn't have all these weird edge cases

source/h2_connection.c
graebm: this is a pretty arbitrary call... but if it's not against the law, we should do what the user tells us to do. Let's just let them send a non-zero error code, even if allow_more_streams is true.

TingDaoK: Yeah, makes sense. Will do.

PR #259, opened 2020-06-05
User rst stream

TingDaoK:
*Issue #, if available:*

*Description of changes:*
1. User can request reset the stream. If stream is closed already, silently ignore the request.
2. Getters for rst_stream received/sent. ONLY FOR ERROR for now, but yeah, it's arguable.
3. make sure stream is still alive until cross thread task runs.

Arguable behaviour design:
1. User request to cancel the stream, with no error, stream should complete with AWS_ERROR_SUCCESS or PROTOCOL_ERROR?
- Taking ERROR_SUCCESS here, since it's not an error. And when we received rst_stream with NO_ERROR, we will complete stream with SUCCESS as well.
2. Rest stream also support uint32 error code? Or ensure it's limited in the range of enum?
- It should probably support uint32 error code... But the internal implementation will need to support it as well... I don't know... Take the easiest for now (limited in the range of enum).
3. ONLY GET ERROR WHEN THERE IS AN ERROR or RST_STREAM received/sent, getters for this. And get error_code includes NO_ERROR?
- I feel both are fine, personally prefer the first one, and implemented that way. But the other way also makes sense.


source/h2_stream.c
```diff
@@ -261,22 +288,36 @@ static void s_stream_cross_thread_work_task(struct aws_channel_task *task, void
         /* window_update_size is ensured to be not greater than AWS_H2_WINDOW_UPDATE_MAX */
         window_update_size = stream->synced_data.window_update_size;
         stream->synced_data.window_update_size = 0;
+        requested_reset_error_code = stream->synced_data.requested_reset_error_code;
 
         s_unlock_synced_data(stream);
     } /* END CRITICAL SECTION */
 
     while (!aws_linked_list_empty(&pending_frames)) {
         struct aws_linked_list_node *node = aws_linked_list_pop_front(&pending_frames);
```
TingDaoK, line 297: all the frames here are window update for now. But, who knows what will we do in the future. (Probably let stream take care of all the frames about stream?

graebm:
haven't read your actual code yet, but responding to questions...

>     1. User request to cancel the stream, with no error, stream should complete with AWS_ERROR_SUCCESS or PROTOCOL_ERROR?
> 
>     * Taking ERROR_SUCCESS here, since it's not an error. And when we received rst_stream with NO_ERROR, we will complete stream with SUCCESS as well.

I would NOT consider that success. The request/response did not complete. We could introduce a new error-code like AWS_ERROR_HTTP_RST_STREAM_SENT. I would use that error-code regardless of the HTTP/2 code chosen by the user.

BUT BUT BUT there is one edge case where the stream would still complete successfully:
"a server MAY request that the client abort transmission of a request without error by sending a RST_STREAM with an error code of NO_ERROR after sending a complete response (i.e., a frame with the END_STREAM flag)"
We could leave a TODO for that edge case, since it's a server-only feature.

>     1. Rest stream also support uint32 error code? Or ensure it's limited in the range of enum?
> 
> 
>     * It should probably support uint32 error code... But the internal implementation will need to support it as well... I don't know... Take the easiest for now (limited in the range of enum).

Let's change it all to be uint32_t, since it's legal to make up your own error code

> 
> 
>     1. ONLY GET ERROR WHEN THERE IS AN ERROR or RST_STREAM received/sent, getters for this. And get error_code includes NO_ERROR?
> 
> 
>     * I feel both are fine, personally prefer the first one, and implemented that way. But the other way also makes sense.

Not sure what you mean here.

source/h2_frames.c
```diff
@@ -144,11 +144,11 @@ const char *aws_http2_error_code_to_str(enum aws_http2_error_code h2_error_code)
...
 struct aws_h2err aws_h2err_from_h2_code(enum aws_http2_error_code h2_error_code) {
-    AWS_PRECONDITION(h2_error_code > AWS_HTTP2_ERR_NO_ERROR && h2_error_code < AWS_HTTP2_ERR_COUNT);
+    AWS_PRECONDITION(h2_error_code >= AWS_HTTP2_ERR_NO_ERROR && h2_error_code < AWS_HTTP2_ERR_COUNT);
 
     return (struct aws_h2err){
         .h2_code = h2_error_code,
-        .aws_code = AWS_ERROR_HTTP_PROTOCOL_ERROR,
+        .aws_code = h2_error_code == AWS_HTTP2_ERR_NO_ERROR ? AWS_ERROR_SUCCESS : AWS_ERROR_HTTP_PROTOCOL_ERROR,
     };
 }
```
graebm, line 153: maybe don't modify this function
When the user sends RST_STREAM just build the aws_h2err manually:
```
struct aws_h2err h2err = {
    .h2_code = http2_error_from_user,
    .aws_code = AWS_ERROR_HTTP_RST_STREAM_SENT,
};
```

source/h2_stream.c
```diff
@@ -261,22 +288,36 @@ static void s_stream_cross_thread_work_task(struct aws_channel_task *task, void
...
+        }
         aws_h2_connection_enqueue_outgoing_frame(connection, frame);
     }
     /* The largest legal value will be 2 * max window size, which is way less than INT64_MAX, so if the window_size_self
      * overflows, remote peer will find it out. So just apply the change and ignore the possible overflow.*/
     stream->thread_data.window_size_self += window_update_size;
 
+    if (requested_reset_error_code != AWS_HTTP2_ERR_COUNT) {
+        /* TODO:: If error happens inside this function, should we shutdown the connection? */
+        s_send_rst_and_close_stream(stream, aws_h2err_from_h2_code(requested_reset_error_code));
```
graebm, line 312: yup.
probably want to call the connection's s_shutdown_due_to_write_err()

source/h2_stream.c
```diff
@@ -261,22 +288,36 @@ static void s_stream_cross_thread_work_task(struct aws_channel_task *task, void
...
+    }
+
     /* It's likely that frames were queued while processing cross-thread work.
      * If so, try writing them now */
     aws_h2_try_write_outgoing_frames(connection);
+
+end:
+    aws_http_stream_release(&stream->base);
```
graebm, line 320: niiiice catch

source/h2_stream.c
```diff
@@ -358,6 +415,73 @@ static void s_stream_update_window(struct aws_http_stream *stream_base, size_t i
...
+        AWS_H2_STREAM_LOG(TRACE, stream, "Scheduling stream cross-thread work task");
+        /* increment the refcount of stream to keep it alive until the task runs */
+        aws_atomic_fetch_add(&stream->base.refcount, 1);
+        aws_channel_schedule_task_now(connection->base.channel_slot->channel, &stream->cross_thread_work_task);
+        return AWS_OP_SUCCESS;
+    }
+
+    if (!stream_is_activated) {
+        /* stream has not ben activated yet, it's invalid to update the stream window. */
+        AWS_H2_STREAM_LOG(
```
graebm, line 448: copy/paste comment mentions "stream window"

source/h2_stream.c
```diff
@@ -358,6 +415,73 @@ static void s_stream_update_window(struct aws_http_stream *stream_base, size_t i
...
+    bool stream_is_activated;
+    bool cross_thread_work_should_schedule = false;
+
+    { /* BEGIN CRITICAL SECTION */
+        s_lock_synced_data(stream);
+
+        err |= stream->synced_data.requested_reset_error_code != AWS_HTTP2_ERR_COUNT;
+        stream_is_activated = stream->synced_data.is_activated;
+        if (!err && stream_is_activated) {
+            cross_thread_work_should_schedule = !stream->synced_data.is_cross_thread_work_task_scheduled;
```
graebm, line 432: We probably ALSO shouldn't send RST_STREAM if the stream is complete?

The spec says you shouldn't send multiple RST_STREAM
But really it probably just opens us up to a bunch of weird bugs to let users queue work after we know the stream is closed

source/h2_stream.c
```diff
@@ -358,6 +415,73 @@ static void s_stream_update_window(struct aws_http_stream *stream_base, size_t i
...
+    struct aws_h2_stream *stream = AWS_CONTAINER_OF(stream_base, struct aws_h2_stream, base);
+    struct aws_h2_connection *connection = s_get_h2_connection(stream);
+    int err = 0;
+    bool stream_is_activated;
+    bool cross_thread_work_should_schedule = false;
+
+    { /* BEGIN CRITICAL SECTION */
+        s_lock_synced_data(stream);
+
+        err |= stream->synced_data.requested_reset_error_code != AWS_HTTP2_ERR_COUNT;
```
graebm, line 429: error codes can be any uint32
either use a separate bool to track if this is set, or use a larger data type with a different magic value

source/request_response.c
```diff
@@ -568,12 +568,19 @@ void aws_http_message_set_body_stream(struct aws_http_message *message, struct a
...
-    AWS_PRECONDITION(stream->vtable->http1_write_chunk);
+int aws_http1_stream_write_chunk(struct aws_http_stream *http1_stream, struct aws_http1_chunk_options *options) {
+    AWS_PRECONDITION(http1_stream);
+    AWS_PRECONDITION(http1_stream->vtable);
     AWS_PRECONDITION(options);
-    return stream->vtable->http1_write_chunk(stream, options);
+    if (!http1_stream->vtable->http1_write_chunk) {
+        AWS_LOGF_TRACE(
+            AWS_LS_HTTP_STREAM,
+            "id=%p: HTTP/1 stream only function invoked on other stream, igoring call.",
```
graebm, line 578: ~~igoring~~ ignoring
this typo happens a few times...

include/aws/http/private/h2_stream.h
```diff
@@ -77,13 +77,32 @@ struct aws_h2_stream {
...
         /* New `aws_h2_frames *` stream control frames created by user that haven't moved to connection `thread_data`
          * yet */
         struct aws_linked_list pending_frame_list;
+
+        /* The aws_http2_error_code user requested to remote peer via rst_stream. Set to AWS_HTTP2_ERR_COUNT if no
+         * rst_stream has been requested so far. */
+        enum aws_http2_error_code requested_reset_error_code;
+
+        /* True, if stream has been activated by user. */
+        bool is_activated;
```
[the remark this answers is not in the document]
TingDaoK: Oh, yeah, I was thinking about this after reading you first comment!

source/h2_stream.c
```diff
@@ -358,6 +421,67 @@ static void s_stream_update_window(struct aws_http_stream *stream_base, size_t i
...
+    if (!stream_is_activating) {
+        AWS_H2_STREAM_LOG(
+            ERROR, stream, "Reset stream failed. State for the stream doesn't allow RST_STREAM to send. (closed/idle)");
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
+
+    if (reset_called) {
+        AWS_H2_STREAM_LOG(ERROR, stream, "Reset stream failed. Reset stream has been called before.");
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
```
graebm, line 462: not sure if this should be an exception ....
like, if the user has multiple codepaths that try to reset the stream if something goes wrong, and multiple things go wrong, might be nicer to just silently ignore the subsequent calls...

source/h2_stream.c
```diff
@@ -504,7 +634,10 @@ int aws_h2_stream_encode_data_frame(
 
         /* Failed to write DATA, treat it as a Stream Error */
         AWS_H2_STREAM_LOGF(ERROR, stream, "Error encoding stream DATA, %s", aws_error_name(aws_last_error()));
-        s_send_rst_and_close_stream(stream, aws_h2err_from_last_error());
+        struct aws_h2err returned_h2err = s_send_rst_and_close_stream(stream, aws_h2err_from_last_error());
+        if (aws_h2err_failed(returned_h2err)) {
+            aws_h2_connection_shutdown_due_to_write_err(connection, returned_h2err.aws_code);
+        }
```
graebm, line 640: whoa, nice catch!

source/h2_stream.c
```diff
@@ -261,22 +287,44 @@ static void s_stream_cross_thread_work_task(struct aws_channel_task *task, void
...
     /* The largest legal value will be 2 * max window size, which is way less than INT64_MAX, so if the window_size_self
      * overflows, remote peer will find it out. So just apply the change and ignore the possible overflow.*/
     stream->thread_data.window_size_self += window_update_size;
 
+    if (reset_called) {
+        /* TODO:: It's possible that server sends a rst_stream with NO_ERROR, after full response has been sent. In this
+         * case the aws_code here should be AWS_ERROR_HTTP_SUCCESS */
+        struct aws_h2err h2err;
+        h2err.h2_code = user_reset_error_code;
+        h2err.aws_code = AWS_ERROR_HTTP_RST_STREAM_SENT;
```
graebm, line 315: we could totally handle this TODO righ now
```
if (is_server && stream->thread_data.state == HALF_CLOSED_LOCAL)
    /* (RFC-7540 8.1) A server MAY request that the client abort transmission of a request without error by sending a RST_STREAM with an error code of NO_ERROR after sending a complete response */
    h2err.aws_code = AWS_ERROR_SUCCESS;
else:
    h2err.aws_code = AWS_ERROR_HTTP_RST_STREAM_SENT
```

source/h2_connection.c
```diff
@@ -1935,6 +1936,18 @@ int aws_h2_stream_activate(struct aws_http_stream *stream) {
         aws_channel_schedule_task_now(connection->base.channel_slot->channel, &connection->cross_thread_work_task);
     }
 
+    /* set stream's activated flag, make sure stream's cross thread work will be scheduled after connection's */
+    { /* BEGIN CRITICAL SECTION */
+        int err = aws_mutex_lock(&h2_stream->synced_data.lock);
```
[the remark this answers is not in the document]
justinboswell: +1 to this idea. Also need an unlock function.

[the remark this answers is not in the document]
TingDaoK: Yep, will do

include/aws/http/private/h2_stream.h
```diff
@@ -56,6 +56,13 @@ enum aws_h2_stream_state {
     AWS_H2_STREAM_STATE_COUNT,
 };
 
+/* simplified stream state for API implementation */
+enum aws_h2_stream_api_state {
+    AWS_H2_STREAM_API_STATE_INIT,
+    AWS_H2_STREAM_API_STATE_ACTIVATED,
```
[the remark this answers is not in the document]
TingDaoK: yeah, you did. ┑(￣Д ￣)┍. Anyway, yeah, it's better to be active

source/h2_stream.c
```diff
@@ -241,17 +254,30 @@ struct aws_h2_stream *aws_h2_stream_new_request(
...
 
+    if (aws_h2_stream_get_state(stream) == AWS_H2_STREAM_STATE_CLOSED) {
+        /* stream is closed, silently ignoring the requests from user */
+        AWS_H2_STREAM_LOG(
+            DEBUG, stream, "Stream closed before cross thread work task runs, ignoring everything was sent by user.");
+        goto end;
+    }
+
+    /* Not sending window update at half closed remote state */
+    bool ignore_window_udpate = aws_h2_stream_get_state(stream) == AWS_H2_STREAM_STATE_HALF_CLOSED_REMOTE;
```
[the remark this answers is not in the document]
TingDaoK: I am going to send the code spell checker a ticket about this, if I can...

source/h2_stream.c
TingDaoK: I don't know... I just feel that this kind of check will be more accurate if happens when the user calls the reset_stream function, instead of when the task runs.

source/h2_stream.c
```diff
@@ -261,22 +287,44 @@ static void s_stream_cross_thread_work_task(struct aws_channel_task *task, void
...
 
         s_unlock_synced_data(stream);
     } /* END CRITICAL SECTION */
 
     while (!aws_linked_list_empty(&pending_frames)) {
         struct aws_linked_list_node *node = aws_linked_list_pop_front(&pending_frames);
         struct aws_h2_frame *frame = AWS_CONTAINER_OF(node, struct aws_h2_frame, node);
+        if (ignore_window_udpate && (frame->type == AWS_H2_FRAME_T_WINDOW_UPDATE)) {
+            /* Do not send the window udpate frame and clean it up */
+            aws_h2_frame_destroy(frame);
```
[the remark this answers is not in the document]
TingDaoK: Yeah, it's better. And I guess we can just remove the pending frame list for stream now. Nothing will be there anymore. We can always add it back if/when we need it

source/h2_stream.c
```diff
@@ -328,8 +377,9 @@ static void s_stream_update_window(struct aws_http_stream *stream_base, size_t i
 
         err |= aws_add_size_checked(stream->synced_data.window_update_size, increment_size, &sum_size);
         err |= sum_size > AWS_H2_WINDOW_UPDATE_MAX;
+        stream_is_activating = stream->synced_data.api_state == AWS_H2_STREAM_API_STATE_ACTIVATED;
 
-        if (!err) {
+        if (!err && stream_is_activating) {
```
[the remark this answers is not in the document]
TingDaoK: I guess you mean we create the WINDOW_UPDATE frame in cross_thread_work task... Agree, but still most of error handling should happen here.

source/h2_stream.c
TingDaoK: Changed my mind, the edge case is just too crazy to happen in real world. Will do.

Issue #193, opened 2020-03-05
Add option to force tunnelling for proxied plaintext connections


adamthom-amzn:
Similar to curl's `--proxytunnel` option, it would be convenient if the HTTP proxy options included a flag to force tunneling for all proxied calls (via CONNECT).  Current behavior is to use CONNECT only for TLS requests.  Some additional context from one of our users:

> we pushed for CONNECT request to be the primary method of establishing proxy connections in the Netty4 client because it provides cleaner protocol semantics, is applicable to more protocols (you can establish a TCP tunnel for any protocol using a CONNECT request), and in general makes the proxy connecting code easier to support.  We've thought about deprecating absolute resource handshakes for a while.  We implemented it to support [the current] client's proxy behavior.  However, this handshake type forces [our product] to function as an L7 proxy instead of an L4 proxy as the contents of the HTTP request must be buffered and sent to the destination.  If you have to choose between one mechanism or another for establishing proxy connections I would side toward CONNECT requests for everything as it supports both HTTP and HTTPS.

bretambrose:
Support for this was added recently (https://github.com/awslabs/aws-c-http/blob/main/include/aws/http/proxy.h#L37).  It's currently bound out in the C++ CRT and the other CRTs will follow shortly over the next few weeks.

PR #261, opened 2020-06-07
Proxy tunnel

bretambrose:
https://github.com/awslabs/aws-c-http/issues/193

* Adds support for plaintext tunneling proxy connections

Per-request from Coral team.

Not backwards compatible; requires per-CRT fixups.  In addition to exposing the proxy connection type, CRTs will need to conditionally switch the proxy connection type from HTTP_FORWARD to HTTPS_TUNNEL if tls connection options are set on the overall connection configuration.

It would be nice to have a public, live endpoint that accepted plaintext CONNECT requests for an integration test.


include/aws/http/connection.h
```diff
@@ -131,6 +131,29 @@ struct aws_http_connection_monitoring_options {
...
 
+/**
+ * Supported proxy connection types
+ */
+enum aws_http_proxy_connection_type {
+    /**
+     * Use the proxy to forward http requests.  Attempting to use both this mode and TLS on the tunnel destination
+     * is a configuration error.
+     */
+    AWS_HPCT_HTTP_FORWARD = 0,
```
graebm, line 142: trivial: AWS_HTTP_PCT_FORWARD? so it only has "http" in the name once?

source/proxy_connection.c
```diff
@@ -449,28 +452,41 @@ static void s_aws_http_on_stream_complete_tls_proxy(struct aws_http_stream *stre
...
-    if (channel == NULL || s_vtable->setup_client_tls(aws_channel_get_first_slot(channel), context->tls_options)) {
-        AWS_LOGF_ERROR(
-            AWS_LS_HTTP_CONNECTION,
-            "(%p) Proxy connection failed to start TLS negotiation with error %d(%s)",
-            (void *)context->connection,
-            aws_last_error(),
-            aws_error_str(aws_last_error()));
-        s_aws_http_proxy_user_data_shutdown(context);
-        return;
+        if (channel == NULL || s_vtable->setup_client_tls(left_of_tls_slot, context->tls_options)) {
```
graebm, line 474: `channel` will never be NULL

graebm (approved):

Issue #270, opened 2020-06-30
Double task scheduling when aws_http1_stream_write_chunk() is used.


chehefen:
aws_http1_stream_write_chunk() triggers outgoing_stream_task to be scheduled here: https://github.com/awslabs/aws-c-http/blob/master/source/h1_connection.c#L441. However, outgoing_stream_task could also be scheduled here: https://github.com/awslabs/aws-c-http/blob/master/source/h1_connection.c#L870. These two together can cause the double scheduling of the outgoing_stream_task and cause some linklist assertion to fail when aws_http1_stream_write_chunk is called. In my case, second aws_http1_stream_write_chunk() call fails after first aws_http1_stream_write_chunk call has completed successfully.

graebm:
Fixed in master!
thanks for the excellent diagnosis

PR #274, opened 2020-07-10
Revamp task scheduling logic in h1 connection and stream

graebm:
ISSUE
There were bugs in h1's task scheduling logic https://github.com/awslabs/aws-c-http/issues/270 when chunks were involved. The logic was complicated before chunks were involved, and throwing chunks in the mix pushed it over the edge. It certainly wasn't going to accommodate the more complicated stuff we have coming down the pipe.

SOLUTION:
Make it work like h2. It's still complicated, but it can accommodate lots more functionality, and it's nice whenever the h1 and h2 implementations can be similar.


source/h1_connection.c
```diff
@@ -221,70 +222,18 @@ static bool s_connection_new_requests_allowed(const struct aws_http_connection *
 static int s_stream_send_response(struct aws_http_stream *stream, struct aws_http_message *response) {
     AWS_PRECONDITION(stream);
     AWS_PRECONDITION(response);
-
-    int err;
-    int send_err = AWS_ERROR_SUCCESS;
     struct aws_h1_stream *h1_stream = AWS_CONTAINER_OF(stream, struct aws_h1_stream, base);
-    struct aws_h1_connection *connection = AWS_CONTAINER_OF(stream->owning_connection, struct aws_h1_connection, base);
```
graebm, line 228: I moved the `stream_send_response()` logic from h1_connection.c to h1_stream.c

source/h1_connection.c
```diff
@@ -381,49 +306,45 @@ int aws_h1_stream_activate(struct aws_http_stream *stream) {
...
+                aws_error_name(connection->synced_data.new_stream_error_code));
+            return aws_raise_error(connection->synced_data.new_stream_error_code);
         }
 
         stream->id = aws_http_connection_get_next_stream_id(base_connection);
+        if (!stream->id) {
+            aws_h1_connection_unlock_synced_data(connection);
+            /* aws_http_connection_get_next_stream_id() raises its own error. */
+            return AWS_OP_ERR;
+        }
```
graebm, line 326: the error handling code-flow in `activate_stream()` bothered me, it was all over the place

now it always unlocks and returns early, which I usually don't approve of, but it was the cleanest thing I could come up with

source/h1_stream.c
```diff
@@ -4,103 +4,174 @@
...
-    aws_http_connection_update_window(stream->owning_connection, increment_size);
+static struct aws_h1_connection *s_get_h1_connection(const struct aws_h1_stream *stream) {
+    return AWS_CONTAINER_OF(stream->base.owning_connection, struct aws_h1_connection, base);
+}
+
+static void s_stream_lock_synced_data(struct aws_h1_stream *stream) {
+    aws_h1_connection_lock_synced_data(s_get_h1_connection(stream));
 }
 
-static int s_body_chunks_init(struct aws_h1_stream *stream) {
```
graebm, line 26: These diffs are a mess.
A bunch of chunk initialization code was removed from h1_stream.c
the new code is unrelated

source/h1_stream.c
```diff
@@ -217,9 +286,79 @@ struct aws_h1_stream *aws_h1_stream_new_request_handler(const struct aws_http_re
...
+
     stream->base.server_data = &stream->base.client_or_server_data.server;
     stream->base.server_data->on_request_done = options->on_request_done;
     aws_atomic_fetch_add(&stream->base.refcount, 1);
 
     return stream;
 }
+
+int aws_h1_stream_send_response(struct aws_h1_stream *stream, struct aws_http_message *response) {
+    struct aws_h1_connection *connection = s_get_h1_connection(stream);
```
graebm, line 303: This implementation was moved from h1_connection.c to h1_stream.c,
seems like it belongs here more
and h1_connection.c is nearly 2000 lines long, it certainly doesn't need any more code than strictly necessary

source/h1_encoder.c
```diff
@@ -610,15 +628,16 @@ static int s_state_fn_unchunked_body(struct aws_h1_encoder *encoder, struct aws_
  * Encoder is essentially "paused" here if no chunks are available. */
 static int s_state_fn_chunk_next(struct aws_h1_encoder *encoder, struct aws_byte_buf *dst) {
     (void)dst;
+    AWS_ASSERT(encoder->message->chunk_list); /* we let this be NULL in tests, but not in real world */
 
-    if (aws_linked_list_empty(&encoder->message->chunk_list)) {
+    if (aws_linked_list_empty(encoder->message->chunk_list)) {
```
justinboswell, line 633: Could this just check null on ```chunk_list``` and no assert needed? Or is that a THIS SHOULD NEVER EVER BE situation? Why would we ever allow it to be null in tests if that's the case?

source/h1_stream.c
```diff
@@ -4,103 +4,174 @@
...
-        if (NULL != stream->body_chunks.current_chunk) {
-            s_clean_up_body_chunk(&stream->body_chunks.current_chunk);
-        }
-    } while (aws_h1_populate_current_stream_chunk(&stream->body_chunks));
-    aws_mutex_clean_up(&stream->body_chunks.lock);
+
+    AWS_LOGF_TRACE(AWS_LS_HTTP_STREAM, "id=%p: Running stream cross-thread work task.", (void *)&stream->base);
+
+    /* BEGIN CRITICAL SECTION */
+    s_stream_lock_synced_data(stream);
```
justinboswell, line 62: Maybe put an anonymous scope here like all the other critical sections?

source/h1_stream.c
```diff
@@ -4,103 +4,174 @@
...
     }
-    if (AWS_OP_SUCCESS != aws_chunk_line_from_options(options, &chunk->chunk_line)) {
-        s_clean_up_body_chunk(&chunk);
-        return AWS_OP_ERR;
+
+    int error_code = 0;
+    bool should_schedule_task = false;
+
+    /* BEGIN CRITICAL SECTION */
+    s_stream_lock_synced_data(stream);
```
justinboswell, line 124: anonymous scope?

source/h1_stream.c
```diff
@@ -217,9 +286,79 @@ struct aws_h1_stream *aws_h1_stream_new_request_handler(const struct aws_http_re
...
+    bool body_headers_ignored = stream->base.request_method == AWS_HTTP_METHOD_HEAD;
+    if (aws_h1_encoder_message_init_from_response(
+            &encoder_message, stream->base.alloc, response, body_headers_ignored, &stream->thread_data.chunk_list)) {
+        error_code = aws_last_error();
+        goto error;
+    }
+
+    bool should_schedule_task = false;
+    { /* BEGIN CRITICAL SECTION */
+        s_stream_lock_synced_data(stream);
```
justinboswell, line 318: so you're locking stream data, and unlocking connection data in this block, is that correct/expected? Can you document that?

tests/test_h1_server.c
```diff
@@ -1558,7 +1558,10 @@ static int s_test_error_from_callback(struct aws_allocator *allocator, enum requ
     struct aws_http_message *response;
     ASSERT_SUCCESS(
         s_create_response(&response, 200, headers, AWS_ARRAY_SIZE(headers), &error_from_outgoing_body_stream));
-    ASSERT_SUCCESS(aws_http_stream_send_response(request->request_handler, response));
+
+    /* send_respones() may succeed or fail, depending on when things shut down */
```
justinboswell, line 1562: response

tests/test_h1_client.c
```diff
@@ -29,7 +29,8 @@ static struct aws_http_message *s_new_default_get_request(struct aws_allocator *
     return request;
 }
 
-static void s_destroy_stream_on_complete(void *user_data) {
```
TingDaoK, line 32: Do we need regression test for the issue fixed?

source/h1_connection.c
```diff
@@ -221,70 +222,18 @@ static bool s_connection_new_requests_allowed(const struct aws_http_connection *
...
-        AWS_LS_HTTP_STREAM,
-        "id=%p: Sending response on the stream failed, error %d (%s)",
-        (void *)stream,
-        aws_last_error(),
-        aws_error_name(aws_last_error()));
-    return AWS_OP_ERR;
+/* Schedule the cross-thread work task.
+ * `connection->synced_data.is_cross_thread_work_task_scheduled` must be set to true
+ * (while holding the lock) before calling this function. Do NOT call this if
+ * `is_cross_thread_work_task_scheduled` was ALREADY true */
```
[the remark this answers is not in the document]
graebm: Yes, it's very brittle, but I don't think the alternatives are better.

Alternative 1: This function grabs the lock and checks whether the task should be scheduled. Downside: Many public API calls involve 2 lock() calls instead of 1.

Alternative 2: Call this function while holding the lock. It flips the bool and schedules the task if necessary. Downside: If the task runs synchronously (ex: scheduling channel task during channel shutdown) then we're still holding the lock, and that might lead to callbacks firing that also try to grab the lock.

I could make the function name ridiculously explicit like `s_schedule_cross_thread_work_task_after_flipping_bool_true()`
seriously

include/aws/http/private/h1_stream.h
```diff
@@ -32,32 +47,50 @@ struct aws_h1_stream {
...
+         * Has mirror variable in thread_data */
+        bool has_outgoing_response : 1;
+
+        /* Whether the outgoing message is using chunked encoding */
+        bool using_chunked_encoding : 1;
     } synced_data;
 };
 
-AWS_EXTERN_C_BEGIN
+/* DO NOT export functions below. They're only used by other .c files in this library */
```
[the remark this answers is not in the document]
graebm: That's a great idea, but I would save it for another PR
I'll throw this in a TODO

source/h1_encoder.c
graebm: I'll fix the tests to not be lazy jerks

source/h1_stream.c
graebm: yeah, the anonymous scope is really just to make the block stand out more visually
we haven't been 100% consistent
will change where I can

source/h1_stream.c
graebm: Oops, it was an accident to call stream_lock() and connection_unlock(). Changed to call stream_unlock()
But they do the same thing because the connection and all its streams share a single lock.
I added comments anywhere we touch both the stream and connection state while holding a lock.

(We do give locks to the individual streams in H2, which expects multiple simultaneous streams per connection, but it didn't seem worth the complexity here)

source/h1_connection.c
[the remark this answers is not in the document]
graebm: Removed function.
Just manually scheduling the task from the same functions that dealt with the locking minutiae

source/h1_connection.c
```diff
@@ -485,29 +406,48 @@ struct aws_http_stream *s_make_request(
...
     size_t window_update_size = connection->synced_data.window_update_size;
-    AWS_LOGF_TRACE(
-        AWS_LS_HTTP_CONNECTION,
-        "id=%p: Zeroing window update size, was %zu",
-        (void *)&connection->base,
-        window_update_size);
     connection->synced_data.window_update_size = 0;
 
+    bool has_new_client_streams = !aws_linked_list_empty(&connection->synced_data.new_client_stream_list);
+    aws_linked_list_move_all_back(
```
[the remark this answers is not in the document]
graebm: HALLELUJAH THIS IS GLORIOUS

tests/test_h1_client.c
graebm: 🙇‍♂️

PR #277, opened 2020-07-17
HTTP/1 respects the stream window.

graebm:
### ISSUE
Previously, the "stream window" in HTTP/1 was a lie. There was really just the connection-window. We encountered bugs when a connection was re-used for another stream. The stream **believed** it started with initial-window-size, but really it got whatever connection-window was left over from the previous stream. The connection could stall because the connection-window size was nearly zero, but the stream thought its window was wide open. Likewise, if the previous stream left a larger connection-window, the new stream could receive data that exceeded its initial-window-size.

### SOLUTION
Honor the stream window. All new streams start with initial-window-size. Streams cannot receive more body data than their stream-window allows. Users can no longer control the connection window directly, because it is doing gymnastics to honor these contracts.

### DETAILS
aws_io_messages are put in a queue before processing. This allows the connection to receive more data than a stream's window might allow, and process the data later when the stream's window opens, or the next stream begins. The connection-window is capped at `read_buffer_capacity`, and shrinks when there is unprocessed data in the queue.

users can customize`read_buffer_capacity`, but we'll choose something for them if they leave it 0 in the options.


include/aws/http/private/h1_connection.h
```diff
@@ -64,17 +64,27 @@ struct aws_h1_connection {
...
+         * and process the data later when HTTP-stream's window opens or the next stream begins.
+         *
+         * The `aws_io_message.copy_mark` is used to track progress on partially processed messages.
+         * `unprocessed_bytes` is the sum of all unprocessed bytes across all queued messages.
+         * `capacity` is the limit for how many unprocessed bytes we'd like in the queue.
+         */
         struct {
             struct aws_linked_list messages;
-            /* TODO: more variables will go in this struct in the near-future */
+            size_t unprocessed_bytes;
```
justinboswell, line 78: ```pending_bytes``` to be consistent with everything else?

source/h1_connection.c
```diff
@@ -13,6 +13,8 @@
 #include <aws/http/status_code.h>
 #include <aws/io/logging.h>
 
+#include <inttypes.h>
```
justinboswell, line 16: What made this necessary, it should be covered by common.h?

source/h1_connection.c
```diff
@@ -228,64 +227,77 @@ static int s_stream_send_response(struct aws_http_stream *stream, struct aws_htt
...
-            AWS_LS_HTTP_CONNECTION,
-            "id=%p: Connection cross-thread work task was already scheduled",
-            (void *)connection_base);
+    const size_t desired_size = connection->thread_data.has_switched_protocols
+                                    ? s_calculate_midchannel_desired_connection_window(connection)
+                                    : s_calculate_stream_mode_desired_connection_window(connection);
+
+    const size_t increment_size = aws_sub_size_saturating(desired_size, connection->thread_data.window_size);
+    if (increment_size > 0) {
+        connection->thread_data.window_size += increment_size;
```
justinboswell, line 285: Should we not record this until AFTER we successfully update the window on the channel?

source/h1_connection.c
```diff
@@ -1512,36 +1533,42 @@ static int s_handler_process_read_message(
...
-        message->message_data.len);
-
-    aws_linked_list_push_back(&connection->thread_data.read_buffer.messages, &message->queueing_handle);
+        AWS_LS_HTTP_CONNECTION, "id=%p: Incoming message of size %zu.", (void *)&connection->base, message_size);
 
-    if (s_try_process_read_messages(connection)) {
-        goto shutdown;
-    }
+    /* Shrink connection window by amount of data received */
+    AWS_ASSERT(message_size <= connection->thread_data.window_size);
```
justinboswell, line 1542: Is this actually assert-able? We can guarantee that this never happens?

source/h1_connection.c
```diff
@@ -1641,37 +1696,21 @@ static int s_try_process_next_stream_read_message(struct aws_h1_connection *conn
     size_t bytes_processed = prev_cursor_len - message_cursor.len;
     queued_msg->copy_mark += bytes_processed;
 
+    AWS_ASSERT(connection->thread_data.read_buffer.unprocessed_bytes >= bytes_processed);
```
justinboswell, line 1699: Would we be better off saturate sub-ing this?

source/h1_connection.c
graebm: I used `PRIu64` in a LOGF, which comes from `inttypes.h`

you're thinking of stuff like `uint64_t` which comes from `stdint.h`, which we do cover in common.h

source/h1_connection.c
graebm: My logic here works by calculating the `desired` window, and issuing a request for the difference between `desired` and `current`.

The `window_size` I'm updating here is NOT the official window according to the channel (that would be `aws_channel_slot.window_size`). That variable isn't updated until the channel runs the window update task. If I based my math off that variable, I could end up asking for too much because there are outstanding updates that haven't landed yet.

I'll add comments explaining this...

graebm: I had the thought that we might want to modify how `aws_channel_slot.window_size` works, so every channel handler doesn't feel the need to duplicate it for its own internal math.

Like, keep the task that informs handlers about downstream updates, but 

Like, keep using a task to inform a handler's neighbor that its window has been updated, but keep `aws_channel_slot.window_size` as up to date as possible

source/h1_connection.c
graebm: Yeah, the channel checks that `aws_channel_slot.window_size` will accommodate a message before allowing it to be sent.

The assert is because this is my personal calculation of window size, not the official window size, and I'm *fairly* certain my math is correct.

source/h1_connection.c
```diff
@@ -1098,7 +1096,17 @@ static int s_decoder_on_body(const struct aws_byte_cursor *data, bool finished,
     AWS_LOGF_TRACE(
         AWS_LS_HTTP_STREAM, "id=%p: Incoming body: %zu bytes received.", (void *)&incoming_stream->base, data->len);
 
-    connection->thread_data.body_bytes_decoded += data->len;
+    if (connection->base.manual_window_management) {
+        /* Let stream window shrink by amount of body data received */
+        AWS_ASSERT(incoming_stream->thread_data.window_size >= data->len); /* TODO: assert or real check?*/
```
TingDaoK, line 1101: I vote for assert here. The length of data should never be larger than the stream window size here, and we ensure that by only pass data less than stream window size into decoder

source/h1_stream.c
```diff
@@ -74,16 +79,61 @@ static void s_stream_cross_thread_work_task(struct aws_channel_task *task, void
...
+        aws_h1_connection_try_write_outgoing_stream(connection);
+    }
+
+    /* Add to window size using saturated sum to prevent overflow.
+     * Saturating is fine because it's a u64, the stream could never receive that much data. */
+    stream->thread_data.window_size = aws_add_u64_saturating(stream->thread_data.window_size, pending_window_update);
+    if ((pending_window_update > 0) && (api_state == AWS_H1_STREAM_API_STATE_ACTIVE)) {
+        /* Now that stream window is larger, connection might have buffered
+         * data to send, or might need to increment its own window */
+        aws_h1_connection_try_process_read_messages(connection);
```
TingDaoK, line 91: It might be better to early stop here if the stream is not the current incoming stream?

tests/test_h1_client.c
```diff
@@ -2307,101 +2322,204 @@ H1_CLIENT_TEST_CASE(h1_client_request_close_header_with_chunked_encoding_and_pip
...
+    /* send 49 byte response 1 byte at a time, so we can see the message queue in action */
     const char *response_str = "HTTP/1.1 200 OK\r\n"
-                               "Content-Length: 9\r\n"
+                               "Content-Length: 10\r\n"
                                "\r\n"
-                               "Call Momo";
-    ASSERT_SUCCESS(testing_channel_push_read_str(&tester.testing_channel, response_str));
-
-    /* drain the task queue, in case there's an update window task in there from the headers */
+                               "0123456789";
```
TingDaoK, line 2472: Lol, I did count the length of "Call Momo" serval times.

source/h1_connection.c
graebm: Hmmmm.
Yeah, I'm always on the fence about using ASSERTS to ensure your math is doing what you intend, vs being safe but being unaware that your math is wrong.

Maybe I'll do both: ASSERT but also saturate
Nothing will explode if these particular numbers are off, but I'd love to fix it if they ever get that way

graebm: I went and did some more careful considerations of places I was just ASSERTing my math was all correct

source/h1_stream.c
graebm: eh, that's complicated. the function can handle it
doesn't seem worth optimizing for since it's pretty rare for a connection to have multiple streams

source/h1_connection.c
graebm: I went with: assert, but also saturate in case I'm wrong

source/h1_connection.c
graebm: Changed to actual error check

tests/CMakeLists.txt
```diff
@@ -122,7 +121,7 @@ add_test_case(h1_client_new_request_allowed)
 add_test_case(h1_client_midchannel_sanity_check)
 add_test_case(h1_client_midchannel_read)
 add_test_case(h1_client_midchannel_read_immediately)
-add_test_case(h1_client_midchannel_read_with_small_downstream_window)
+#add_test_case(h1_client_midchannel_read_with_small_downstream_window)
```
TingDaoK, line 124: Why disable this test?

graebm: whoa, good catch.
Turned it back on, it works.
I'd disabled all "window" tests at some point and forget to turn this back on

source/connection.c
```diff
@@ -848,6 +858,16 @@ int aws_http_client_connect_internal(
         goto error;
     }
 
+    if (options->manual_window_management && options->initial_window_size == 0 &&
```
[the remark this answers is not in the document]
graebm: yes.
I had to do that merge like 5 times because I suck at git
but yes, I did that

source/h1_stream.c
```diff
@@ -74,16 +79,62 @@ static void s_stream_cross_thread_work_task(struct aws_channel_task *task, void
...
+                should_schedule_task = true;
+            }
+        }
+
+        s_stream_unlock_synced_data(stream);
+    } /* END CRITICAL SECTION */
+
+    if (should_schedule_task) {
+        /* Keep stream alive until task completes */
+        aws_atomic_fetch_add(&stream->base.refcount, 1);
```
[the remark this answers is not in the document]
graebm: There is an assumption that the caller of this function is retaining a refcount on the stream

PR #292, opened 2020-10-12
undef debug token before we use it, so swiftpm doesn't break the build.

JonathanHenson:


source/h2_connection.c
```diff
@@ -17,6 +17,8 @@
 #    pragma warning(disable : 4204) /* non-constant aggregate initializer */
 #endif
 
+#undef DEBUG
```
graebm, line 20: comment why

graebm (approved):

PR #301, opened 2021-01-19
Attach tls handler in proper spot when doing nested tls

bretambrose:
* forward port (from ProxyRequestFlow) of tls slot fix when using nested tls

### DIFF
--- a/source/proxy_connection.c
+++ b/source/proxy_connection.c
@@ -448,12 +448,17 @@ static void s_aws_http_on_stream_complete_tls_proxy(struct aws_http_stream *stre
     context->state = AWS_PBS_TLS_NEGOTIATION;
     struct aws_channel *channel = aws_http_connection_get_channel(context->connection);
 
-    /*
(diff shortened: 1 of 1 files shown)


source/proxy_connection.c
```diff
@@ -448,12 +448,17 @@ static void s_aws_http_on_stream_complete_tls_proxy(struct aws_http_stream *stre
...
+    if (context->proxy_config->tls_options != NULL) {
+        /*
+         * If making secure (double TLS) proxy connection, we need to go after the second slot:
+         *
+         * Socket -> TLS(proxy) -> TLS(origin server) -> Http
+         */
+        left_of_tls_slot = left_of_tls_slot->adj_right;
+    }
+
+    if (channel == NULL || s_vtable->setup_client_tls(left_of_tls_slot, context->tls_options)) {
```
graebm, line 461: I don't think it's possible for `channel` to be NULL.
And if it was possible, line 451 is using channel before line 461 checks whether channel is NULL

graebm (approved):

PR #307, opened 2021-02-23
add timeout for integration test

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/integration-testing/http_client_test.py
+++ b/integration-testing/http_client_test.py
@@ -18,15 +18,21 @@ sys.argv = sys.argv[:1]
 def run_command(args):
     # gather all stderr and stdout to a single string that we print only if things go wrong
     process = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
-    output = process.communicate()[0]
-
-    if process.returncode != 0:
+    try:
+        # set a 100 sec timeout
+        output = process.communicate(timeout=100)[0]
+    except subprocess.TimeoutExpired:
+        process.kill()
         args_str = subprocess.list2cmdline(args)
-        print(args_str)
-        for line in output.splitlines():
-            print(line.decode())
+        raise RuntimeError("100 secs timeout happens from: {cmd}".format(cmd=args_str))
+    else:
+        if process.returncode != 0:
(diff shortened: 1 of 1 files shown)


integration-testing/http_client_test.py
```diff
@@ -18,15 +18,21 @@
...
     # gather all stderr and stdout to a single string that we print only if things go wrong
     process = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
-    output = process.communicate()[0]
-
-    if process.returncode != 0:
+    try:
+        # set a 100 sec timeout
+        output = process.communicate(timeout=100)[0]
+    except subprocess.TimeoutExpired:
+        process.kill()
```
graebm, line 25: Hmm I was worried that we're not showing the logs in the case of timeout, but the docs say you can call .communicate() again, they actually recommend it:

[docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate)
> The child process is not killed if the timeout expires, so in order to cleanup properly a well-behaved application should kill the child process and finish communication

(Check out the code sample in the link)

Maybe we can still get the full output? something like:

```
bool timedout = False
try:
    output = process.communicate(timeout=TIMEOUT)[0]
except subprocess.TimedoutExpired:
    timedout = True
    process.kill()
    output = process.communicate()[0]

if process.returncode != 0:
    ... all the same printing as before...
    if timedout:
        print("Process killed due to timeout after {} secs".format(TIMEOUT))

TingDaoK: Yeah, I saw the doc. Hmm, agreed on we should finish the communication, but probably the return code is not informative as we killed the process manually. Still think we need another RuntimeError in the end.

graebm: I was assuming the return code would be non-zero if the process was killed. You'd have to test to find out. or just make it: `if timedout or process.returncode != 0:`

TingDaoK: yeah, I have already tested it using aws-crt-nodejs. It's `-9`, if I remember it correctly, but `if timedout or process.returncode != 0` will be a better choice

graebm (approved):

PR #310, opened 2021-03-05
tests: Fix multiple definition of s_tester varaible.

Biswa96:
The same variable is used in test_h2_client.c and test_h1_server.c file.


bretambrose (commented):
What compiler/platform combination does this throw an error on?

Biswa96:
In Windows with gcc.

bretambrose (approved):

PR #321, opened 2021-04-29
More validation of HTTP/1.1 messages.

graebm:
Add validation for the following, according to the ABNF defined in RFC-7230.
- header field-names
- header field-values
- request method
- request-path aka URI
- response reason-phrase

Previously, we did an OK job validating incoming messages, but did no validation whatsoever of outgoing messages.

Caveats:
- URI validation is extremely basic for now.


justinboswell (approved):

TingDaoK (approved):
Change looks good. 
Just curious about the case you mentioned on duplicates `Content-length`, this change seems not helping?

PR #324, opened 2021-06-28
Remove http_connection_window_update

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- For h1, we don't support connection window
- For h2, we always automatically update the connection window and user can control the stream window.


TingDaoK:
Nope, we still need it

PR #327, opened 2021-07-12
aws_http_connection_update_window -> aws_http2_connection_update_window

TingDaoK:
*Issue #, if available:*

*Description of changes:*


include/aws/http/private/connection_impl.h
```diff
@@ -42,9 +42,9 @@ struct aws_http_connection_vtable {
     void (*close)(struct aws_http_connection *connection);
     bool (*is_open)(const struct aws_http_connection *connection);
     bool (*new_requests_allowed)(const struct aws_http_connection *connection);
-    void (*update_window)(struct aws_http_connection *connection, size_t increment_size);
 
     /* HTTP/2 specific functions */
+    int (*update_window)(struct aws_http_connection *connection, size_t increment_size);
```
TingDaoK, line 47: from `void` -> `int`. Also start to think about maybe it's better to have a return code for stream window update as well? It can fail for h2 as well. (Will never fail for h1)

include/aws/http/connection.h
```diff
@@ -548,6 +544,26 @@ int aws_http2_connection_get_received_goaway(
...
+ * the flow-control window of the connection will shrink as body data is received for all the streams created on it.
+ * (headers, padding, and other metadata do not affect the window).
+ * The initial connection flow-control window is 65,535.
+ * Once the connection's flow-control window reaches to 0, all the streams on the connection stop receiving any further
+ * data.
+ *
+ * If `manual_window_management` is false, this call will have no effect.
+ * The connection maintains its flow-control windows such that
+ * no back-pressure is applied and data arrives as fast as possible.
+ *
```
graebm, line 560: Document that if the connection is closed this call has no effect

Document that this function can only fail due to invalid input

We might want to do a pass like this over similar functions that act on the living connection. With the thinking that: we don't want to raise errors sometimes because the connection is closing, if the user wouldn't do anything in response to that error

TingDaoK: Checked all the similar functions, I think we are good.
We fail the call when connection is closed, but they make sense in those case. Eg:
1. ping, user will wait for ping complete callback if we don't fail it hard
2. send goaway. In case user wants allow more streams, but connection closed for another goaway that set a Last-Stream-ID, there will be misunderstanding if we don't fail the call.

TingDaoK: Talked offline, convinced that user will not have different actions when the goaway is not sent as the connection already closed.

include/aws/http/connection.h
```diff
@@ -548,6 +546,31 @@ int aws_http2_connection_get_received_goaway(
...
+ * If you are not connected, this call will have no effect.
+ *
+ * Fails when the input is invalid. The limit of the Maximum Size is 2147483647. If the increment size cause the
```
DavidOgunsAWS, line 565: nit: Most programmers would eyeball 2147483647 and know it's  the maximum value of 32-bit signed integer, but it's clearer to write it out or in the form `2**31 - 1`

source/h2_stream.c
```diff
@@ -406,6 +407,11 @@ static void s_stream_update_window(struct aws_http_stream *stream_base, size_t i
             "window size is 2147483647. We got %zu, which will cause the flow-control window to exceed the maximum",
             increment_size);
         aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+
+        struct aws_h2err returned_h2err = s_send_rst_and_close_stream(stream, aws_h2err_from_last_error());
+        if (aws_h2err_failed(returned_h2err)) {
+            aws_h2_connection_shutdown_due_to_write_err(connection, returned_h2err.aws_code);
+        }
```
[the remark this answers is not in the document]
graebm: Also, part of me thinks this should ALWAYS be how we deal with window overflow:
If stream window overflows: kill the stream
If connection window overflows: kill the connection

It's much more obvious to the user how to deal with something like that. Like, I'd be cool not letting *any* of these functions return `int` because *how do you deal with these errors, as a user*? But the user ALWAYS finds out about, and has a way to deal with, connections and streams dying unexpectedly.

source/h2_stream.c
```diff
@@ -22,6 +22,10 @@ static int s_stream_get_sent_error_code(struct aws_http_stream *stream_base, uin
 
 static void s_stream_cross_thread_work_task(struct aws_channel_task *task, void *arg, enum aws_task_status status);
 static struct aws_h2err s_send_rst_and_close_stream(struct aws_h2_stream *stream, struct aws_h2err stream_error);
+static int s_stream_reset_stream_internal(
+    struct aws_http_stream *stream_base,
+    uint32_t http2_error,
+    int aws_error_code);
```
graebm, line 28: would it look nicer here to pass `struct aws_h2err stream_error` instead of two separate error codes: `uint32_t http2_error, int aws_error_code`?

source/h2_stream.c
```diff
@@ -287,6 +291,7 @@ static void s_stream_cross_thread_work_task(struct aws_channel_task *task, void
     bool reset_called;
     size_t window_update_size;
     uint32_t user_reset_error_code;
+    int reset_aws_error_code;
```
graebm, line 294: it would definitely be less confusing to store the combined `struct aws_h2err reset_error_code`

instead of 2 separate variables

source/h2_stream.c
```diff
@@ -403,14 +411,20 @@ static void s_stream_update_window(struct aws_http_stream *stream_base, size_t i
...
             "The increment size is too big for HTTP/2 protocol, max flow-control "
-            "window size is 2147483647. We got %zu, which will cause the flow-control window to exceed the maximum",
+            "window size is 2**31 - 1. We got %zu, which will cause the flow-control window to exceed the maximum",
             increment_size);
-        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
-        return;
+        aws_raise_error(AWS_ERROR_OVERFLOW_DETECTED);
+
+        if (s_stream_reset_stream_internal(stream_base, AWS_HTTP2_ERR_INTERNAL_ERROR, AWS_ERROR_OVERFLOW_DETECTED))
+            aws_h2_connection_shutdown_due_to_write_err(connection, aws_last_error());
```
graebm, line 419: don't call `aws_h2_connection_shutdown_due_to_write_err()` from here, it's only meant to be called from the event-loop thread

you can just ignore any errors from s_stream_reset_stream(). it can only fail if the stream is already shutting down

source/h2_stream.c
```diff
@@ -403,14 +411,20 @@ static void s_stream_update_window(struct aws_http_stream *stream_base, size_t i
             ERROR,
             stream,
             "The increment size is too big for HTTP/2 protocol, max flow-control "
-            "window size is 2147483647. We got %zu, which will cause the flow-control window to exceed the maximum",
+            "window size is 2**31 - 1. We got %zu, which will cause the flow-control window to exceed the maximum",
             increment_size);
```
graebm, line 415: same confusion of sum vs increment_size, let's just simplify the message
> The stream's flow-control window has been incremented beyond 2**31 -1, the max for HTTP/2. The stream will close.

source/h2_stream.c
graebm: man ... makes me wish we'd named every helper function s_stream_synced_xxxxxxx() or s_stream_thread_xxxxxxx()

TingDaoK: yeah... Ryan did a good job to aws-c-s3, when I read the code there, I was hoping we did the same thing here😞 I'll see if I have time to do that

source/h2_stream.c
TingDaoK: Totally forgot we have `struct aws_h2err`

source/h2_stream.c
graebm: that's great to hear about the aws-c-s3 code. I haven't had the opportunity to dig in there yet

graebm (approved):
trivial fixes & ship

PR #333, opened 2021-08-23
Proxy envrionment variable support

TingDaoK:
*Issue #, if available:*

- Add support of set proxy from environment variable and integration test for proxy


bretambrose:
I strongly prefer an opt-in setting on the manager options to use the proxy environment variables.

source/connection_manager.c
```diff
@@ -808,10 +849,56 @@ struct aws_http_connection_manager *aws_http_connection_manager_new(
...
+                    aws_http_proxy_strategy_new_basic_auth(allocator, &config);
+                proxy_options->proxy_strategy = proxy_strategy;
+            }
+
+            if (proxy_options) {
+                options->proxy_options = proxy_options;
+                manager->proxy_config = aws_http_proxy_config_new_from_manager_options(allocator, options);
+                /* config has the deep copy of the needed options, we can clean up resources right after it */
+                options->proxy_options = NULL; /* it's nasty but simple */
+                if (proxy_options->proxy_strategy) {
```
bretambrose, line 888: You don't need to null check this release

source/connection_manager.c
```diff
@@ -808,10 +849,56 @@ struct aws_http_connection_manager *aws_http_connection_manager_new(
     }
 
     if (options->proxy_options) {
+        /* if proxy options are set, ignore the environment variable */
         manager->proxy_config = aws_http_proxy_config_new_from_manager_options(allocator, options);
         if (manager->proxy_config == NULL) {
             goto on_error;
         }
+    } else {
+        struct aws_http_proxy_options *proxy_options = NULL;
```
bretambrose, line 858: I would rather this entire else block go into its own function

source/connection_manager.c
```diff
@@ -808,10 +849,56 @@ struct aws_http_connection_manager *aws_http_connection_manager_new(
...
+        }
+        if (proxy_uri) {
+            proxy_options = aws_mem_calloc(allocator, 1, sizeof(struct aws_http_proxy_options));
+            proxy_options->host = proxy_uri->host_name;
+            proxy_options->port = proxy_uri->port;
+            /* Support basic authentication. */
+            if (proxy_uri->password.len) {
+                /* Has no empty password set */
+                struct aws_http_proxy_strategy_basic_auth_options config = {
+                    .proxy_connection_type = AWS_HPCT_HTTP_FORWARD,
```
bretambrose, line 874: forwarding Based on what?

More generally, what are the semantics of http_proxy vs. https_proxy?  How do you determine if this should use tunneling or forwarding (which doesn't get set here meaning it uses the legacy value which I doubt is correct behavior)?    Also if you're making a connection to an https proxy you need a tls context.

TingDaoK:
> I strongly prefer an opt-in setting on the manager options to use the proxy environment variables.

I don't really have a strong option on it, so, I'll do that. What about if the manger set to use the env variable and the option is also set. Should it be overwrite? I feel like we should still ignore the env variable in this case. What's your idea?

source/connection_manager.c
TingDaoK: Added a proxy env_var setting to the connection manager options that contains the connection type and tls options. Hmmm, but what if the tls options is not set? I checked the proxy test and seems like we never set the tls options for the proxy options?

bretambrose:
> > I strongly prefer an opt-in setting on the manager options to use the proxy environment variables.
> 
> I don't really have a strong option on it, so, I'll do that. What about if the manger set to use the env variable and the option is also set. Should it be overwrite? I feel like we should still ignore the env variable in this case. What's your idea?

manually set proxy options should always supercede environment settings I think.  That being said, if you make the connection manager control an enum, we could always add a third option which would take env over manual set if there's demand for it.

justinboswell (commented):
Looks good to me, but @bretambrose is Mr. Proxy.

source/connection_manager.c
```diff
@@ -756,6 +763,133 @@ static void s_schedule_connection_culling(struct aws_http_connection_manager *ma
...
+                    goto done;
+                }
+                aws_tls_connection_options_init_from_ctx(&default_tls_connection_options, tls_ctx);
+                /* tls options hold a ref to the ctx */
+                aws_tls_ctx_release(tls_ctx);
+                if (aws_tls_connection_options_set_server_name(
+                        &default_tls_connection_options, allocator, &proxy_uri.host_name)) {
+                    AWS_LOGF_ERROR(
+                        AWS_LS_HTTP_CONNECTION_MANAGER, "Failed set server name for TLS connection options.");
+                    goto done;
```
bretambrose, line 861: this leaks the tls context and connection options and proxy options

source/connection_manager.c
```diff
@@ -756,6 +763,133 @@ static void s_schedule_connection_culling(struct aws_http_connection_manager *ma
...
+                AWS_ZERO_STRUCT(tls_ctx_options);
+                /* create a default tls options */
+                aws_tls_ctx_options_init_default_client(&tls_ctx_options, allocator);
+                /* turn off verify peer for proxy connection setup by default */
+                aws_tls_ctx_options_set_verify_peer(&tls_ctx_options, false);
+                tls_ctx = aws_tls_client_ctx_new(allocator, &tls_ctx_options);
+                aws_tls_ctx_options_clean_up(&tls_ctx_options);
+                if (!tls_ctx) {
+                    AWS_LOGF_ERROR(AWS_LS_HTTP_CONNECTION_MANAGER, "Failed to create default TLS context.");
+                    goto done;
```
bretambrose, line 852: leaks proxy_options

source/connection_manager.c
```diff
@@ -756,6 +763,133 @@ static void s_schedule_connection_culling(struct aws_http_connection_manager *ma
...
+        }
+        /* Support basic authentication. */
+        if (proxy_uri.password.len) {
+            /* Has no empty password set */
+            struct aws_http_proxy_strategy_basic_auth_options config = {
+                .proxy_connection_type = proxy_options->connection_type,
+                .user_name = proxy_uri.user,
+                .password = proxy_uri.password,
+            };
+            struct aws_http_proxy_strategy *proxy_strategy = aws_http_proxy_strategy_new_basic_auth(allocator, &config);
```
bretambrose, line 874: could combine into a single assignment

source/connection_manager.c
```diff
@@ -756,6 +763,133 @@ static void s_schedule_connection_culling(struct aws_http_connection_manager *ma
...
+            proxy_options->proxy_strategy = proxy_strategy;
+        }
+        manager->proxy_config = aws_http_proxy_config_new_from_proxy_options(allocator, proxy_options);
+        /* config has the deep copy of the needed options, we can clean up resources right after it */
+        aws_tls_connection_options_clean_up(&default_tls_connection_options);
```
bretambrose, line 879: seems like all of these need to be move up to the top of the function and cleaned up unconditionally in done

include/aws/http/connection_manager.h
```diff
@@ -37,10 +53,32 @@ struct aws_http_connection_manager_options {
...
     struct aws_byte_cursor host;
     uint16_t port;
+    /* Proxy configuration for http connection */
+    const struct aws_http_proxy_options *proxy_options;
+    /*
+     * Optional.
+     * Configuration for using proxy from environment variable.
+     * Only works when proxy_options is not set. Zero out as default settings.
+     */
+    struct {
```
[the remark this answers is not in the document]
TingDaoK: Good call! Moved the logic to proxy.c and supported from raw http👍

source/proxy_connection.c
```diff
@@ -1085,6 +1129,95 @@ int aws_http_client_connect_via_proxy(const struct aws_http_client_connection_op
...
+    bool success = false;
+    if (s_proxy_uri_init_from_env_variable(options->allocator, options, &proxy_uri, &found)) {
+        /* Envrionment is set but failed to parse it */
+        goto done;
+    }
+    if (found) {
+        proxy_options.host = proxy_uri.host_name;
+        proxy_options.port = proxy_uri.port;
+        proxy_options.connection_type = options->proxy_ev_settings->connection_type;
+        if (!proxy_options.connection_type) {
```
bretambrose, line 1149: Let's make this an explicit comparison against an enum value.

source/proxy_connection.c
```diff
@@ -1085,6 +1129,95 @@ int aws_http_client_connect_via_proxy(const struct aws_http_client_connection_op
...
+            if (options->proxy_ev_settings->tls_options) {
+                proxy_options.tls_options = options->proxy_ev_settings->tls_options;
+            } else {
+                struct aws_tls_ctx *tls_ctx = NULL;
+                struct aws_tls_ctx_options tls_ctx_options;
+                AWS_ZERO_STRUCT(tls_ctx_options);
+                /* create a default tls options */
+                aws_tls_ctx_options_init_default_client(&tls_ctx_options, options->allocator);
+                /* turn off verify peer for proxy connection setup by default */
+                aws_tls_ctx_options_set_verify_peer(&tls_ctx_options, false);
```
bretambrose, line 1168: This is very questionable.  I imagine it's necessary for any integration tests to pass (since our tls proxy integration tests use a proxy that presents a signed localhost certificate which isn't going to pass server name validation) but for people using this in the wild with a real properly-set-up proxy, server name validation is critical.  I'm not sure how to resolve this, but we can't default it to off.

source/proxy_connection.c
```diff
@@ -1085,6 +1129,95 @@ int aws_http_client_connect_via_proxy(const struct aws_http_client_connection_op
...
+        if (!proxy_options.connection_type) {
+            if (options->tls_options) {
+                /* Use Tunneling when main connection use TLS. */
+                proxy_options.connection_type = AWS_HPCT_HTTP_TUNNEL;
+            } else {
+                /* Use forwarding proxy when main connection use clear text. */
+                proxy_options.connection_type = AWS_HPCT_HTTP_FORWARD;
+            }
+        }
+        if (aws_byte_cursor_eq_ignore_case(&proxy_uri.scheme, &aws_http_scheme_https)) {
```
bretambrose, line 1158: Can we split this (the tls handling) out into a separate function?

source/proxy_connection.c
bretambrose: One possibility is to allow for an environment variable to disable server name validation and set that in the integration test environment setup script.  It means we don't have a test for the "real" path where server name validation is done though, but that may be a TBD.

TingDaoK: Yeah! Totally make sense. Added another env variable. Set it up for tests as well.

source/proxy_connection.c
```diff
@@ -1150,6 +1150,7 @@ static int s_setup_proxy_tls_env_variable(
             proxy_verify_peer_string != NULL && aws_string_eq_c_str(proxy_verify_peer_string, "off")) {
```
bretambrose, line 1150: Can we change this to AWS_PROXY_NO_VERIFY_PEER and then just check for definition rather than the value itself.

bretambrose (approved):
Fix n ship

PR #335, opened 2021-09-07
Chunked trailer

ilevyor:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/aws/http/request_response.h
+++ b/include/aws/http/request_response.h
@@ -356,6 +356,20 @@ struct aws_http1_chunk_options {
     void *user_data;
 };
 
+struct aws_http1_trailer_options {
+
+    /**
+     * list of trailing headers to send as part of the chunked trailer
+     */
+
+    const struct aws_http_headers *trailing_headers;
+
+    /**
+     * User provided data passed to the on_complete callback on its invocation.
+     */
+    void *user_data;
+};
+
 #define AWS_HTTP_REQUEST_HANDLER_OPTIONS_INIT                                                                          \
     { .self_size = sizeof(struct aws_http_request_handler_options), }
 
@@ -627,6 +641,10 @@ void aws_http_message_set_body_stream(struct aws_http_message *message, struct a
 AWS_HTTP_API int aws_http1_stream_write_chunk(
     struct aws_http_stream *http1_stream,
     const struct aws_http1_chunk_options *options);
+
+AWS_HTTP_API int aws_http1_stream_write_chunked_trailer(
+    struct aws_http_stream *http1_stream,
+    const struct aws_http1_trailer_options *options);
 
 /**
  * Get the message's aws_http_headers.
--- a/source/h1_encoder.c
+++ b/source/h1_encoder.c
@@ -194,7 +194,8 @@ static size_t s_headers_size(const struct aws_http_headers *headers) {
         err |= aws_add_size_checked(header.value.len, total, &total);
         err |= aws_add_size_checked(4, total, &total); /* ": " + "\r\n" */
         if (err) {
-            return AWS_OP_ERR;
+            /* this message could be better, also should this fail in some way? */
+            AWS_LOGF_ERROR(AWS_LS_HTTP_STREAM, "failed to calculate header size");
         }
     }
     return total;
--- a/source/h1_encoder.c
+++ b/source/h1_encoder.c
@@ -496,6 +496,7 @@ struct aws_h1_trailer *aws_h1_trailer_new(
         aws_mem_release(allocator, trailer);
         return NULL;
     }
+    s_headers_to_buffer(&trailer->trailer_data, options->trailing_headers);
     trailer->allocator = allocator;
     return trailer;
 }
--- a/source/h1_encoder.c
+++ b/source/h1_encoder.c
@@ -181,6 +181,46 @@ static void s_write_headers(struct aws_byte_buf *dst, const struct aws_http_mess
     AWS_ASSERT(wrote_all);
 }
 
+/* @graebm would it be worthwhile to extract the inner part of this loop into a function shared by my headers count, and
+ * the scan function? */
+static size_t s_headers_size(const struct aws_http_header *headers) {
+    const size_t num_headers = aws_http_headers_get_header_count(headers);
+    size_t total = 0;
+    for (int i = 0; i < num_headers; i++) {
+        struct aws_http_header header;
+        aws_http_headers_get_index(headers, &header, i);
+        int err = 0;
+        err |= aws_add_size_checked(header.name.len, total, &total);
+        err |= aws_add_size_checked(header.value.len, total, &total);
+        err |= aws_add_size_checked(4, total, &total); /* ": " + "\r\n" */
+        if (err) {
+            return AWS_OP_ERR;
+        }
+    }
+    return total;
+}
+
+/* same as s_write_headers, but takes just the headers, instead of the whole message. @graebm can I delete write
+ * headers, and use this instead? */
+static void s_headers_to_buffer(struct aws_byte_buf *dst, const struct aws_http_header *headers) {
+
+    const size_t num_headers = aws_http_headers_get_header_count(headers);
+
+    bool wrote_all = true;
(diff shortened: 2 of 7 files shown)


source/h1_stream.c
```diff
@@ -225,11 +225,94 @@ static int s_stream_write_chunk(struct aws_http_stream *stream_base, const struc
...
+        s_stream_lock_synced_data(stream);
+        /* Can only add trailers while stream is active. */
+        if (stream->synced_data.api_state != AWS_H1_STREAM_API_STATE_ACTIVE) {
+            error_code = (stream->synced_data.api_state == AWS_H1_STREAM_API_STATE_INIT)
+                             ? AWS_ERROR_HTTP_STREAM_NOT_ACTIVATED
+                             : AWS_ERROR_HTTP_STREAM_HAS_COMPLETED;
+            goto unlock;
+        }
+
+        /* @graebm section 4.1 of the RFC suggests to me that this check is still necesssary, but I'm uncertain */
```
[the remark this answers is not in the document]
graebm: related, since this implementation only supports adding trailers 1 time, we would need a similar `has_added_trailer` bool that we check/set here

source/request_response.c
```diff
@@ -573,6 +573,23 @@ int aws_http1_stream_write_chunk(struct aws_http_stream *http1_stream, const str
...
+            (void *)http1_stream);
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
+
+    return http1_stream->vtable->http1_write_trailer(http1_stream, options);
+}
+
 struct aws_input_stream *aws_http_message_get_body_stream(const struct aws_http_message *message) {
     AWS_PRECONDITION(message);
     return message->body_stream;
```
graebm, line 595: gonna needs tests eventually too

source/h1_encoder.c
```diff
@@ -181,6 +181,47 @@ static void s_write_headers(struct aws_byte_buf *dst, const struct aws_http_mess
...
+    for (size_t i = 0; i < num_headers; i++) {
+        struct aws_http_header header;
+        aws_http_headers_get_index(headers, i, &header);
+        int err = 0;
+        err |= aws_add_size_checked(header.name.len, total, &total);
+        err |= aws_add_size_checked(header.value.len, total, &total);
+        err |= aws_add_size_checked(4, total, &total); /* ": " + "\r\n" */
+        if (err) {
+            /* this message could be better, also should this fail in some way? */
+            AWS_LOGF_ERROR(AWS_LS_HTTP_STREAM, "failed to calculate header size");
```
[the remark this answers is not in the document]
graebm: you might be curious why `void s_write_headers()` just uses AWS_ASSERT while `int s_scan_outgoing_headers()` can raise errors:

the reason is that the scan() functions is checking for all possible errors, and we pre-allocate size in the buffer that we write into with `s_write_headers()`, so it SHOULD be impossible for anything to go wrong in `s_write_headers()`, which is why we just use asserts to sanity-check for our own programming errors.

include/aws/http/request_response.h
```diff
@@ -628,6 +642,10 @@ AWS_HTTP_API int aws_http1_stream_write_chunk(
     struct aws_http_stream *http1_stream,
     const struct aws_http1_chunk_options *options);
 
+AWS_HTTP_API int aws_http1_stream_write_chunked_trailer(
```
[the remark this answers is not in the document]
ilevyor: So that's a good point, because I was trying to do the latter in my unit testing...

source/h1_encoder.c
```diff
@@ -181,6 +181,47 @@ static void s_write_headers(struct aws_byte_buf *dst, const struct aws_http_mess
     AWS_ASSERT(wrote_all);
 }
 
+/* @graebm would it be worthwhile to extract the inner part of this loop into a function shared by my headers count, and
+ * the scan function? */
+static size_t s_headers_size(const struct aws_http_headers *headers) {
```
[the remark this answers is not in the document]
ilevyor: Special headers are not allowed. Isn't that a reason to check for them though? To make sure that they are not sent in the trailer?

source/h1_encoder.c
```diff
@@ -443,6 +497,36 @@ static void s_populate_chunk_line_buffer(
...
+        aws_mem_release(allocator, trailer);
+        return NULL;
+    }
+    s_write_headers(&trailer->trailer_data, trailing_headers);
+    return trailer;
+}
+
+void aws_h1_trailer_destroy(struct aws_h1_trailer *trailer) {
+    /* should this check be here? it makes it so that I need to check before calling this function */
+    AWS_PRECONDITION(trailer);
```
graebm, line 525: for destroy() and release() calls, we usually add a check like:
```
if (trailer == NULL) {
    return;
}
```
so that they can be called blindly without checking if the ptr is null or not

so like, if destroy() worked that way, you could alter trailer_new() to do all its error handling like:
```
    if (step 1 goes wrong) {
        goto error;
    }

    if (step 2 goes wrong) {
        goto error;
    }

    if (step 3 goes wrong) {
        goto error;
    }

    ...
    /* success! */
    return trailer;
    
error:
    aws_h1_trailer_destroy(trailer);
    return NULL;
}
```

source/h1_encoder.c
```diff
@@ -181,6 +181,57 @@ static void s_write_headers(struct aws_byte_buf *dst, const struct aws_http_mess
     AWS_ASSERT(wrote_all);
 }
 
+static int s_headers_size(const struct aws_http_headers *headers, size_t *out_size) {
```
graebm, line 184: I have nit-picky feelings about how we can share this code for trailers and traditional headers but it would take too long to type out, let's just talk tomorrow

source/h1_encoder.c
```diff
@@ -748,11 +832,15 @@ static int s_state_fn_chunk_end(struct aws_h1_encoder *encoder, struct aws_byte_
...
+    bool done;
+    /* if a chunked trailer was set */
+    if (encoder->message->trailer) {
+        done = s_encode_buf(encoder, dst, &encoder->message->trailer->trailer_data);
+    } else {
+        done = s_write_crlf(dst);
+    }
     if (!done) {
-        /* Remain in this state until done writing out CRLF */
+        /* Remain in this state until we're done writing out body */
```
graebm, line 843: trivial: ~done writing out body~ -> done writing out trailer

tests/test_h1_client.c
```diff
@@ -326,6 +326,59 @@ H1_CLIENT_TEST_CASE(h1_client_request_send_body_chunked) {
...
 
+H1_CLIENT_TEST_CASE(h1_client_request_send_chunked_trailer) {
+    (void)ctx;
+    struct tester tester;
+    ASSERT_SUCCESS(s_tester_init(&tester, allocator));
+    struct aws_http_headers *trailers = aws_http_headers_new(allocator);
+    const struct aws_http_header trailer = {
+        .name = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("chunked"),
+        .value = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("trailer"),
+    };
```
graebm, line 337: trivial: add another header, since there's only 1 test may as well stress what you can

ilevyor: Good point, since this was actually broken for a while and the test didn't catch it. I caught it when running through the debugger. This debugger thing is pretty useful.

source/h1_encoder.c
```diff
@@ -443,6 +497,36 @@ static void s_populate_chunk_line_buffer(
...
+    const struct aws_http_headers *trailing_headers) {
+    /* Allocate trailer along with storage for the trailer-line */
+    size_t trailer_size = 0;
+    if (s_headers_size(trailing_headers, &trailer_size)) {
+        return NULL;
+    }
+
+    struct aws_h1_trailer *trailer = aws_mem_calloc(allocator, 1, sizeof(struct aws_h1_trailer));
+    trailer->allocator = allocator;
+    if (!trailer) {
```
TingDaoK, line 511: Check null before assigning. Otherwise, it will crash.

PR #338, opened 2021-09-22
Added aws_http2_stream_write_data, allowing H2 data frames to be written at any time

justinboswell:
Adds support for a new H2 API: `aws_http2_stream_write_data` which allows applications to supply data as it comes in, rather than supplying a stream that is constantly polled and is usually empty. This should allow use cases like event streams and transcribe to be efficient and scalable.


include/aws/http/request_response.h
```diff
@@ -356,6 +377,27 @@ struct aws_http1_chunk_options {
...
+/**
+ * Encoding options for manual H2 data frame writes
+ */
+struct aws_h2_data_options {
+    struct aws_input_stream *data;
+
+    /**
+     * Invoked when the data stream is no longer in use, whether or not it was successfully sent.
+     * Optional.
+     * See `aws_http1_stream_write_chunk_complete_fn`.
```
TingDaoK, line 391: trivial: copy/paste error

source/request_response.c
```diff
@@ -573,6 +573,21 @@ int aws_http1_stream_write_chunk(struct aws_http_stream *http1_stream, const str
...
+    AWS_PRECONDITION(h2_stream);
+    AWS_PRECONDITION(h2_stream->vtable);
+    AWS_PRECONDITION(options);
+    if (!h2_stream->vtable->http2_write_data) {
+        AWS_LOGF_TRACE(
+            AWS_LS_HTTP_STREAM,
+            "id=%p: H2 stream only function (write_data) invoked on other stream, ignoring call",
+            (void *)h2_stream);
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
```
TingDaoK, line 586: there is an old [pr](https://github.com/awslabs/aws-c-http/pull/327/files#diff-d7eec29a8674c2301bf0ced41b2720d9f20faa547a45348d4140915aa9dd323aR346), we want to enforce (otherwise crash) only h2 connection/stream used at those version specific functions. Mike convinced me that user cannot do anything to recover from this kind of error. We'd better just crash. So, maybe instead of handling it, just `AWS_PRECONDITION()`

include/aws/http/private/h2_connection.h
```diff
@@ -68,6 +68,14 @@ struct aws_h2_connection {
          * Waiting for WINDOW_UPDATE to set them free */
         struct aws_linked_list stalled_window_streams_list;
 
+        /* List using aws_h2_stream.node.
+         * Contains all streams that are open, but are only sending data when notified, rather than polling
+         * for it (e.g. event streams)
+         * Streams are moved to the outgoing_streams_list until they send pending data, then are moved back
+         * to this list to sleep until more data comes in
+         */
+        struct aws_linked_list evented_streams_list;
```
TingDaoK, line 77: Just unclear about what is this for, are we doing this to support event stream over h2? Or we also support passing more data to the stream overtime?

include/aws/http/request_response.h
```diff
@@ -233,6 +233,12 @@ struct aws_http_make_request_options {
      * See `aws_http_on_stream_complete_fn`.
      */
     aws_http_on_stream_complete_fn *on_complete;
+
+    /**
+     * When using h2, request body data will be provided over time. The body stream on the request will only be
+     * read when body data has been supplied via `aws_http2_stream_write_data`
+     */
+    bool h2_use_evented_body_stream;
```
TingDaoK, line 241: We have different `aws_http2_connection_options` and `aws_http1_connection_options` for connection, not sure if we want the similar stuff for request. For now, I don't think we will need it as I don't see any more request related options specific to version. But, just bring it up to table 😀

include/aws/http/private/h2_connection.h
justinboswell: Both, the idea is that the data is not available at request time, but is instead a long running connection with data available sparsely over time. Event streams, transcribe, etc.

source/request_response.c
justinboswell: Oh, sweet. I was just copying the local code, but I'll do that instead.

include/aws/http/request_response.h
justinboswell: Well, these wouldn't be connection options for H2, these are more stream/request related.

include/aws/http/request_response.h
justinboswell: Fixed.

include/aws/http/private/h2_connection.h
TingDaoK: Great! Mike and I talked about it during h2 work planing, adding more data to stream overtime is one of the todos. Thanks for handling that!

include/aws/http/request_response.h
TingDaoK: Oh, I was thinking about to have something like `aws_http1_make_request_options` and `aws_http2_make_request_options`, in case we will have more version specific options added. For now, I don't think we will need it as I don't see there will be much to add.

justinboswell: That makes sense, we could do that at any point.

source/h2_stream.c
```diff
@@ -629,6 +685,14 @@ int aws_h2_stream_encode_data_frame(
         return AWS_OP_SUCCESS;
     }
 
+    /* EOF/complete on input stream for manual writes just means move to the next one, the body
+     * never truly completes until the stream is closed
```
aajtodd, line 689: the input stream or the h2 stream? 

It looks like `body_complete` is always set to `false` for a "manual" write stream. I probably just missed it but where are we able to signal that the stream is truly done sending data? or does it not matter in practice?

include/aws/http/request_response.h
```diff
@@ -233,6 +233,12 @@ struct aws_http_make_request_options {
      * See `aws_http_on_stream_complete_fn`.
      */
     aws_http_on_stream_complete_fn *on_complete;
+
+    /**
+     * When using h2, request body data will be provided over time. The stream will only be polled for writing
+     * when data has been supplied via `aws_h2_stream_write_data`
+     */
+    bool h2_use_manual_data_writes;
```
aajtodd, line 241: Do we protect anywhere that this is valid (i.e. we actually got an H2 connection) if set to `true` after the connection is established?

aajtodd (commented):
I'll leave thorough review of the internals to the CRT team but overall looks good.

How are response streams handled? I'm assuming [aws_http_on_incoming_body_fn](https://github.com/awslabs/aws-c-http/blob/a48fa7e08f64bb80e2fb93f041f5bddd8bcf8c46/include/aws/http/request_response.h#L228) just continues to fire until the stream is closed by either the server or after the request side is closed?

source/h2_stream.c
justinboswell: this is the input stream.

It does not matter at all that we signal manually that the stream is done, because closing it will take care of that. Closing a stream takes care of closing off any open frames and sends the final frame to the server.

include/aws/http/request_response.h
justinboswell: it's more than it's only read by H2 code, the same way we do HTTP/1.1 chunk extensions. Also, this cannot be changed on the stream, this is only for the request, it's effectively temporary (only for the `make_request` call). The request will copy it into the stream if the stream is H2.

justinboswell:
> How are response streams handled? I'm assuming [aws_http_on_incoming_body_fn](https://github.com/awslabs/aws-c-http/blob/a48fa7e08f64bb80e2fb93f041f5bddd8bcf8c46/include/aws/http/request_response.h#L228) just continues to fire until the stream is closed by either the server or after the request side is closed?

That is correct, you'll just keep getting additional body frames.

graebm (commented):
stopped a little bit in. call me and we'll do a live review. more efficient way of exchanging what we know

source/h2_stream.c
```diff
@@ -999,3 +1067,30 @@ struct aws_h2err aws_h2_stream_on_decoder_rst_stream(struct aws_h2_stream *strea
...
+    struct aws_h2_stream_data_write *pending_write =
+        aws_mem_calloc(connection->base.alloc, 1, sizeof(struct aws_h2_stream_data_write));
+    pending_write->data_stream = options->data;
+    pending_write->on_complete = options->on_complete;
+    pending_write->user_data = options->user_data;
+
+    s_lock_synced_data(stream);
+    aws_linked_list_push_back(&stream->synced_data.pending_write_list, &pending_write->node);
+    /* If the stream is currently asleep */
+    if (stream->synced_data.waiting_for_writes) {
```
graebm, line 1088: need to check:
- using manual_write
- haven't called end_stream()
- stream is not complete <-- it might have died

maybe only accept if stream is in ACTIVE state?
if you accept in INIT state then you need to be sure the callback fires if cleanup happens without them ever having activated

if they think it's successful, they're waiting for the callback to happen

source/h2_stream.c
```diff
@@ -344,6 +357,17 @@ static void s_stream_destroy(struct aws_http_stream *stream_base) {
     aws_mutex_clean_up(&stream->synced_data.lock);
     aws_http_message_release(stream->thread_data.outgoing_message);
 
+    /* chuck the outgoing write on the pending list for unified clean up */
+    if (stream->thread_data.outgoing_write) {
+        aws_linked_list_push_front(&stream->synced_data.pending_write_list, &stream->thread_data.outgoing_write->node);
+    }
+    /* clean up any pending writes */
+    while (!aws_linked_list_empty(&stream->synced_data.pending_write_list)) {
```
graebm, line 365: we might also want to purge this list in on_stream_complete. because the stream/connection might have died, so why keep holding onto these resources. especially since someone's callback-userdata might have a reference to the stream itself

look at what we did purging http1 chunks

source/h2_stream.c
```diff
@@ -583,6 +607,38 @@ int aws_h2_stream_on_activated(struct aws_h2_stream *stream, bool *out_has_outgo
...
+    /* finish/clean up the current write operation */
+    struct aws_h2_stream_data_write *write_op = stream->thread_data.outgoing_write;
+    s_stream_data_write_destroy(stream, write_op);
+    stream->thread_data.outgoing_write = NULL;
+
+    s_lock_synced_data(stream);
+    {
+        /* check to see if there are more queued writes */
+        if (!aws_linked_list_empty(&stream->synced_data.pending_write_list)) {
+            /* Advance to the next write op for next time */
```
graebm, line 622: this all looks correct.
but maybe??? simpler to just have 2 pending_write_lists and you don't need the lock here

include/aws/http/request_response.h
```diff
@@ -356,6 +377,27 @@ struct aws_http1_chunk_options {
...
+struct aws_h2_data_write_options {
+    struct aws_input_stream *data;
+
+    /**
+     * Invoked when the data stream is no longer in use, whether or not it was successfully sent.
+     * Optional.
+     * See `aws_h2_stream_write_data_complete_fn`.
+     */
+    aws_h2_stream_write_data_complete_fn *on_complete;
+
```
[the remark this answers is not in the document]
graebm: I can't recall and we discussed it, and actually thing a separate end_stream() function isn't necessarily a great idea since it will probably add extra work, an empty frame, etc, and doesn't map 1:1 with the protocol as well

source/h2_stream.c
```diff
@@ -1054,3 +1197,136 @@ struct aws_h2err aws_h2_stream_on_decoder_rst_stream(struct aws_h2_stream *strea
...
+                s_stream_data_write_destroy(stream, pending_write, AWS_ERROR_INVALID_STATE);
+                AWS_LOGF_ERROR(
+                    AWS_LS_HTTP_STREAM,
+                    "Cannot write DATA frames to a stream after end, stream=%p",
+                    (void *)stream_base);
+                /* Fail with error, otherwise, people can wait for on_complete callback that will never be invoked. */
+                return aws_raise_error(AWS_ERROR_INVALID_STATE);
+            }
+            pending_write->on_complete = options->on_complete;
+            pending_write->user_data = options->user_data;
```
graebm, line 1237: add a comment about why you set this so late
/* Not setting this until we're sure we succeeded, so that callback doesn't fire on cleanup if we fail */

source/h2_stream.c
```diff
@@ -1054,3 +1197,136 @@ struct aws_h2err aws_h2_stream_on_decoder_rst_stream(struct aws_h2_stream *strea
...
+    /* queue this new write into the pending write list for the stream */
+    struct aws_h2_stream_data_write *pending_write =
+        aws_mem_calloc(stream->base.alloc, 1, sizeof(struct aws_h2_stream_data_write));
+    pending_write->data_stream = aws_input_stream_acquire(options->data);
+    pending_write->end_stream = false;
+    bool schedule_cross_thread_work = false;
+    { /* BEGIN CRITICAL SECTION */
+        s_lock_synced_data(stream);
+        {
+            if (stream->synced_data.api_state != AWS_H2_STREAM_API_STATE_ACTIVE) {
```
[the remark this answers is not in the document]
graebm: Add a warning to the public write_data() function that it's always possible for STREAM_HAS_COMPLETED to be raised if the stream ends unexpectedly

source/h2_stream.c
```diff
@@ -358,6 +416,35 @@ static void s_stream_destroy(struct aws_http_stream *stream_base) {
...
 }
 
+void aws_h2_stream_on_closed(struct aws_h2_stream *stream, int error_code) {
+    struct aws_linked_list pending_writes;
+    aws_linked_list_init(&pending_writes);
+    { /* BEGIN CRITICAL SECTION */
+        /* clean up any pending writes */
+        s_lock_synced_data(stream);
+        /* The stream is complete now, this will prevent further writes from being queued */
+        stream->synced_data.api_state = AWS_H2_STREAM_API_STATE_COMPLETE;
```
graebm, line 426: In every code path we traced the state was already set
I don't like it getting set again unnecessarily

graebm: here's an idea I like right now but feels a bit sketchy?

- DON'T grab the lock, the stream should be complete, the things we touch here should not be allowed to change anymore. leave a big comment about this though because it's weird
- ASSERT that api_state is complete or init

source/h2_stream.c
```diff
@@ -242,8 +249,6 @@ struct aws_h2_stream *aws_h2_stream_new_request(
     /* Stream refcount starts at 1, and gets incremented again for the connection upon a call to activate() */
     aws_atomic_init_int(&stream->base.refcount, 1);
 
-    /* Init H2 specific stuff */
-    stream->thread_data.state = AWS_H2_STREAM_STATE_IDLE;
     enum aws_http_version message_version = aws_http_message_get_protocol_version(options->request);
     switch (message_version) {
         case AWS_HTTP_VERSION_1_1:
```
graebm, line 254: unrelated to this PR:
Can we add a TODO to get rid of automatic transformation of HTTP/1 messages? It seems much better to have this be explicit. We discussed earlier today how the transformation may need to know more information that we don't have available right now (ex: schema)

source/h2_stream.c
```diff
@@ -266,11 +271,29 @@ struct aws_h2_stream *aws_h2_stream_new_request(
             goto error;
     }
 
+    /* Init H2 specific stuff */
+    stream->thread_data.state = AWS_H2_STREAM_STATE_IDLE;
+    /* stream end is implicit if the request isn't using manual data writes */
+    stream->synced_data.manual_write_ended = !options->http2_use_manual_data_writes;
+    stream->manual_write = options->http2_use_manual_data_writes;
+    /* init outgoing write queue */
```
graebm, line 279: this comment is unnecessary

source/h2_stream.c
```diff
@@ -560,7 +647,42 @@ struct aws_h2err aws_h2_stream_window_size_change(struct aws_h2_stream *stream,
...
+    struct aws_h2_stream_data_write *write = AWS_CONTAINER_OF(node, struct aws_h2_stream_data_write, node);
+    return write;
+}
+
+static struct aws_input_stream *s_h2_stream_get_data_stream(struct aws_h2_stream *stream) {
+    struct aws_h2_stream_data_write *write = s_h2_stream_get_current_write(stream);
+    return write->data_stream;
+}
+
+static bool s_h2_stream_current_write_ends_stream(struct aws_h2_stream *stream) {
```
graebm, line 680: trivial: s_h2_stream_does_current_write_end_stream()

source/h2_stream.c
```diff
@@ -600,8 +724,14 @@ int aws_h2_stream_on_activated(struct aws_h2_stream *stream, bool *out_has_outgo
         stream->thread_data.state = AWS_H2_STREAM_STATE_HALF_CLOSED_LOCAL;
         AWS_H2_STREAM_LOG(TRACE, stream, "Sending HEADERS with END_STREAM. State -> HALF_CLOSED_LOCAL");
     }
-
-    *out_has_outgoing_data = has_body_stream;
+    if (stream->manual_write) {
```
graebm, line 727: trivial: newline between these different if/else blocks

source/h2_stream.c
```diff
@@ -1054,3 +1197,136 @@ struct aws_h2err aws_h2_stream_on_decoder_rst_stream(struct aws_h2_stream *strea
...
+        AWS_H2_STREAM_LOG(TRACE, stream, "Scheduling stream cross-thread work task");
+        /* increment the refcount of stream to keep it alive until the task runs */
+        aws_atomic_fetch_add(&stream->base.refcount, 1);
+        aws_channel_schedule_task_now(connection->base.channel_slot->channel, &stream->cross_thread_work_task);
+    }
+
+    return AWS_OP_SUCCESS;
+}
+
+static int s_stream_end_stream_get_length(struct aws_input_stream *stream, int64_t *length) {
```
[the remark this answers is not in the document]
TingDaoK: Well, the input_stream_new_from cursor doesn't have the ownership of the buffer. It will just do a copy of cursor... So, maybe we still need this fake input_stream

source/h2_stream.c
```diff
@@ -346,10 +382,46 @@ static void s_stream_cross_thread_work_task(struct aws_channel_task *task, void
...
+    if (write->on_complete) {
+        write->on_complete(&stream->base, error_code, write->user_data);
+    }
+    if (write->data_stream) {
+        aws_input_stream_release(write->data_stream);
+    }
+    aws_mem_release(stream->base.alloc, write);
+}
+
+static void s_h2_stream_destroy_pending_writes(struct aws_h2_stream *stream, int error_code) {
```
graebm, line 401: s_h2_stream_destroy_pending_writes() doesn't need an error_code param, the rest of the streams will all complete with `AWS_ERROR_HTTP_STREAM_HAS_COMPLETED`

include/aws/http/private/h2_stream.h
```diff
@@ -53,6 +53,23 @@ enum aws_h2_stream_api_state {
     AWS_H2_STREAM_API_STATE_COMPLETE,
 };
 
+/* Indicates the state of the body of the HTTP/2 stream */
+enum aws_h2_stream_body_state {
+    AWS_H2_STREAM_BODY_STATE_NONE,           /* Has no body for the HTTP/2 stream */
```
graebm, line 58: ~NONE~ -> `NO_BODY`

source/h2_stream.c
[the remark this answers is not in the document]
graebm: still think we can get away with using aws_input_stream_new_from_cursor()

graebm (approved):
fix & ship

PR #342, opened 2021-10-12
refcount input stream

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/h1_encoder.c
```diff
@@ -551,12 +551,13 @@ struct aws_h1_chunk *aws_h1_chunk_new(struct aws_allocator *allocator, const str
     chunk->user_data = options->user_data;
     chunk->chunk_line = aws_byte_buf_from_empty_array(chunk_line_storage, chunk_line_size);
     s_populate_chunk_line_buffer(&chunk->chunk_line, options);
-
+    aws_input_stream_acquire(chunk->data);
```
graebm, line 554: trivial: do it like `chunk->data = aws_input_stream_acquire(options->chunk_data)` (needs to happen a few lines up)

include/aws/http/private/h1_stream.h
```diff
@@ -55,6 +55,9 @@ struct aws_h1_stream {
     /* Buffer for incoming data that needs to stick around. */
     struct aws_byte_buf incoming_storage_buf;
 
+    /* Keep the underlying message alive */
+    struct aws_http_message *http_message;
```
graebm, line 59: remove this. we didn't used to keep a pointer to the request from the stream, so we don't need to add one now

but we DO need to do reference counting on aws_h1_encoder_message->body

graebm (approved):

PR #345, opened 2021-10-19
HTTP/2 Connection pool

TingDaoK:
*Issue #, if available:*

*Description of changes:*

HTTP/2 connection pool.
For now, it's only the headers and the idea in the comment.

- We may need the load balance? 


TingDaoK:
We will create an easy http2 connection manager and a stream manager on the top for real http2

PR #350, opened 2021-12-01
connection window control

TingDaoK:
- Connection window control for http2 should be separate from the stream one
- Fixed a bug when converting http1 request to http2, the body is not used
- Sending Window Update for padding automatically 


include/aws/http/connection.h
```diff
@@ -215,6 +215,24 @@ struct aws_http2_connection_options {
...
+
+    /**
+     * Optional.
+     * Set to true to manually manage the flow-control window of whole HTTP/2 connection.
+     *
+     * If false, the connection will maintain its flow-control windows such that
+     * no back-pressure is applied and data arrives as fast as possible.
+     *
+     * If true, the flow-control window of the whole connection will shrink as body data
+     * is received (headers, padding, and other metadata do not affect the window) for every streams
```
graebm, line 227: Padding [DOES](https://httpwg.org/specs/rfc7540.html#DATA) affect it

> The entire DATA frame payload is included in flow control, including the **Pad Length** and **Padding** fields if present


And I don't think our API offers any way for users to know how much padding they're received 😐

source/h2_connection.c
```diff
@@ -300,8 +300,11 @@ static struct aws_h2_connection *s_connection_new(
     connection->base.http_version = AWS_HTTP_VERSION_2;
     /* Init the next stream id (server must use even ids, client odd [RFC 7540 5.1.1])*/
     connection->base.next_stream_id = (server ? 2 : 1);
+    /* Stream window management */
     connection->base.manual_window_management = manual_window_management;
```
graebm, line 304: trivial: it's a lot of changes, but worth renaming the internal variable ~base.manual_window_management~ -> `base.stream_manual_window_management`?

source/request_response.c
```diff
@@ -843,6 +843,7 @@ struct aws_http_message *aws_http2_message_new_from_http1(
     if (!message) {
         return NULL;
     }
+    AWS_LOGF_TRACE(AWS_LS_HTTP_GENERAL, "Creating HTTP/2 message from HTTP/1 message ip: %p", (void *)http1_msg)
```
graebm, line 846: trivial: ~ip~ -> `id`

source/request_response.c
```diff
@@ -881,6 +896,13 @@ struct aws_http_message *aws_http2_message_new_from_http1(
         if (aws_http_headers_add(copied_headers, aws_http_header_path, path_cursor)) {
             goto error;
         }
+        AWS_LOGF_TRACE(
+            AWS_LS_HTTP_GENERAL,
+            "Added header \"%.*s %.*s\" to new HTTP/2 header",
```
graebm, line 901: trivial: currently you have the name and value together in the same quotes, I suggest putting something between them. Also, move them to the end of the msg so the arbitrarily long header values fall at the end of the msg

currently this msg will look like:
`Added header ":scheme https" to new HTTP/2 header`

but maybe something like one of these reads better?
`Added header to new HTTP/2 message: ":scheme": "https"`
`Added header to new HTTP/2 message - ":scheme": "https"`
`Added header to new HTTP/2 message - :scheme = https`

include/aws/http/connection.h
TingDaoK: 🤦‍♂️Oh, shoot. Our comments is wrong all the time...
We shrink the windows with the whole length of the data frame, but users don't have anyway to know about the padding.....
So, I think we can offer:
- on_data_begin with the length of the whole payload?
- on_padding with the length of padding?
- We only shrink the length without padding (We send out window update for user when there is padding)? I don't like we do thing behind for user... But I think that's probably fine...

graebm (approved):

include/aws/http/private/h2_decoder.h
```diff
@@ -50,7 +50,13 @@ struct aws_h2_decoder_vtable {
 
     /* For DATA frame: _begin() is called, then 0+ _i() calls, then _end().
      * No other decoder callbacks will occur in this time */
-    struct aws_h2err (*on_data_begin)(uint32_t stream_id, uint32_t payload_len, bool end_stream, void *userdata);
+    struct aws_h2err (*on_data_begin)(
+        uint32_t stream_id,
+        uint32_t payload_len,          /* Whole payload length including padding and padding length */
+        uint32_t auto_managed_win_len, /* The length of padding and the byte for padding length, which is the
```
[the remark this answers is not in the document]
TingDaoK: renamed to `total_padding_bytes`, which implies the byte for padding length is counted.

source/h2_connection.c
```diff
@@ -1162,33 +1187,44 @@ struct aws_h2err s_decoder_on_data_begin(uint32_t stream_id, uint32_t payload_le
         return aws_h2err_from_h2_code(AWS_HTTP2_ERR_FLOW_CONTROL_ERROR);
     }
 
+    if (auto_managed_win_len) {
+        /* Update the padding for user, as we don't offer user any info about padding */
+        if (s_connection_send_update_window(connection, auto_managed_win_len)) {
+            return aws_h2err_from_last_error();
+        }
+        CONNECTION_LOGF(
+            INFO,
```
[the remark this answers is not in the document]
TingDaoK: I thought this is something we did differently than spec, so the level should be higher than usual trace. Will put it `DEBUG`

TingDaoK:
- Connection window control for http2 should be separate from the stream one
- Fixed a bug when converting http1 request to http2, the body is not used
- Sending Window Update for padding automatically

PR #359, opened 2022-01-21
Improve ref count semantics on http message

bretambrose:

### DIFF
--- a/include/aws/http/request_response.h
+++ b/include/aws/http/request_response.h
@@ -617,14 +617,14 @@ struct aws_http_message *aws_http2_message_new_response(struct aws_allocator *al
  * aws_http_message_release() is called by all those with a hold on it.
  */
 AWS_HTTP_API
-void aws_http_message_acquire(struct aws_http_message *message);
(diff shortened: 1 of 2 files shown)


sbSteveK (approved):

include/aws/http/request_response.h
```diff
@@ -615,16 +615,21 @@ struct aws_http_message *aws_http2_message_new_response(struct aws_allocator *al
 /**
  * Acquire a hold on the object, preventing it from being deleted until
  * aws_http_message_release() is called by all those with a hold on it.
+ *
+ * This function returns the passed in message (possibly NULL) so that acquire-and-assign can be done with a single
```
TingDaoK, line 619: It feels a program error to acquire on a null message to me.

bretambrose: Well that's the ref count semantics that I think we should use.  The point is to make using acquire/release not require conditional logic.  If something should be non-null feel free to check it.  But acquire/release should not crash and should have the semantics that this uses.  Same reasoning as the uh private PR from this afternoon.

graebm (approved):

TingDaoK (approved):

PR #360, opened 2022-01-25
ConnectionManager HTTP/2 updates

TingDaoK:
- Fix a bug about goaway received. Not release a connection that's in used.
- Don't vend connection to user until the settings done. To help user knows that the connection is really ready to use.

TODO: 
- Does it worth to put tests about http2 stuff here as we also test the stream manager stuff for http2 as the user of connection manager? I think we can just add todo here...

### DIFF
--- a/source/connection_manager.c
+++ b/source/connection_manager.c
@@ -1423,8 +1423,6 @@ static void s_aws_http_connection_manager_on_connection_setup(
 
     aws_mutex_lock(&manager->lock);
 
-    bool is_shutting_down = manager->state == AWS_HCMST_SHUTTING_DOWN;
-
     AWS_FATAL_ASSERT(manager->internal_ref[AWS_HCMCT_PENDING_CONNECTIONS] > 0);
     s_connection_manager_internal_ref_decrease(manager, AWS_HCMCT_PENDING_CONNECTIONS, 1);
     if (!error_code) {
--- a/source/connection_manager.c
+++ b/source/connection_manager.c
@@ -259,8 +259,8 @@ struct aws_http_connection_manager {
     /*
      * Internal refcount that keeps connection manager alive.
      *
-     * It's a sum of external_ref_count, vended_connection_count, pending_connects_count and open_connection_count,
-     * besides the `struct aws_connection_management_transaction` alive.
+     * It's a sum of vended_connection_count, pending_connects_count and open_connection_count,
+     * besides the `struct aws_connection_management_transaction` alive and one for any external usage.
      *
      * Once this refcount drops to zero, connection manager should either be cleaned up all the memory all waiting for
      * the last task to clean un the memory and do nothing else.
@@ -941,7 +941,6 @@ on_error:
 }
 
 void aws_http_connection_manager_acquire(struct aws_http_connection_manager *manager) {
-    aws_ref_count_acquire(&manager->internal_ref_count);
     aws_mutex_lock(&manager->lock);
     AWS_FATAL_ASSERT(manager->external_ref_count > 0);
     manager->external_ref_count += 1;
@@ -966,6 +965,7 @@ void aws_http_connection_manager_release(struct aws_http_connection_manager *man
                 (void *)manager);
             manager->state = AWS_HCMST_SHUTTING_DOWN;
             s_aws_http_connection_manager_build_transaction(&work);
+            aws_ref_count_release(&manager->internal_ref_count);
         }
     } else {
         AWS_LOGF_ERROR(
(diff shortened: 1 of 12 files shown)


source/connection_manager.c
```diff
@@ -61,6 +63,13 @@ bool aws_http_connection_manager_system_vtable_is_valid(const struct aws_http_co
 
 enum aws_http_connection_manager_state_type { AWS_HCMST_UNINITIALIZED, AWS_HCMST_READY, AWS_HCMST_SHUTTING_DOWN };
 
+/* vended_connection_count, pending_connects_count and open_connection_count */
+enum aws_http_connection_manager_count_type {
+    AWS_HCMCT_VENDED_CONNECTION,
+    AWS_HCMCT_PENDING_CONNECTIONS,
+    AWS_HCMCT_OPEN_CONNECTION,
```
graebm, line 70: comment what OPEN (how it sometimes but not always overlaps with VENDED)

source/connection_manager.c
```diff
@@ -265,6 +292,7 @@ struct aws_http_connection_manager_snapshot {
     size_t idle_connection_count;
     size_t pending_acquisition_count;
     size_t pending_connects_count;
+    size_t pending_settings_count;
     size_t vended_connection_count;
     size_t open_connection_count;
```
graebm, line 297: if any of these are the the sum of others please comment

really they could all use a nice 5 word description

source/connection_manager.c
```diff
@@ -523,11 +527,59 @@ static void s_aws_connection_management_transaction_init(
...
+    }
+}
+
+static void s_connection_manager_count_decrease(
+    struct aws_http_connection_manager *manager,
+    enum aws_http_connection_manager_count_type count_type,
+    size_t num) {
+    switch (count_type) {
+        case AWS_HCMCT_VENDED_CONNECTION:
+            manager->vended_connection_count -= num;
```
graebm, line 569: maybe do something to group or name or make clear that these 3 counts are special and you need to call this helper function?

source/connection_manager.c
```diff
@@ -523,11 +527,59 @@ static void s_aws_connection_management_transaction_init(
...
+            break;
+        default:
+            AWS_ASSERT(false);
+    }
+    for (size_t i = 0; i < num; i++) {
+        aws_ref_count_acquire(&manager->internal_ref_count);
+    }
+}
+
+static void s_connection_manager_count_decrease(
```
graebm, line 563: commenting that this should only ever be called during a transaction, and it can never be the thing that brings refounct to 0

source/connection_manager.c
```diff
@@ -1262,11 +1316,81 @@ static void s_aws_http_connection_manager_h2_on_goaway_received(
...
     struct aws_connection_management_transaction work;
     s_aws_connection_management_transaction_init(&work, manager);
-    /* Release this connection as no new stream will be allowed */
-    work.connection_to_release = http2_connection;
+
     aws_mutex_lock(&manager->lock);
+    /* Goaway received, remove the connection from idle and release it, if it's there. But, not decrease the
+     * open_connection_count as the shutdown callback will be invoked, we still need the manager to be alive */
+    const struct aws_linked_list_node *end = aws_linked_list_end(&manager->idle_connections);
+    for (struct aws_linked_list_node *node = aws_linked_list_begin(&manager->idle_connections); node != end;
```
graebm, line 1324: this is O(N) but you say there's precedent so 🤷‍♀️

source/connection_manager.c
```diff
@@ -1262,11 +1316,81 @@ static void s_aws_http_connection_manager_h2_on_goaway_received(
...
+
+    aws_mutex_lock(&manager->lock);
+
+    bool is_shutting_down = manager->state == AWS_HCMST_SHUTTING_DOWN;
+
+    AWS_FATAL_ASSERT(manager->pending_settings_count > 0);
+    --manager->pending_settings_count;
+
+    if (!error_code) {
+        if (is_shutting_down || s_idle_connection(manager, http2_connection)) {
```
graebm, line 1365: is it possible to share this code with the setup callback?

source/connection_manager.c
TingDaoK: Hmmm, I realize that they all have detailed description from the connection manger structure. Like [here](https://github.com/awslabs/aws-c-http/pull/360/files#diff-60eb9519b6fcfa87a65c15d68252caf78293eaff955953d4da882c3876186354R204). So, maybe we should be good here.

graebm (approved):

PR #366, opened 2022-03-10
Fix refcount issue

TingDaoK:
*Issue #, if available:*

- While connection manager is closing, if the culling task runs, it will increase the refcount back from zero, and decrease it after the task, which will let the clean up happen twice.
- A bug while acquire the external refcount should NOT acquire the internal refcount.

*Description of changes:*

- Culling task hold internal refcount, and cancel the task when the external refcount drops to zero instead of the internal refcount.

### DIFF
--- a/source/connection_manager.c
+++ b/source/connection_manager.c
@@ -1524,13 +1524,9 @@ static void s_cull_idle_connections(struct aws_http_connection_manager *manager)
                 (void *)manager,
                 (void *)current_idle_connection->connection);
         }
-        s_aws_http_connection_manager_get_snapshot(manager, &work.snapshot);
-    } else {
-        AWS_LOGF_TRACE(
-            AWS_LS_HTTP_CONNECTION_MANAGER,
-            "id=%p: culling idle connections while manager shutting down. Do nothing",
-            (void *)manager);
     }
+
+    s_aws_http_connection_manager_get_snapshot(manager, &work.snapshot);
 
     aws_mutex_unlock(&manager->lock);
 
--- a/source/connection_manager.c
+++ b/source/connection_manager.c
@@ -176,6 +176,7 @@ struct aws_http_connection_manager {
     /*
      * The number of all established, idle connections.  So
      * that we don't have compute the size of a linked list every time.
+     * It doesn't contribute to internal refcount as AWS_HCMCT_OPEN_CONNECTION inclues all idle connections as well.
      */
     size_t idle_connection_count;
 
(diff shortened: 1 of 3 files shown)


source/connection_manager.c
```diff
@@ -726,37 +730,14 @@ static void s_final_destruction_task(struct aws_task *task, void *arg, enum aws_
     struct aws_http_connection_manager *manager = arg;
```
graebm, line 730: comment a few lines up needs to change "If there's no cull task we just destroy the
* manager directly without a cross-thread task" isn't true anymore

source/connection_manager.c
```diff
@@ -767,11 +748,9 @@ static void s_schedule_connection_culling(struct aws_http_connection_manager *ma
 
     if (manager->cull_task == NULL) {
         manager->cull_task = aws_mem_calloc(manager->allocator, 1, sizeof(struct aws_task));
-        if (manager->cull_task == NULL) {
-            return;
-        }
-
         aws_task_init(manager->cull_task, s_cull_task, manager, "cull_idle_connections");
+        /* For the task to properly run and cancel, we need to keep manager alive */
+        aws_ref_count_acquire(&manager->internal_ref_count);
```
graebm, line 753: there's error handling at end of function that would clean up cull_task

either:
- decrement the refcount there
- OR the error is actually impossible now that allocations can't fail, so just remove error-handling altogether

source/connection_manager.c
```diff
@@ -783,6 +762,8 @@ static void s_schedule_connection_culling(struct aws_http_connection_manager *ma
     }
 
     uint64_t cull_task_time = 0;
+
+    aws_mutex_lock(&manager->lock);
```
graebm, line 766: there's error handling that will `goto` out of here without releasing the lock

source/connection_manager.c
```diff
@@ -958,6 +940,14 @@ void aws_http_connection_manager_release(struct aws_http_connection_manager *man
                 (void *)manager);
             manager->state = AWS_HCMST_SHUTTING_DOWN;
             s_aws_http_connection_manager_build_transaction(&work);
+            if (manager->cull_event_loop != NULL) {
+                /* When manager shutting down, schedule the task to cancel the cull task if exist. */
+                AWS_FATAL_ASSERT(manager->cull_task);
```
graebm, line 947: trivial: makes more sense to me to check if the `cull_task != NULL`

source/connection_manager.c
```diff
@@ -1187,14 +1177,8 @@ void aws_http_connection_manager_acquire_connection(
...
-    if (manager->state != AWS_HCMST_READY) {
-        AWS_LOGF_ERROR(
-            AWS_LS_HTTP_CONNECTION_MANAGER,
-            "id=%p: Acquire connection called when manager in shut down state",
-            (void *)manager);
-
-        request->error_code = AWS_ERROR_HTTP_CONNECTION_MANAGER_INVALID_STATE_FOR_ACQUIRE;
-    }
+    /* It's a use after free crime, we don't want to handle */
+    AWS_FATAL_ASSERT(manager->state == AWS_HCMST_READY);
```
graebm, line 1183: 🤣

graebm (approved):
🚀

PR #368, opened 2022-03-21
Update to latest builder

bretambrose:
Update to latest builder


.github/workflows/ci.yml
```diff
@@ -23,8 +23,6 @@ jobs:
     strategy:
       matrix:
         image:
-          - manylinux1-x64
-          - manylinux1-x86
```
TingDaoK, line 27: We talked with CLI team, as they still support manylinux1, it will be nice for us to have test on it until they drop it(Will be sometime later this year). So, maybe add manylinux1 back to builder?

TwistedTwigleg (approved):

PR #370, opened 2022-03-31
Metric for manager

TingDaoK:
Provide API to fetch metric for managers

### DIFF
--- a/include/aws/http/connection_manager.h
+++ b/include/aws/http/connection_manager.h
@@ -25,6 +25,14 @@ typedef void(aws_http_connection_manager_on_connection_setup_fn)(
 
 typedef void(aws_http_connection_manager_shutdown_complete_fn)(void *user_data);
 
+struct aws_http_manager_metric {
+    /* The number of additional concurrent requests that can be supported by the HTTP manager without needing to
+     * establish additional connections to the target server. */
+    size_t available_concurrency;
+    /* The number of requests that are awaiting concurrency to be made available from the HTTP manager. */
+    size_t pending_concurrency_acquires;
+};
+
 /*
  * Connection manager configuration struct.
  *
@@ -142,6 +150,14 @@ AWS_HTTP_API
 int aws_http_connection_manager_release_connection(
     struct aws_http_connection_manager *manager,
     struct aws_http_connection *connection);
+
+/**
+ * Fetch the current manager metric from connection manager.
+ */
+AWS_HTTP_API
+void aws_http_connection_manager_fetch_metric(
+    struct aws_http_connection_manager *manager,
+    struct aws_http_manager_metric *out_metric);
 
 AWS_EXTERN_C_END
 
--- a/include/aws/http/http2_stream_manager.h
+++ b/include/aws/http/http2_stream_manager.h
@@ -18,6 +18,7 @@ struct proxy_env_var_settings;
 struct aws_http2_setting;
 struct aws_http_make_request_options;
 struct aws_http_stream;
+struct aws_http_manager_metric;
 
 /**
  * Always invoked asynchronously when the stream was created, successfully or not.
--- a/include/aws/http/http2_stream_manager.h
+++ b/include/aws/http/http2_stream_manager.h
@@ -160,5 +160,16 @@ void aws_http2_stream_manager_acquire_stream(
     struct aws_http2_stream_manager *http2_stream_manager,
     const struct aws_http2_stream_manager_acquire_stream_options *acquire_stream_option);
 
+/**
+ * Fetch the current metric from stream manager.
+ *
+ * @param http2_stream_manager
+ * @param out_metric The metric to be fetched
+ */
+AWS_HTTP_API
+void aws_http2_stream_manager_fetch_metric(
+    struct aws_http2_stream_manager *http2_stream_manager,
+    struct aws_http_manager_metric *out_metric);
+
 AWS_EXTERN_C_END
 #endif /* AWS_HTTP2_STREAM_MANAGER_H */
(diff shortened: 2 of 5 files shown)


include/aws/http/connection_manager.h
```diff
@@ -25,6 +25,14 @@ typedef void(aws_http_connection_manager_on_connection_setup_fn)(
 
 typedef void(aws_http_connection_manager_shutdown_complete_fn)(void *user_data);
 
+struct aws_http_manager_metric {
+    /* The number of additional concurrent requests that can be supported by the HTTP manager without needing to
+     * establish additional connections to the target server. */
+    size_t available_concurrency;
```
graebm, line 31: ask java team what they want here:

"The number of additional concurrent requests that can be supported by the HTTP client without needing to establish"

or

"For HTTP/2 operations, this is equal to the number of streams that are currently idle/unused."

because we make new connections when an "ideal" number is reached, not just when a "max" number is reached

include/aws/http/connection_manager.h
```diff
@@ -143,6 +151,14 @@ int aws_http_connection_manager_release_connection(
     struct aws_http_connection_manager *manager,
     struct aws_http_connection *connection);
 
+/**
+ * Fetch the current manager metric from connection manager.
+ */
+AWS_HTTP_API
+void aws_http_connection_manager_fetch_metric(
+    struct aws_http_connection_manager *manager,
```
graebm, line 159: const

include/aws/http/http2_stream_manager.h
```diff
@@ -160,5 +161,16 @@ void aws_http2_stream_manager_acquire_stream(
...
 
+/**
+ * Fetch the current metric from stream manager.
+ *
+ * @param http2_stream_manager
+ * @param out_metric The metric to be fetched
+ */
+AWS_HTTP_API
+void aws_http2_stream_manager_fetch_metric(
+    struct aws_http2_stream_manager *http2_stream_manager,
```
graebm, line 172: const

include/aws/http/private/http2_stream_manager_impl.h
```diff
@@ -151,6 +151,11 @@ struct aws_http2_stream_manager {
         size_t internal_refcount_stats[AWS_SMCT_COUNT];
 
         bool finish_pending_stream_acquisitions_task_scheduled;
+
+        /**
+         * Number of streams can be made without make any new connections. Only used for metric purpose.
+         */
+        size_t available_concurrency;
```
graebm, line 158: naming: I like `metric_available_concurrency`. helps emphasize that we're not using this to drive behavior

source/http2_stream_manager.c
```diff
@@ -621,6 +626,9 @@ static void s_sm_connection_on_scheduled_stream_finishes(
             aws_random_access_set_remove(&stream_manager->synced_data.ideal_available_set, sm_connection);
             work.sm_connection_to_release = sm_connection;
             --stream_manager->synced_data.holding_connections_count;
+            /* Released one connection, decrease the available concurrency by the number of streams that connection can
+             * handle */
+            stream_manager->synced_data.available_concurrency -= sm_connection->max_concurrent_streams;
```
graebm, line 631: radical idea: what if we just grabbed a lock in the gather_metrics() function and calculated everything there

that would avoid any bugs from careful adding/subtracting at exactly the right moments

and simplify the code overall, less variables, etc

include/aws/http/connection_manager.h
```diff
@@ -25,6 +25,23 @@ typedef void(aws_http_connection_manager_on_connection_setup_fn)(
...
+ * Metrics for logging and debugging purpose.
+ */
+struct aws_http_manager_metric {
+    /**
+     * The number of additional concurrent requests that can be supported by the HTTP manager without needing to
+     * establish additional connections to the target server.
+     *
+     * For connection manager, it equals to connections that's idle.
+     * For stream manager, it equals to the number of streams that are possiblily made without creating new connection.
+     * Altough the implementation can create new connection without fully filling it.
```
graebm, line 38: trivial: spelling:
- possiblily
- Altough

include/aws/http/connection_manager.h
```diff
@@ -25,6 +25,23 @@ typedef void(aws_http_connection_manager_on_connection_setup_fn)(
 
 typedef void(aws_http_connection_manager_shutdown_complete_fn)(void *user_data);
 
+/**
+ * Metrics for logging and debugging purpose.
+ */
+struct aws_http_manager_metric {
```
graebm, line 31: trivial/debatable: sounds weird to me as "metric". "metrics" sounds more natural.

include/aws/http/connection_manager.h
graebm: trivial: if you change "metric" -> "metrics" update this function name too

include/aws/http/http2_stream_manager.h
graebm: trivial: is this non-const because you need to acquire the lock?
If so, I'd be OK with making this const, and then just casting away the const-ness when you need to futz with the locks.

In principal it's const

include/aws/http/connection_manager.h
```diff
@@ -161,12 +161,12 @@ int aws_http_connection_manager_release_connection(
...
- * Fetch the current manager metric from connection manager.
+ * Fetch the current manager metrics from connection manager.
  */
 AWS_HTTP_API
-void aws_http_connection_manager_fetch_metric(
-    struct aws_http_connection_manager *manager,
-    struct aws_http_manager_metric *out_metric);
+void aws_http_connection_manager_fetch_metrics(
+    const struct aws_http_connection_manager *manager,
+    struct aws_http_manager_metrics *out_metric);
```
graebm, line 169: trivial: ~out_metric~ -> out_metrics

some compilers issue warnings if the param name is different between the .h and .c file

graebm (approved):
fix & ship

PR #371, opened 2022-03-31
Fix our CI ignores branches that have "/" in name

TingDaoK:
Fix our CI ignore branches that have / in name

this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


graebm (approved):
I don't understand why, but this looks better regardless

PR #372, opened 2022-04-01
Flow control window update

TingDaoK:
- Changed the config for stream manager to support config of initial settings and connection manual window update.
- Optimize: When connection manual window management is disabled, update the connection window size to max to minimize the affect of flow control. Make it like no flow control at all.

TODO:
Maybe should do the similar thing that if stream level flow control is disabled, we should just add one initial setting to make the stream window initial from max.


include/aws/http/http2_stream_manager.h
```diff
@@ -56,20 +56,28 @@ struct aws_http2_stream_manager_options {
...
+     * arrives as fast as possible.
+     * - For connection level window control, `conn_manual_window_management` will enable manual control. The
+     * inital window size is not controlable.
+     * - For stream level window control, `enable_read_back_pressure` will enable manual control. The initail window
+     * size needs to be set through `initial_settings_array`.
+     */
+    struct aws_http2_setting *initial_settings_array;
+    size_t num_initial_settings;
+    size_t max_closed_streams;
+    bool conn_manual_window_management;
```
graebm, line 73: need comments for `bool conn_manual_window_management`

include/aws/http/http2_stream_manager.h
```diff
@@ -56,20 +56,28 @@ struct aws_http2_stream_manager_options {
...
+
     /**
      * HTTP/2 Stream window control.
      * If set to true, the read back pressure mechanism will be enabled for streams created.
      * The initial window size can be set through `initial window size`
```
graebm, line 78: trivial: "The initial window size can be set through `initial window size`" not accurate anymore

source/h2_connection.c
```diff
@@ -1722,6 +1722,17 @@ static void s_handler_installed(struct aws_channel_handler *handler, struct aws_
     /* enqueue the initial settings frame here */
     aws_linked_list_push_back(&connection->thread_data.outgoing_frames_queue, &init_settings_frame->node);
 
+    /* If not manual connection window management, update the connection window to max. */
+    if (!connection->conn_manual_window_management) {
+        uint32_t initial_window_update_size = AWS_H2_WINDOW_UPDATE_MAX - 1 - AWS_H2_INIT_WINDOW_SIZE;
```
graebm, line 1727: take away the `-1`
it's not needed
people are going to think it's needed

source/h2_connection.c
```diff
@@ -1722,6 +1722,17 @@ static void s_handler_installed(struct aws_channel_handler *handler, struct aws_
...
+        struct aws_h2_frame *connection_window_update_frame =
+            aws_h2_frame_new_window_update(connection->base.alloc, 0, initial_window_update_size);
```
graebm, line 1729: trivial: `0 /* stream_id */`

graebm (approved):

PR #375, opened 2022-05-04
More local host test

TingDaoK:
- Add a non-tls h2 server script and test our client with prior knowledge against it
- Stream Manager stress test against local host
- Test download and upload
  - download of 2.5G file works well
     - 85.95 sec for linux 
     - 58.53 sec for mac 
     - 62.65 sec for windows
  - Upload of 2.5G file is weirdly slow for linux.
  - Now using 0.25G upload
    - 162.37 sec for linux 
    - 6.43 sec for mac
    - 5.44 sec for windows
  - For upload performance, we tested against nginx server for mac and linux. We didn't see any performance issue around them. So, we assume the performance issue is related to hyper h2 implementation.
  - Upload 0.25G data to nginx and seems like nginx will store the data in disk as temp file.
    - 9.02 sec for mac
    - 4.00 sec for linux


.builder/action/aws-c-http-test.py
```diff
@@ -33,6 +32,8 @@ def run(self, env):
             # consumer
             os.chdir('../../aws-c-http')
 
+        # actions.append(['ctest', '--output-on-failure', '-R',
+        #                'localhost_integ_h2_*'])
```
graebm, line 36: ??
did you meant to add this
usually a bad practice to commit commented-out code

tests/py_localhost/README.md
```diff
@@ -1,12 +1,36 @@
 # Local server
 
-A local server based on the [example](https://github.com/python-hyper/h2/blob/master/examples/asyncio/asyncio-server.py) from hyper h2 server. Minor changed based on the example to response the headers of requests back within the headers from `/echo`.
+Local server based on [python-hyper/h2](https://github.com/python-hyper/h2).
 
 ## How to run the server
 
-Python 3.5+ required.
+Python 3.5+ required. Command here works for OSX and linux.
```
graebm, line 7: trivial: don't need this comment. I think it's understood that the exact command (python vs python3 vs python3.10 etc) might differ

tests/py_localhost/server.py
```diff
@@ -81,7 +86,12 @@ def data_received(self, data: bytes):
     def request_received(self, headers: List[Tuple[str, str]], stream_id: int):
         self.raw_headers = headers
         headers = collections.OrderedDict(headers)
+        path = headers[':path']
         method = headers[':method']
+        if method == "PUT" or method == "POST":
+            self.file_path = os.path.join(os.path.curdir, path[1:])
+            if os.path.exists(self.file_path):
+                os.remove(self.file_path)
```
graebm, line 94: and this is why we're not running a public webserver.
An attacker could pass a path like `/../../../../some_important_program` and replace files with whatever they want

But this isn't public, so we can do whatever hacky crap we want

graebm (approved):

.builder/action/aws-c-http-test.py
TingDaoK: forgot to remove the code for testing

PR #377, opened 2022-05-25
H2 monitor

TingDaoK:
- Add support of connection monitor for HTTP/2


source/h2_connection.c
```diff
@@ -874,6 +888,16 @@ static int s_encode_data_from_outgoing_streams(struct aws_h2_connection *connect
         aws_linked_list_push_back(outgoing_streams_list, aws_linked_list_pop_front(&stalled_streams_list));
     }
 
+    if (!pre_empty && aws_linked_list_empty(outgoing_streams_list)) {
```
graebm, line 891: trivial: move this block to be after the `if (aws_error_code)` block

source/h2_connection.c
```diff
@@ -2743,3 +2791,55 @@ static size_t s_handler_message_overhead(struct aws_channel_handler *handler) {
...
+     * If user is adding chunks 1 by 1, there can naturally be a gap in the upload.
+     * If the user lets the stream-window go to zero, there can naturally be a gap in the download. */
+    s_pull_up_stats_timestamps(connection);
+
+    void *stats_base = &connection->thread_data.stats;
+    aws_array_list_push_back(stats, &stats_base);
+    if (aws_hash_table_get_entry_count(&connection->thread_data.active_streams_map) != 0) {
+        /* The statistic has been reported, reset the state only if there is any active streams. */
+        connection->thread_data.stats.was_non_active = false;
+    }
```
graebm, line 2844: you said: I should remove this because it will get reset anyway

source/h2_connection.c
```diff
@@ -2743,3 +2791,55 @@ static size_t s_handler_message_overhead(struct aws_channel_handler *handler) {
...
+
+static void s_gather_statistics(struct aws_channel_handler *handler, struct aws_array_list *stats) {
+
+    struct aws_h2_connection *connection = handler->impl;
+    AWS_PRECONDITION(aws_channel_thread_is_callers_thread(connection->base.channel_slot->channel));
+
+    /* TODO: Need update the way we calculate statistics, to account for user-controlled pauses.
+     * If user is adding chunks 1 by 1, there can naturally be a gap in the upload.
+     * If the user lets the stream-window go to zero, there can naturally be a gap in the download. */
+    s_pull_up_stats_timestamps(connection);
```
graebm, line 2837: trivial: I vote you just make this one function instead of 2

source/h2_connection.c
```diff
@@ -2743,3 +2791,55 @@ static size_t s_handler_message_overhead(struct aws_channel_handler *handler) {
     /* "All frames begin with a fixed 9-octet header followed by a variable-length payload" (RFC-7540 4.1) */
     return 9;
 }
+
+static void s_reset_statistics(struct aws_channel_handler *handler) {
+    struct aws_h2_connection *connection = handler->impl;
+    aws_crt_statistics_http2_channel_reset(&connection->thread_data.stats);
+    return;
```
graebm, line 2798: check again here if we have active streams

source/h2_connection.c
```diff
@@ -1853,6 +1888,13 @@ static void s_move_stream_to_thread(
         goto error;
     }
 
+    if (aws_hash_table_get_entry_count(&connection->thread_data.active_streams_map) == 0) {
+        /* transition from nothing to read -> something to read */
+        uint64_t now_ns = 0;
+        aws_channel_current_clock_time(connection->base.channel_slot->channel, &now_ns);
+        connection->thread_data.incoming_timestamp_ns = now_ns;
+    }
```
graebm, line 1896: I said: move this later in the function so we only do it if everything's successful.

then you said we have issues we might have issues removing the stream and doing math but haven't set this

then I said: maybe we use 0 to mean "i got nothing" instead of checking whether the map is empty

source/h2_connection.c
```diff
@@ -1865,6 +1907,12 @@ static void s_move_stream_to_thread(
     }
 
     if (has_outgoing_data) {
+        if (aws_linked_list_empty(&connection->thread_data.outgoing_streams_list)) {
+            /* transition from nothing to write -> something to write */
+            uint64_t now_ns = 0;
+            aws_channel_current_clock_time(connection->base.channel_slot->channel, &now_ns);
+            connection->thread_data.outgoing_timestamp_ns = now_ns;
+        }
         aws_linked_list_push_back(&connection->thread_data.outgoing_streams_list, &stream->node);
```
graebm, line 1916: consider moving this to a helper function so that https://github.com/awslabs/aws-c-http/pull/338 can take advantage of it

source/h2_connection.c
TingDaoK: we still need to check the map is empty or not as the there could be multiple streams created and completed, only when we flip from something to read to nothing to read, we need to do the math, vise verse.

But, we could add a double check using 0.

source/h2_connection.c
TingDaoK: will move when we feel like need to

include/aws/http/private/h2_connection.h
```diff
@@ -108,6 +109,13 @@ struct aws_h2_connection {
         int channel_shutdown_error_code;
         bool channel_shutdown_immediately;
         bool channel_shutdown_waiting_for_goaway_to_be_written;
+
+        struct aws_crt_statistics_http2_channel stats;
```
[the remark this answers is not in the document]
graebm: Add a TODO about this

graebm (approved):

PR #378, opened 2022-05-31
More Protocol verification

TingDaoK:
- Add protocol verification for connection-specific headers ([RFC9113 8.2.2](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#name-connection-specific-header-))
- Add verification for content-length matching the received data payload length. ([RFC-9113 8.1.1](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#name-malformed-messages))


source/h2_decoder.c
```diff
@@ -1340,7 +1340,17 @@ static struct aws_h2err s_process_header_field(
                     return aws_h2err_from_last_error();
                 }
                 break;
-            /* TODO: Validate connection-specific header field (RFC7540 8.1.2.2) */
+            case AWS_HTTP_HEADER_TRANSFER_ENCODING:
+            case AWS_HTTP_HEADER_UPGRADE:
+            case AWS_HTTP_HEADER_KEEP_ALIVE:
+            case AWS_HTTP_HEADER_PROXY_CONNECTION: {
+                /* connection-specific header field are treated as malformed (RFC9113 8.2.2) */
+                DECODER_LOG(ERROR, decoder, "Connection-specific headers found.");
```
graebm, line 1348: be more clear that this is bad
"Connection-specific header found, not allowed in HTTP/2"

source/h2_decoder.c
```diff
@@ -1340,7 +1340,17 @@ static struct aws_h2err s_process_header_field(
...
+                DECODER_LOGF(
+                    DEBUG, decoder, "Connection-specific header name is '" PRInSTR "'", AWS_BYTE_CURSOR_PRI(name));
```
graebm, line 1350: doesn't seem like a huge secret to log the header name at error level. I mean, it's 1/4 possible names

source/h2_stream.c
```diff
@@ -749,13 +751,25 @@ struct aws_h2err aws_h2_stream_on_decoder_headers_i(
...
+
+                stream->base.client_data->response_status = (int)status_code;
+            } break;
+            case AWS_HTTP_HEADER_CONTENT_LENGTH: {
+                uint64_t content_length = 0;
+                if (aws_byte_cursor_utf8_parse_u64(header->value, &content_length)) {
+                    AWS_H2_STREAM_LOG(ERROR, stream, "Invalid content-length value");
+                    goto malformed;
+                }
+                stream->thread_data.content_length = (int64_t)content_length;
```
graebm, line 769: make sure this isn't larger than int64_t, or use uint64_t and make sure it's not exactly max-value
or use a bool instead of -1

source/h2_stream.c
```diff
@@ -871,6 +885,28 @@ struct aws_h2err aws_h2_stream_on_decoder_data_begin(
         return s_send_rst_and_close_stream(stream, aws_h2err_from_h2_code(AWS_HTTP2_ERR_PROTOCOL_ERROR));
     }
 
+    if (stream->thread_data.content_length != -1) {
+        stream->thread_data.received_data_length += payload_len; /* Should never exceed int64_t in real life */
+        if (end_stream && stream->thread_data.received_data_length != (uint64_t)stream->thread_data.content_length) {
```
graebm, line 890: just use uint64's all around and a separate bool and the u64_checked() functions so I can stop worrying about overflow

graebm: ALSO
check that we haven't exceeded it, even if !end_stream
AND
then if it is end_stream check that they're perfectly matched

graebm (commented):
YAY!

include/aws/http/private/h2_stream.h
```diff
@@ -69,6 +69,12 @@ struct aws_h2_stream {
         int64_t window_size_self;
         struct aws_http_message *outgoing_message;
         bool received_main_headers;
+
+        bool content_length_received;
+        /* The received content length, if any */
+        uint64_t content_length;
+        /* The total length of payload of data frame received */
+        uint64_t received_data_length;
```
graebm, line 77: make it clear that these variables are about the "incoming" message and not the "outgoing" message

maybe:
```
/* Set if incoming message has content-length header */
bool has_incoming_content_length;
uint64_t incoming_content_length;
/* The total length of payload of data frame received */
uint64_t incoming_data_length;
```

include/aws/http/request_response.h
```diff
@@ -612,6 +612,18 @@ struct aws_http_message *aws_http2_message_new_request(struct aws_allocator *all
...
+ * Create an HTTP/2 message from HTTP/1.1 message.
+ * pseudo headers will be created from the context and added to the headers of new message.
+ * Normal headers will be copied to the headers of new message.
+ * Note: if `host` exist, it will stay and `:authority` will be added using the information.
+ * `:scheme` is default to be "https". If a different scheme wants to be used, create the HTTP/2 message directly
+ */
+AWS_HTTP_API
+struct aws_http_message *aws_http2_message_new_from_http1(
+    struct aws_http_message *http1_msg,
+    struct aws_allocator *alloc);
```
graebm, line 625: trivial: we usually have allocator be the first argument to a new() function

source/h2_stream.c
```diff
@@ -749,13 +747,24 @@ struct aws_h2err aws_h2_stream_on_decoder_headers_i(
...
+                (void)err;
+
+                stream->base.client_data->response_status = (int)status_code;
+            } break;
+            case AWS_HTTP_HEADER_CONTENT_LENGTH: {
+                if (aws_byte_cursor_utf8_parse_u64(header->value, &stream->thread_data.content_length)) {
+                    AWS_H2_STREAM_LOG(ERROR, stream, "Invalid content-length value");
+                    goto malformed;
+                }
+                stream->thread_data.content_length_received = true;
```
graebm, line 764: check to ensure that we're not getting duplicate content-length headers. Request smuggling!

source/h2_stream.c
```diff
@@ -871,6 +880,38 @@ struct aws_h2err aws_h2_stream_on_decoder_data_begin(
...
+        if (end_stream && stream->thread_data.received_data_length != stream->thread_data.content_length) {
+            AWS_H2_STREAM_LOGF(
+                ERROR,
+                stream,
+                "Total received data payload=%" PRIu64
+                " does not match the received content-length header, which=%" PRIi64 ". Closing malformed stream",
+                stream->thread_data.received_data_length,
+                stream->thread_data.content_length);
+            return s_send_rst_and_close_stream(stream, aws_h2err_from_h2_code(AWS_HTTP2_ERR_PROTOCOL_ERROR));
+        }
```
graebm, line 912: we might need to do this check elsewhere
it's possible for on_data_begin to never fire
the message might declare content-length and then send END_STREAM in the HEADERS frame
maybe do the `!=` check in the end_stream callback?

also, check out this language:
> unless the message is defined as having no content.

which makes me think of a HEAD request where the response has content-length but no actual data :|
check how we do it in HTTP/1

include/aws/http/private/http_impl.h
```diff
@@ -64,6 +64,9 @@ enum aws_http_header_name {
     AWS_HTTP_HEADER_RETRY_AFTER,
     AWS_HTTP_HEADER_VARY,
     AWS_HTTP_HEADER_WARNING,
+    AWS_HTTP_HEADER_UPGRADE,
+    AWS_HTTP_HEADER_KEEP_ALIVE,
+    AWS_HTTP_HEADER_PROXY_CONNECTION,
```
graebm, line 69: Add a todo to our transformation function:

> An intermediary transforming an HTTP/1.x message to HTTP/2 MUST remove connection-specific header fields as discussed in [Section 7.6.1](https://rfc-editor.org/rfc/rfc9110#section-7.6.1) of [[HTTP](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#RFC9110)], or their messages will be treated by other HTTP/2 endpoints as [malformed](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#malformed)

source/request_response.c
```diff
@@ -893,7 +894,22 @@ struct aws_http_message *aws_http2_message_new_from_http1(
             aws_http_header_scheme.ptr,
             (int)scheme_cursor.len,
             scheme_cursor.ptr);
-        /* :authority SHOULD NOT be created when translating HTTP/1 request.(RFC 7540 8.1.2.3) */
+
+        struct aws_byte_cursor host_value;
```
graebm, line 898: paste in a link to RFC and comment (or summary of it)

include/aws/http/private/http_impl.h
TingDaoK: there already has a TODO for it. I guess I can just do it in this PR

```
/* TODO: handle connection-specific header field (RFC7540 8.1.2.2) */
```

source/h2_stream.c
```diff
@@ -813,6 +827,11 @@ struct aws_h2err aws_h2_stream_on_decoder_headers_end(
             return s_send_rst_and_close_stream(stream, aws_h2err_from_last_error());
         }
     }
+    if (stream->thread_data.check_incoming_content_length) {
+        if (ignore_content_length) {
+            stream->thread_data.check_incoming_content_length = false;
```
graebm, line 832: when you make future changes, make sure we still check for duplicate content-length headers

source/h2_stream.c
```diff
@@ -871,6 +890,21 @@ struct aws_h2err aws_h2_stream_on_decoder_data_begin(
         return s_send_rst_and_close_stream(stream, aws_h2err_from_h2_code(AWS_HTTP2_ERR_PROTOCOL_ERROR));
     }
 
+    if (stream->thread_data.check_incoming_content_length) {
+        stream->thread_data.incoming_data_length += payload_len;
```
graebm, line 894: use the add_u64_checked() function for overflow checking

source/h2_stream.c
```diff
@@ -975,6 +1009,26 @@ struct aws_h2err aws_h2_stream_on_decoder_end_stream(struct aws_h2_stream *strea
...
+         * Clients MUST NOT accept a malformed response.
+         */
+        AWS_H2_STREAM_LOGF(
+            ERROR,
+            stream,
+            "Total received data payload=%" PRIu64 " does not match the received content-length header, which=%" PRIi64
+            ". Closing malformed stream",
+            stream->thread_data.incoming_data_length,
+            stream->thread_data.incoming_content_length);
+        return s_send_rst_and_close_stream(stream, aws_h2err_from_h2_code(AWS_HTTP2_ERR_PROTOCOL_ERROR));
```
graebm, line 1029: double check that this is the right way to handle malformed msg errors.

I forget the nuances, but the decoder needs to keep processing the HEADER frames because HPACK

source/request_response.c
```diff
@@ -948,8 +968,29 @@ struct aws_http_message *aws_http2_message_new_from_http1(
...
+                 * fields as discussed in Section 7.6.1 of [HTTP]. (RFC=9113 8.2.2)
+                 */
+                AWS_LOGF_TRACE(
+                    AWS_LS_HTTP_GENERAL,
+                    "Skip connection-specific headers - \"%.*s\" ",
+                    (int)lower_name_cursor.len,
+                    lower_name_cursor.ptr);
+                aws_byte_buf_reset(&lower_name_buf, false);
+                continue;
+                break;
```
graebm, line 988: remove the `break;` that's weird. continue is weird too here's a comment

```
continue; /* jump to next for-loop iteration */
```

tests/test_h2_client.c
```diff
@@ -564,6 +564,7 @@ TEST_CASE(h2_client_stream_with_h1_request_message) {
         DEFINE_HEADER("Accept", "*/*"),
         DEFINE_HEADER("Host", "example.com"),
         DEFINE_HEADER("Content-Length", "5"),
+        DEFINE_HEADER("Transfer-Encoding", "chunked"), /* Connection-specific header should be skiped */
```
graebm, line 567: trivial: use a different connection-specific header.
It's super illegal in HTTP/1 to have both content-length and transfer-encoding

source/h2_stream.c
TingDaoK: > Malformed requests or responses that are detected MUST be treated as a [stream error](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#StreamErrorHandler) ([Section 5.4.2](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#StreamErrorHandler)) of type [PROTOCOL_ERROR](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#PROTOCOL_ERROR).

We did it right

source/request_response.c
```diff
@@ -959,8 +979,28 @@ struct aws_http_message *aws_http2_message_new_from_http1(
...
+                 * An intermediary transforming an HTTP/1.x message to HTTP/2 MUST remove connection-specific header
+                 * fields as discussed in Section 7.6.1 of [HTTP]. (RFC=9113 8.2.2)
+                 */
+                AWS_LOGF_TRACE(
+                    AWS_LS_HTTP_GENERAL,
+                    "Skip connection-specific headers - \"%.*s\" ",
+                    (int)lower_name_cursor.len,
+                    lower_name_cursor.ptr);
+                aws_byte_buf_reset(&lower_name_buf, false);
+                continue; /* jump to next for-loop iteration */
```
graebm, line 998: this seems fragile: repeating the `aws_byte_buf_reset()` from the end of the for-loop.

how about moving `aws_byte_buf_reset()` to the start of the for-loop instead

or having a `bool copy_header = true;` and then you do this:
```
copy_header = false;
break;
```
instead of the weird continue-from-case

source/h2_stream.c
```diff
@@ -1114,6 +1152,36 @@ struct aws_h2err aws_h2_stream_on_decoder_end_stream(struct aws_h2_stream *strea
      * an actual frame type. It's a flag on DATA or HEADERS frames, and we
      * already checked the legality of those frames in their respective callbacks. */
 
+    if (stream->thread_data.content_length_received) {
+        if (stream->base.request_method != AWS_HTTP_METHOD_HEAD &&
+            stream->base.client_data->response_status != AWS_HTTP_STATUS_CODE_304_NOT_MODIFIED) {
```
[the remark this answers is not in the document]
graebm: Huh. You're right that the HTTP/2 spec isn't required to enforce the body-less-ness of a HEAD or 304 response.

HTTP/1.1 is required to treat HEAD and 304 as having no body, because HTTP/1.1 is just going to treat the response as done after the headers section. It can't exactly validate it, because there's no "framing" just a raw stream of bytes.

I was going to say it seemed like an opportunity for request smuggling. Where HTTP/2 thinks it's 1 message, but if that's forwarded to HTTP/1.1 you could trick it into processing multiple messages...

But this only applies to responses. And I'm not sure *response* smuggling is a security concern? only *request* smuggling?

graebm (approved):
read my comments ... but looks about ready to go

PR #380, opened 2022-06-20
stream manager prior knowledge 

TingDaoK:
- More clear configuration for prior knowledge. More clear validation as well. 

### DIFF
--- a/include/aws/http/http2_stream_manager.h
+++ b/include/aws/http/http2_stream_manager.h
@@ -51,12 +51,12 @@ struct aws_http2_stream_manager_options {
     /**
      * If TLS options is set, you also need to handle ALPN, otherwise, may not able to get HTTP/2 connection and fail
      * the stream manager.
-     * If TLS options not set, you can set prior_knowledge to true to establish HTTP/2 connection. We don't support
-     * upgrade from HTTP/1 connection.
+     * If TLS options not set, you can set prior_knowledge to true to establish HTTP/2 connection. Upgrade from HTTP/1
+     * connection is not supported.
      */
     const struct aws_tls_connection_options *tls_connection_options;
     /* Use prior knowledge or not. Error out when TLS is set and this is true. */
-    const bool prior_knowledge;
+    bool prior_knowledge;
     struct aws_byte_cursor host;
     uint16_t port;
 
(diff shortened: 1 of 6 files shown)


include/aws/http/http2_stream_manager.h
```diff
@@ -51,9 +51,12 @@ struct aws_http2_stream_manager_options {
     /**
      * If TLS options is set, you also need to handle ALPN, otherwise, may not able to get HTTP/2 connection and fail
      * the stream manager.
-     * If TLS options not set, prior knowledge will be used.
+     * If TLS options not set, you can set prior_knowledge to true to establish HTTP/2 connection. We don't support
+     * upgrade from HTTP/1 connection.
```
graebm, line 55: trivial: ~We don't support~
maybe: Upgrade from HTTP/1 connection is not supported.

It's fine to say "you" when talking to users, but don't use "we" in the public documentation (it's fine to use in C files)

include/aws/http/http2_stream_manager.h
```diff
@@ -51,9 +51,12 @@ struct aws_http2_stream_manager_options {
...
      */
     const struct aws_tls_connection_options *tls_connection_options;
+    /* Use prior knowledge or not. Error out when TLS is set and this is true. */
+    const bool prior_knowledge;
```
graebm, line 59: trivial: `const bool` isn't common in our codebase, we generally just use `const` for pointers, but don't bother marking by-value variables with `const`

tests/test_stream_manager.c
```diff
@@ -48,6 +48,7 @@ struct sm_tester_options {
 
     struct aws_byte_cursor *uri_cursor;
     const enum aws_log_level *log_level;
+    const bool prior_knowledge;
```
graebm, line 51: trivial: `const bool` is weird

source/connection_manager.c
```diff
@@ -812,6 +812,13 @@ struct aws_http_connection_manager *aws_http_connection_manager_new(
         return NULL;
     }
 
+    if (options->tls_connection_options && options->prior_knowledge_http2) {
+        AWS_LOGF_ERROR(
+            AWS_LS_HTTP_CONNECTION_MANAGER, "Invalid options - prior knowledge cannot be set when TLS is used");
```
graebm, line 817: Is it an error if neither `tls_connection_options` or `prior_knowledge_http2` is set? because we don't support Upgrade from HTTP/1?

graebm: related:
should we support prior-knowledge with TLS? Like, if TLS is used but not ALPN? Is that even a thing? We should only bother trying to support it if we have an actual internal use-case

graebm (approved):

PR #381, opened 2022-06-22
support close on server error and period ping for stream manager

TingDaoK:
- support close connection on server error for stream manager
- support period ping for stream manager
- Fixed a bug that stream manager will hang forever

### DIFF
--- a/source/http2_stream_manager.c
+++ b/source/http2_stream_manager.c
@@ -33,7 +33,7 @@
 #define STREAM_MANAGER_LOG(level, stream_manager, text) STREAM_MANAGER_LOGF(level, stream_manager, "%s", text)
 
 /* 3 seconds */
-static const uint64_t s_default_ping_timeout_ms = 3000;
+static const size_t s_default_ping_timeout_ms = 3000;
 
 static void s_stream_manager_start_destroy(struct aws_http2_stream_manager *stream_manager);
 static void s_aws_http2_stream_manager_build_transaction_synced(struct aws_http2_stream_management_transaction *work);


include/aws/http/http2_stream_manager.h
```diff
@@ -97,6 +97,26 @@ struct aws_http2_stream_manager_options {
...
+     * Note: if set, it must be large than the time of ping timeout setting.
+     */
+    size_t connection_ping_period_sec;
+    /**
+     * Optional.
+     * Network connection will be closed if a ping response is not received
+     * within this amount of time (nanoseconds).
+     * If you specify 0, a default value of 3 seconds is used.
+     */
+    size_t connection_ping_timeout_ns;
```
graebm, line 118: using sec for one and ns for the other feels weird to me. How about we just make both of them ms? seems like a happy medium

include/aws/http/http2_stream_manager.h
```diff
@@ -97,6 +97,26 @@ struct aws_http2_stream_manager_options {
...
+     * The period for all the connections held by stream manager to send a PING in seconds.
+     * If you specify 0, manager will NOT send any PING.
+     * Note: if set, it must be large than the time of ping timeout setting.
+     */
+    size_t connection_ping_period_sec;
+    /**
+     * Optional.
+     * Network connection will be closed if a ping response is not received
+     * within this amount of time (nanoseconds).
+     * If you specify 0, a default value of 3 seconds is used.
```
graebm, line 116: trivial: let's just say "a default value is used" instead of documenting 3.

If we document 3, that documentation is going to get copy/pasted into the CRTs, then someday we'll change the defaults and the CRT documentation will be wrong

include/aws/http/private/http2_stream_manager_impl.h
```diff
@@ -25,12 +25,19 @@ enum aws_h2_sm_connection_state_type {
...
  * connection, this structure should die */
 struct aws_h2_sm_connection {
+    struct aws_allocator *allocator;
     struct aws_http2_stream_manager *stream_manager;
     struct aws_http_connection *connection;
     uint32_t num_streams_assigned;   /* From a stream assigned to the connection until the stream completed
                                                      or failed to be created from the connection. */
     uint32_t max_concurrent_streams; /* lower bound between user configured and the other side */
 
+    /* task to send ping periodically from connection thread. */
```
graebm, line 35: Can we avoid adding this complexity by using TLS keepalive instead?

I know the Java SDK used HTTP PING to stop their connections from closing due to inactivity, but maybe there's another way? Can we find out which service was timing them out, try to reproduce the issue with our stream manager, and then see if TLS keepalive solves the issue?

less code is best code

source/http2_stream_manager.c
```diff
@@ -636,6 +776,22 @@ static void s_on_stream_complete(struct aws_http_stream *stream, int error_code,
...
+        AWS_ASSERT(status_code != 0); /* The get status should not fail */
+        if (status_code / 100 == 5) {
+            STREAM_MANAGER_LOGF(
+                DEBUG,
+                stream_manager,
+                "%d response received for stream: %p. Closing connection: %p",
+                status_code,
+                (void *)stream,
+                (void *)sm_connection->connection);
+            aws_http_connection_close(sm_connection->connection);
```
graebm, line 792: I see a few problems with this approach:
1) major: There might be other streams on this connection, and they will get suddenly murdered due to the connection shutting down.
2) minor: We could have stopped putting new streams on this connection earlier (when the headers were first received) instead of waiting until the response was complete

Can we use `aws_http2_send_goaway()` instead? I think (???) that would allow us to finish any existing streams, while preventing any new streams from getting put on this connection. Also, we might be able to call aws_http2_send_goaway() earlier, from the on_headers callback, instead of waiting until on_complete

include/aws/http/private/http2_stream_manager_impl.h
TingDaoK: hmmm, will check with Java team. But, did a quick search.

Downside of TCP keepalive from [here](https://stackoverflow.com/questions/23238319/websockets-ping-pong-why-not-tcp-keepalive):
- It operates between proxies rather than end to end.
- As pointed out by @DavidSchwartz, it operates between TCP stacks, not between the applications so therefore it doesn’t tell us whether the application is alive or not.

include/aws/http/http2_stream_manager.h
```diff
@@ -105,6 +105,26 @@ struct aws_http2_stream_manager_options {
...
+     * When set, connection will be closed if 5xx response received from server.
+     */
+    bool close_connection_on_server_error;
+    /**
+     * Optional.
+     * The period for all the connections held by stream manager to send a PING in seconds.
+     * If you specify 0, manager will NOT send any PING.
+     * Note: if set, it must be large than the time of ping timeout setting.
+     */
+    size_t connection_ping_period_sec;
```
graebm, line 119: I can't decide if this logic should live in the connection or not :|
It would be a bit cleaner there, since the h2 stream manager could give a connection back to the connection-mgr, then acquire it again and kick off more pings.

but then I don't want the h2_connection.c code any more complex than it already is so ... 🤷‍♀️

TingDaoK: I think it's okay to stay this level. I can see it may lead to a confusion that the ping for same connection was sent with the same time interval in the future. 

But, the idea for this is as the stream manager user, you don't have access to the connection under the hood. But if people can access connection, they can implement their own logic.

And we also said `The period for all the connections held by stream manager` in the documentation. So, user knows the time interval will only work if the connection is held by the stream manager all the time.

source/http2_stream_manager.c
```diff
@@ -390,11 +394,125 @@ static void s_aws_http2_stream_manager_build_transaction_synced(struct aws_http2
...
+    aws_channel_current_clock_time(channel, &schedule_time);
+    schedule_time += sm_connection->stream_manager->connection_ping_timeout_ns;
+    aws_channel_task_init(
+        &sm_connection->ping_timeout_task,
+        s_connection_ping_timeout_task,
+        sm_connection,
+        "Stream manager connection ping timeout task");
+    aws_channel_schedule_task_future(channel, &sm_connection->ping_timeout_task, schedule_time);
+    /* acquire a refcount for timeout task to run */
+    aws_ref_count_acquire(&sm_connection->ref_count);
```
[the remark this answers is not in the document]
TingDaoK: Did a quick search, don't see anywhere else other than the new code has the issue,

source/http2_stream_manager.c
TingDaoK: We don't prevent new streams if we send goaway. The spec also doesn't mentioned the sender behavior after sending goaway beside not receiving any new stream from other side.

I am not sure about should we enforce people to be like after sending goaway, no more new streams can be made... It seems making sense though... Will do that...

include/aws/http/http2_stream_manager.h
TingDaoK: Hmmm, I assume user will want to have small timeout time setting and a much longer ping period setting. 

For MQTT, we also have `keep_alive_time_secs` and `ping_timeout_ms`. Maybe it's better to have `connection_ping_timeout_ms` instead though, no one wants to use ns I guess.

source/http2_stream_manager.c
graebm: If you want the ping stuff to be merged before the server-error stuff, you might want to split this into 2 different PRs 🤷‍♀️

include/aws/http/http2_stream_manager.h
```diff
@@ -105,6 +105,26 @@ struct aws_http2_stream_manager_options {
     void *shutdown_complete_user_data;
     aws_http2_stream_manager_shutdown_complete_fn *shutdown_complete_callback;
 
+    /**
+     * Optional.
+     * When set, connection will be closed if 5xx response received from server.
+     */
+    bool close_connection_on_server_error;
```
graebm, line 112: Is there anything else that might make the connection undesirable? If we had the API for marking a connection undesirable, then the S3 client could have its own specific logic, instead of baking the logic in at this low level

include/aws/http/http2_stream_manager.h
graebm: I still hate that these use two different units of time
Please use `ms` for both
MQTT uses `sec` because that's required by the spec

source/http2_stream_manager.c
```diff
@@ -390,11 +394,119 @@ static void s_aws_http2_stream_manager_build_transaction_synced(struct aws_http2
...
+    STREAM_MANAGER_LOGF(
+        TRACE, sm_connection->stream_manager, "Sending PING for connection: %p.", (void *)sm_connection->connection);
+    aws_http2_connection_ping(sm_connection->connection, NULL, s_on_ping_complete, sm_connection);
+    sm_connection->thread_data.ping_received = false;
+    struct aws_channel *channel = aws_http_connection_get_channel(sm_connection->connection);
+    uint64_t schedule_time = 0;
+    aws_channel_current_clock_time(channel, &schedule_time);
+    sm_connection->thread_data.next_ping_task_time =
+        schedule_time + sm_connection->stream_manager->connection_ping_period_ns;
+    schedule_time += sm_connection->stream_manager->connection_ping_timeout_ns;
```
graebm, line 489: trivial: the `schedule_time` variable is a bit confusing ... where it's used as "current time" on line 488 but then we update it to be "timeout time" on line 489.

Can we just have two different variables: rename ~schedule_times~ -> `current_time` and then have a new variable later like `uint64_t timeout_time = current_time + sm_connection->stream_manager->connection_ping_timeout_ns`

source/http2_stream_manager.c
```diff
@@ -390,11 +394,119 @@ static void s_aws_http2_stream_manager_build_transaction_synced(struct aws_http2
...
+    if (round_trip_time_ns > sm_connection->stream_manager->connection_ping_timeout_ns) {
+        /* Timeout happened. It may be caught by the timeout task first, but still close the connection. */
+        STREAM_MANAGER_LOGF(
+            ERROR,
+            sm_connection->stream_manager,
+            "ping timeout detected for connection: %p from ping_complete, round trip time in ns is: %" PRIu64
+            ". Closing connection.",
+            (void *)sm_connection->connection,
+            round_trip_time_ns);
+        aws_http_connection_close(sm_connection->connection);
```
graebm, line 421: debatable: why bother closing here? we're already doing this from the timeout task? seems like an unnecessary bit of complexity to do the same thing from two places. If we fix a bug in one place we will probably not remember to fix it in the other

source/http2_stream_manager.c
```diff
@@ -390,11 +394,119 @@ static void s_aws_http2_stream_manager_build_transaction_synced(struct aws_http2
...
+static void s_on_ping_complete(
+    struct aws_http_connection *http2_connection,
+    uint64_t round_trip_time_ns,
+    int error_code,
+    void *user_data) {
+
+    (void)http2_connection;
+    struct aws_h2_sm_connection *sm_connection = user_data;
+    if (error_code) {
+        return;
```
graebm, line 406: don't we need to do ref-counting on `sm_connection` to be sure it stays alive until the ping_complete is received?

TingDaoK: yeah, you are right.

source/http2_stream_manager.c
TingDaoK: yeah, will remove this. Doesn't do anything actually

source/http2_stream_manager.c
TingDaoK: added `stop_new_request` api

include/aws/http/http2_stream_manager.h
TingDaoK: The stream manager user doesn't have control of the stream, so that I don't think high level user will be able to do it by themselves.

graebm: we'll cross that bridge if we come to it

include/aws/http/http.h
```diff
@@ -54,6 +54,7 @@ enum aws_http_errors {
     AWS_ERROR_HTTP_STREAM_MANAGER_SHUTTING_DOWN,
     AWS_ERROR_HTTP_STREAM_MANAGER_CONNECTION_ACQUIRE_FAILURE,
     AWS_ERROR_HTTP_STREAM_MANAGER_UNEXPECTED_HTTP_VERSION,
+    AWS_ERROR_HTTP_CONNECTION_NEW_REQUEST_STOPPED,
```
graebm, line 57: trivial: ~AWS_ERROR_HTTP_CONNECTION_NEW_REQUEST_STOPPED~ -> AWS_ERROR_HTTP_CONNECTION_NEW_REQUESTS_STOPPED

include/aws/http/private/connection_impl.h
```diff
@@ -40,6 +40,7 @@ struct aws_http_connection_vtable {
         const struct aws_http_request_handler_options *options);
     int (*stream_send_response)(struct aws_http_stream *stream, struct aws_http_message *response);
     void (*close)(struct aws_http_connection *connection);
+    void (*stop_new_request)(struct aws_http_connection *connection);
```
graebm, line 43: request**s**

source/h1_connection.c
```diff
@@ -196,6 +198,16 @@ static void s_connection_close(struct aws_http_connection *connection_base) {
     s_stop(connection, false /*stop_reading*/, false /*stop_writing*/, true /*schedule_shutdown*/, AWS_ERROR_SUCCESS);
 }
 
+static void s_connection_stop_new_request(struct aws_http_connection *connection_base) {
+    struct aws_h1_connection *connection = AWS_CONTAINER_OF(connection_base, struct aws_h1_connection, base);
+
+    { /* BEGIN CRITICAL SECTION */
+        aws_h1_connection_lock_synced_data(connection);
+        connection->synced_data.new_stream_error_code = AWS_ERROR_HTTP_CONNECTION_NEW_REQUEST_STOPPED;
```
graebm, line 206: don't change it if it's already non-zero

source/h2_connection.c
```diff
@@ -2128,6 +2129,16 @@ static void s_connection_close(struct aws_http_connection *connection_base) {
     s_stop(connection, false /*stop_reading*/, false /*stop_writing*/, true /*schedule_shutdown*/, AWS_ERROR_SUCCESS);
 }
 
+static void s_connection_stop_new_request(struct aws_http_connection *connection_base) {
+    struct aws_h2_connection *connection = AWS_CONTAINER_OF(connection_base, struct aws_h2_connection, base);
+
+    { /* BEGIN CRITICAL SECTION */
+        s_lock_synced_data(connection);
+        connection->synced_data.new_stream_error_code = AWS_ERROR_HTTP_CONNECTION_NEW_REQUEST_STOPPED;
```
graebm, line 2137: don't change if it's already nonzero

source/http2_stream_manager.c
```diff
@@ -653,17 +650,21 @@ static int s_on_incoming_headers(
         aws_http_stream_get_incoming_response_status(stream, &status_code);
         AWS_ASSERT(status_code != 0); /* The get status should not fail */
         if (status_code / 100 == 5) {
```
graebm, line 652: check against those 4 specific 5xx codes we looked up instead of ANY 5xx

source/http2_stream_manager.c
```diff
@@ -653,17 +650,21 @@ static int s_on_incoming_headers(
...
+                    DEBUG,
+                    stream_manager,
+                    "%d response received for stream: %p. Closing connection: %p",
+                    status_code,
+                    (void *)stream,
+                    (void *)sm_connection->connection);
+                struct aws_byte_cursor debug_data =
+                    aws_byte_cursor_from_c_str("Close connection for 5xx status received.");
+
                 aws_http2_connection_send_goaway(
```
graebm, line 665: let's not even bother sending GOAWAY anymore, since it doesn't really do anything

which means we should probably change ~goaway_scheduled~ to something like `stopped_new_requests`

source/http2_stream_manager.c
```diff
@@ -653,17 +650,21 @@ static int s_on_incoming_headers(
...
+                /* Sending goaway and stops new requests to be made on the connection. */
+                STREAM_MANAGER_LOGF(
+                    DEBUG,
+                    stream_manager,
+                    "%d response received for stream: %p. Closing connection: %p",
+                    status_code,
+                    (void *)stream,
+                    (void *)sm_connection->connection);
+                struct aws_byte_cursor debug_data =
+                    aws_byte_cursor_from_c_str("Close connection for 5xx status received.");
```
graebm, line 663: print actual status number.
~Close connection~ "no longer using connection due to receiving %d server error status code"

source/http2_stream_manager.c
```diff
@@ -919,6 +1081,26 @@ struct aws_http2_stream_manager *aws_http2_stream_manager_new(
...
 
+    stream_manager->connection_ping_timeout_ns =
+        options->connection_ping_timeout_ms
+            ? aws_timestamp_convert(
+                  options->connection_ping_timeout_ms, AWS_TIMESTAMP_MILLIS, AWS_TIMESTAMP_NANOS, NULL)
+            : s_default_ping_timeout_ns;
+    if (options->connection_ping_period_ms) {
+        stream_manager->connection_ping_period_ns =
+            aws_timestamp_convert(options->connection_ping_period_ms, AWS_TIMESTAMP_MILLIS, AWS_TIMESTAMP_NANOS, NULL);
+        if (stream_manager->connection_ping_period_ns <= stream_manager->connection_ping_timeout_ns) {
```
graebm, line 1092: let's just clamp the timeout so its <= the period

graebm (approved):
fix & ship

PR #383, opened 2022-07-14
Address TODOs

TingDaoK:
- Address several TODOs in HTTP/2 code.
- I think: it fixes the occasionally failure from osx local host built by add the connection monitor time to 2 secs(?) 
- Use assert_equal instead of assert_true, which will tell us what the number is on failure.


source/h2_connection.c
```diff
@@ -2112,8 +2109,22 @@ static struct aws_http_stream *s_connection_make_request(
...
     }
+    int error = 0;
+    struct aws_byte_cursor method;
+    AWS_ZERO_STRUCT(method);
+    struct aws_byte_cursor path;
+    AWS_ZERO_STRUCT(path);
+
+    error |= aws_http_message_get_request_method(stream->thread_data.outgoing_message, &method);
+    error |= aws_http_message_get_request_path(stream->thread_data.outgoing_message, &path);
+    AWS_ASSERT(!error);
```
graebm, line 2120: shouldn't we be checking these errors for real?

source/h2_connection.c
```diff
@@ -2842,6 +2856,9 @@ static void s_gather_statistics(struct aws_channel_handler *handler, struct aws_
 
         connection->thread_data.incoming_timestamp_ns = now_ns;
     } else {
```
graebm, line 2858: trivial: a few lines up, put a newline between the two if-statements in a row
```
if (!aws_linked_list_empty(&connection->thread_data.outgoing_streams_list)) {
    ...
}
// PUT NEWLINE HERE
if (aws_hash_table_get_entry_count(&connection->thread_data.active_streams_map) != 0) {
```

source/h2_decoder.c
```diff
@@ -1159,6 +1159,42 @@ static struct aws_h2err s_flush_pseudoheaders(struct aws_h2_decoder *decoder) {
...
+        if (!aws_strutil_is_http_token(aws_byte_cursor_from_string(method_string))) {
+            DECODER_LOG(ERROR, decoder, "Request method is invalid.");
+            DECODER_LOGF(
+                DEBUG,
+                decoder,
+                "Bad method is: '" PRInSTR "'",
+                AWS_BYTE_CURSOR_PRI(aws_byte_cursor_from_string(method_string)));
+            goto malformed;
+        }
+        if (aws_string_eq_c_str(method_string, "CONNECT")) {
```
graebm, line 1180: trivial: put newlines between if-statements when you have multiple in a row:
```
if () {
}
if () {
} else {
}
```
otherwise it kinda looks like a long series of if-else-if-else, or it may look like the else pairs with the 1st if

graebm, line 1180: trivial: ~"CONNECT"~ -> `aws_http_method_connect`

source/h2_decoder.c
```diff
@@ -1159,6 +1159,42 @@ static struct aws_h2err s_flush_pseudoheaders(struct aws_h2_decoder *decoder) {
...
+            if (scheme_string || path_string) {
+                /* RFC-9113 8.5 The ":scheme" and ":path" pseudo-header fields MUST be omitted for CONNECT method */
+                DECODER_LOG(ERROR, decoder, "CONNECT request has :path or :scheme that are must omitted.");
```
graebm, line 1183: do we need to check that these things aren't empty strings?

I see you checking the request_string, but not the others

source/h2_decoder.c
```diff
@@ -1529,12 +1586,6 @@ static struct aws_h2err s_state_fn_header_block_entry(struct aws_h2_decoder *dec
 
     /* Finished decoding HPACK entry! */
 
-    /* #TODO Enforces dynamic table resize rules from RFC-7541 4.2
```
graebm, line 1532: this is how far I got in reviewing your code...

source/h2_frames.c
```diff
@@ -1228,7 +1226,8 @@ int aws_h2_encode_frame(
 
 void aws_h2_frame_encoder_set_setting_header_table_size(struct aws_h2_frame_encoder *encoder, uint32_t data) {
     /* Setting for dynamic table size changed from peer, we will update the dynamic table size when we encoder the next
-     * header block */
+     * header block. The spec requires us to update the table if the updated size is smaller than the current size, but
+     * we can still update it even if not */
     aws_hpack_set_max_table_size(encoder->hpack, data);
     aws_hpack_set_protocol_max_size_setting(encoder->hpack, data);
```
graebm, line 1232: I just spent like an hour trying to make sense of whether or not we're doing the right thing with this hpack resize stuff. MAN this stuff is hard to understand. We just have a lot of variables with similar names, and the 1 hpack class that has both encoding and decoding logic. Can I make some suggestions?

1) Don't have `aws_h2_frame_encoder_set_setting_header_table_size()` call `aws_hpack_set_protocol_max_size_setting()` at all. Turns out that function has no effect on an aws_hpack_context that is being used for encoding.

2) rename ~aws_hpack_set_max_table_size()~ ->
```
/* Call this after receiving SETTINGS_HEADER_TABLE_SIZE from peer and sending the ACK.
 * The hpack-encoder remembers all size updates, and makes sure to encode the proper
 * number of Dynamic Table Size Updates the next time a header block is sent. */
AWS_HTTP_API
void aws_hpack_encoder_update_max_table_size(struct aws_hpack_context *context, uint32_t new_max_size);
```

3) rename ~aws_hpack_set_protocol_max_size_setting()~ ->
```
/* Call this after sending SETTINGS_HEADER_TABLE_SIZE and receiving ACK from the peer.
 * The hpack-decoder remembers all size updates, and makes sure that the peer
 * sends the appropriate Dynamic Table Size Updates in the next header block we receive. */
AWS_HTTP_API
void aws_hpack_decoder_update_max_table_size(struct aws_hpack_context *context, uint32_t new_max_size);
```

4) I still don't think `aws_hpack_set_protocol_max_size_setting()` aka `aws_hpack_decoder_update_max_table_size()` is doing the right thing. In the same way the encoder tracks a `smallest` and `last` size, I think we need to do similar tracking for the decoder so we can properly check that the peer sends the right number of updates?

man, this is why we need to split ~aws_hpack_context~ into `aws_hpack_encoder` and `aws_hpack_decoder`. It's too confusing having everyone's variables and functions all mixed up together

source/hpack_decoder.c
```diff
@@ -278,6 +278,25 @@ int aws_hpack_decode(
                     decoder->progress_entry.u.literal.prefix_size = 4;
                     decoder->progress_entry.state = HPACK_ENTRY_STATE_LITERAL_BEGIN;
                 }
+                if (decoder->dynamic_table_protocol_max_size_setting < decoder->context.dynamic_table.size) {
```
graebm, line 281: trivial: 
```
if
else if
else if
else
if  <-- newline before this if-block please
```

graebm, line 281: I don't think the logic here is correct

The encoder has the correct logic. In the `dynamic_table_size_update` struct it tracks `latest_value` and `smallest_value` separately.
```
    struct {
        size_t latest_value;
        size_t smallest_value;
        bool pending;
    } dynamic_table_size_update;
```

The decoder needs to do something similar, if multiple SETTINGS packets are received back to back, and change the table size to 10, then to 100, then the decoder needs to confirm that it receives an update `<= 10` (RFC-7541 4.2)

include/aws/http/http.h
```diff
@@ -54,6 +54,7 @@ enum aws_http_errors {
     AWS_ERROR_HTTP_STREAM_MANAGER_SHUTTING_DOWN,
     AWS_ERROR_HTTP_STREAM_MANAGER_CONNECTION_ACQUIRE_FAILURE,
     AWS_ERROR_HTTP_STREAM_MANAGER_UNEXPECTED_HTTP_VERSION,
+    AWS_ERROR_HTTP_REQUIRED_PSEUDO_HEADER_MISSING,
```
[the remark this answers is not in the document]
TingDaoK: it was used [here](https://github.com/awslabs/aws-c-http/pull/383/files#diff-259f2d1cff2609f6e30c12235e3daf0d7a002ee0705c0184c9b7739d9195ef92R2107)

source/h2_decoder.c
TingDaoK: It was checked before. from the `s_process_header_field` at [here](https://github.com/awslabs/aws-c-http/pull/383/files#diff-ade185b92b6e3655280095a0215637f556a63fd1e236f60a46663b8e96605046R1291)

source/h2_decoder.c
```diff
@@ -1253,6 +1288,21 @@ static struct aws_h2err s_process_header_field(
         goto malformed;
     }
 
+    if (!aws_strutil_is_http_field_value(header_field->value)) {
```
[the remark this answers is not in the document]
TingDaoK: > When a specific version of HTTP allows such whitespace to appear in a message
A field value MUST NOT start or end with an ASCII whitespace character (ASCII SP or HTAB, 0x20 or 0x09). [here](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#section-8.2.1-3.4). So, I think we should not trim the white space for h2

source/hpack_decoder.c
TingDaoK: Hmmm, I read again, and still feel confusing about:
> An endpoint MUST treat a field block that follows an acknowledgment of the reduction to the maximum dynamic table size as a [connection error](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#ConnectionErrorHandler) ([Section 5.4.1](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#ConnectionErrorHandler)) of type [COMPRESSION_ERROR](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#COMPRESSION_ERROR) if it does not start with a conformant Dynamic Table Size Update instruction.

[here](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#section-4.3.1-4) should be the requirement from the decoder. Especially about `start with a conformant Dynamic Table Size Update instruction`. *What is conformant update instruction for the case you pointing out to start with?* I don't think there is any clear statement about what kind of `Dynamic Table Size Update instruction` to start with.

from hpack spec [here](https://www.rfc-editor.org/rfc/rfc7541#section-4.2):
>the smallest maximum table size that occurs in that interval MUST be signaled in a dynamic table
   size update.  The final maximum size is always signaled, resulting in at most two dynamic table size updates

I guess the first one could be larger than the smallest value as long as the second one is smaller than it. :(

TingDaoK: > resulting in **at most two** dynamic table size updates

However, I don't see any place that enforce the dynamic table size updates to be no more than one in normal case. But I guess they assume that's the case as from [here](https://www.rfc-editor.org/rfc/rfc7541#section-6.3)
>In HTTP/2, this limit is **the last value** of the SETTINGS_HEADER_TABLE_SIZE parameter

Here is my understanding:

- We can receive at most two continuous Dynamic Table Size Update instruction after ack the settings.
- One of those two MUST be smaller than the smallest max.
- Everyone MUST be smaller than the latest max.

PR #385, opened 2022-07-28
Split up hpack.c

graebm:
**Issue**:
`aws_hpack_context` did too many things and was difficult to understand.

**Description of changes:**
Sorry, this looks like a lot of new code, but mostly just moving stuff around. Changes are:
- Split `aws_hpack_context` (hpack.c) into 3 classes:
    - `aws_hpack_context` (hpack.c) - Still handles dynamic and static table logic.
    - `aws_hpack_encoder` (hpack_encoder.c) - Handles encoding of outgoing headers (contains an `aws_hpack_context`)
    - `aws_hpack_decoder` (hpack_encoder.c) - Handles decoding of incoming headers (contains an `aws_hpack_context`)
- These structs are now `init()/clean_up()` by-value structs, they used to be `new()/delete()` heap-only structs.
    - This avoids an unnecessary allocation (total of 4 per HTTP connection)
    - These are package-private structs so don't really care that their internals are visible.

### DIFF
--- a/source/hpack.c
+++ b/source/hpack.c
@@ -4,120 +4,18 @@
  */
 #include <aws/http/private/hpack.h>
 
-#include <aws/http/request_response.h>
-
-#include <aws/compression/huffman.h>
-
-#include <aws/common/byte_buf.h>
-#include <aws/common/hash_table.h>
-#include <aws/common/logging.h>
-#include <aws/common/string.h>
-
-#include <inttypes.h>
-
-/* #TODO split hpack encoder/decoder into different types */
-
 /* #TODO test empty strings */
+
+/* #TODO remove all OOM error handling in HTTP/2 & HPACK. make functions void if possible */
 
 /* RFC-7540 6.5.2 */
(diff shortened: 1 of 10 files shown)


source/hpack.c
```diff
@@ -479,13 +222,15 @@ size_t aws_hpack_find_index(
         /* Check name-and-value first in static table */
         aws_hash_table_find(&s_static_header_reverse_lookup, header, &elem);
         if (elem) {
+            /* TODO: Maybe always set found_value to true? Who cares that the value is empty if they matched? */
```
TingDaoK, line 225: I read through the spec again, doesn't seem like it matters...

I'll vote for always set found_value to true this case. It seems less steps to handle for both side

TingDaoK (approved):
🎉 finally

source/hpack.c
graebm: I'm leaving this TODO so that this commit can be as minimal as possible: just moving code around, no functional changes

PR #388, opened 2022-08-26
Update h2 windowing algo & Http Client benchmark

TingDaoK:
- Initial build of our http client benchmark
- Test using a local host, that using our http client to connect to the host and collect how many requests are made during a certain time.
- To run: 
  - start the python local host: https://github.com/awslabs/aws-c-http/tree/main/tests/py_localhost
  - build aws-c-http with `AWS_BUILD_CANARY=ON` cmake flag
  - `cd build/bin/canary && ./canary`

### ISSUE FOUND & FIXED
- We update the window for connection for each data frame received, which really slows us down for frequent small chunk of data frames receiving.
   - We fixed it by only update the connection window whenever it drops to 50% of the max.
   - Same issue may happens to streams windows
   - Or the padding of the connection window.

- Providing tiny increments to flow control in [WINDOW_UPDATE](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#WINDOW_UPDATE) frames can cause a sender to generate a large number of [DATA](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#DATA) frames. from [here](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#section-10.5-4.1.2.1)
  - Is the client's responsibility to make sure not doing this? Even if the user do manual window update and doing small window update?

- https://github.com/awslabs/aws-c-http/pull/514


bin/canary/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+project(canary C)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_INSTALL_PREFIX}/lib/cmake")
+
+file(GLOB ELASTICURL_SRC
```
DmitriyMusatkin, line 5: nit: canary_src?

bin/canary/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+
+
+target_include_directories(${PROJECT_NAME} PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+        $<INSTALL_INTERFACE:include>)
+
+target_link_libraries(${PROJECT_NAME} aws-c-http)
+
+if (BUILD_SHARED_LIBS AND NOT WIN32)
+    message(INFO " elasticurl will be built with shared libs, but you may need to set LD_LIBRARY_PATH=${CMAKE_INSTALL_PREFIX}/lib to run the application")
```
DmitriyMusatkin, line 21: nit: canary instead of elasticurl

bin/canary/main.c
```diff
@@ -0,0 +1,527 @@
...
+#include <aws/io/logging.h>
+#include <aws/io/socket.h>
+#include <aws/io/stream.h>
+#include <aws/io/tls_channel_handler.h>
+#include <aws/io/uri.h>
+
+#include <inttypes.h>
+
+#ifdef _MSC_VER
+#    pragma warning(disable : 4996) /* Disable warnings about fopen() being insecure */
```
DmitriyMusatkin, line 30: nit: do you need this? there is no fopen in file

source/h2_connection.c
```diff
@@ -1762,6 +1767,8 @@ static void s_handler_installed(struct aws_channel_handler *handler, struct aws_
         aws_linked_list_push_back(
             &connection->thread_data.outgoing_frames_queue, &connection_window_update_frame->node);
         connection->thread_data.window_size_self += initial_window_update_size;
+        /* For automatic window management, we only update connectio windows when it droped blow 50% of MAX. */
+        connection->thread_data.window_size_self_dropped_threshold = AWS_H2_WINDOW_UPDATE_MAX / 2;
```
DmitriyMusatkin, line 1771: nit: pull this magic number into a constant?

TingDaoK: it's derive from a constant... and it's more clear about where it comes from/

tests/test_localhost_integ.c
```diff
@@ -408,12 +408,6 @@ static int s_localhost_integ_h2_upload_stress(struct aws_allocator *allocator, v
     s_tester.alloc = allocator;
 
     size_t length = 2500000000UL;
-#ifdef AWS_OS_LINUX
```
TingDaoK, line 411: remove this, since it seems to be the flow control window issue.
1. we sent out too much
2. the initial window from the local server is small (default to the magic 65535). https://httpwg.org/specs/rfc7540.html#iana-settings
3. The local server code also update the window as data received.

source/h2_stream.c
```diff
@@ -747,6 +747,10 @@ int aws_h2_stream_on_activated(struct aws_h2_stream *stream, enum aws_h2_stream_
     stream->thread_data.window_size_self =
         connection->thread_data.settings_self[AWS_HTTP2_SETTINGS_INITIAL_WINDOW_SIZE];
 
+    if (!connection->base.stream_manual_window_management) {
+        stream->thread_data.window_size_self_dropped_threshold =
+            connection->thread_data.settings_self[AWS_HTTP2_SETTINGS_INITIAL_WINDOW_SIZE] / 2;
+    }
     if (with_data) {
```
graebm, line 754: trivial: whitespace after logical if-block please (or if/else block, or if/elseif/elseif block)
```suggestion
    }
    
    if (with_data) {
```

source/h2_stream.c
```diff
@@ -747,6 +747,10 @@ int aws_h2_stream_on_activated(struct aws_h2_stream *stream, enum aws_h2_stream_
     stream->thread_data.window_size_self =
         connection->thread_data.settings_self[AWS_HTTP2_SETTINGS_INITIAL_WINDOW_SIZE];
 
+    if (!connection->base.stream_manual_window_management) {
+        stream->thread_data.window_size_self_dropped_threshold =
+            connection->thread_data.settings_self[AWS_HTTP2_SETTINGS_INITIAL_WINDOW_SIZE] / 2;
```
graebm, line 752: similar feedback to connection window batching:
- we should probably still batch up WINDOW_UPDATE frames, even if the user is doing it "manual", because most users will do 1 update for each read(), just like we were doing in our "automatic" behavior, which was giving performance problems
- we probably want to send the updates anyway if the total window size is too small

maybe just hardcode these 2 magic numbers: the preferred batch size, and the window size at which we don't bother batching, and use the same values for both stream and connection behavior, whether its on automatic or manual mode

source/h2_connection.c
```diff
@@ -1186,19 +1197,99 @@ struct aws_h2err s_decoder_on_push_promise(uint32_t stream_id, uint32_t promised
...
+    if (window_delta > 0) {
+        struct aws_h2_frame *connection_window_update_frame =
+            aws_h2_frame_new_window_update(connection->base.alloc, 0, window_delta);
+        if (!connection_window_update_frame) {
+            CONNECTION_LOGF(
+                ERROR,
+                connection,
+                "WINDOW_UPDATE frame on connection failed to be sent, error %s",
+                aws_error_name(aws_last_error()));
+            connection->thread_data.window_size_self = previous_window;
```
graebm, line 1283: trivial: maybe this was necessary in some earlier version of your code, but it's not necessary now
```suggestion
```

source/h2_connection.c
```diff
@@ -1186,19 +1197,99 @@ struct aws_h2err s_decoder_on_push_promise(uint32_t stream_id, uint32_t promised
...
+            connection->thread_data.window_size_self,
+            connection->window_size_threshold_to_send_update,
+            connection->thread_data.pending_window_update_size_thread);
+        return AWS_OP_SUCCESS;
     }
-    aws_h2_connection_enqueue_outgoing_frame(connection, connection_window_update_frame);
-    connection->thread_data.window_size_self += window_size;
+
+    /* Cap the window to AWS_H2_WINDOW_UPDATE_MAX */
+    uint32_t previous_window = connection->thread_data.window_size_self;
```
graebm, line 1261: trivial: likewise
```suggestion
```

source/h2_connection.c
```diff
@@ -1186,19 +1197,99 @@ struct aws_h2err s_decoder_on_push_promise(uint32_t stream_id, uint32_t promised
...
+                connection,
+                "WINDOW_UPDATE frame on connection failed to be sent, error %s",
+                aws_error_name(aws_last_error()));
+            connection->thread_data.window_size_self = previous_window;
+            return AWS_OP_ERR;
+        }
+        CONNECTION_LOGF(DEBUG, connection, "Sending connection window by %" PRIu32 ".", window_delta);
+        aws_h2_connection_enqueue_outgoing_frame(connection, connection_window_update_frame);
+        /* The math in aws_h2_calculate_cap_window_update_delta makes sure no overflow afterwards. */
+        connection->thread_data.window_size_self = previous_window + window_delta;
```
graebm, line 1289: trivial: likewise
```suggestion
        connection->thread_data.window_size_self += window_delta;
```

source/h2_connection.c
```diff
@@ -1247,16 +1338,8 @@ struct aws_h2err s_decoder_on_data_begin(
...
-            return aws_h2err_from_last_error();
-        }
-        CONNECTION_LOGF(
-            TRACE,
-            connection,
-            "Automatically updating connection window by %" PRIu32 "(%" PRIu32 " due to padding).",
-            auto_window_update,
-            total_padding_bytes);
+    if (s_connection_send_update_window_if_needed(connection, auto_window_update)) {
+        return aws_h2err_from_last_error();
```
graebm, line 1342: trivial: as I mentioned elsewhere, either don't call this function when auto_window_update is 0, or have the function early-out when it's 0

source/h2_connection.c
```diff
@@ -2180,47 +2261,24 @@ static void s_connection_update_window(struct aws_http_connection *connection_ba
...
+             * Be more user friendly, if the increment size is too large, we will just saturate it to the max.
+             * AWS_H2_WINDOW_UPDATE_MAX will be checked during the actual sending of the window update frame.
+             */
+            connection->synced_data.pending_window_update_size_sync =
+                aws_add_u64_saturating(connection->synced_data.pending_window_update_size_sync, increment_size);
+            connection_open = connection->synced_data.is_open;
         }
         s_unlock_synced_data(connection);
     } /* END CRITICAL SECTION */
-    if (err) {
```
graebm, line 2216: 🎉 yay less chance for errors!

source/h2_stream.c
```diff
@@ -209,24 +209,62 @@ static struct aws_h2err s_check_state_allows_frame_type(
     return aws_h2err_from_h2_code(h2_error_code);
 }
 
-static int s_stream_send_update_window_frame(struct aws_h2_stream *stream, size_t increment_size) {
+static int s_stream_send_update_window_if_needed(struct aws_h2_stream *stream, uint64_t window_size) {
```
graebm, line 212: trivial: again, ~window_size~ sounds like the absolute value, but it's a relative value
```suggestion
static int s_stream_send_update_window_if_needed(struct aws_h2_stream *stream, uint64_t window_update_size) {
```

source/h2_stream.c
```diff
@@ -724,6 +740,13 @@ int aws_h2_stream_on_activated(struct aws_h2_stream *stream, enum aws_h2_stream_
     stream->thread_data.window_size_self =
         connection->thread_data.settings_self[AWS_HTTP2_SETTINGS_INITIAL_WINDOW_SIZE];
 
+    if (connection->stream_window_size_threshold_to_send_update) {
+        stream->window_size_threshold_to_send_update = connection->stream_window_size_threshold_to_send_update;
+    } else {
+        stream->window_size_threshold_to_send_update =
```
graebm, line 746:
```suggestion
        /* Set reasonable default of: 50% initial window size (same as Netty's DEFAULT_WINDOW_UPDATE_RATIO) */
        stream->window_size_threshold_to_send_update =
```

source/h2_stream.c
```diff
@@ -209,24 +209,62 @@ static struct aws_h2err s_check_state_allows_frame_type(
...
+            window_size,
+            stream->thread_data.window_size_self,
+            stream->window_size_threshold_to_send_update,
+            stream->thread_data.pending_window_update_size_thread);
+        return AWS_OP_SUCCESS;
+    }
 
-    if (!stream_window_update_frame) {
+    /* Cap the window to AWS_H2_WINDOW_UPDATE_MAX */
+    int32_t previous_window = stream->thread_data.window_size_self;
```
graebm, line 233: trivial: same thing again where we don't need this variable
```suggestion
```

source/h2_stream.c
```diff
@@ -209,24 +209,62 @@ static struct aws_h2err s_check_state_allows_frame_type(
...
+    if (window_delta > 0) {
+        struct aws_h2_frame *stream_window_update_frame =
+            aws_h2_frame_new_window_update(stream->base.alloc, stream->base.id, window_delta);
+        if (!stream_window_update_frame) {
+            AWS_H2_STREAM_LOGF(
+                ERROR,
+                stream,
+                "WINDOW_UPDATE frame on stream failed to be sent, error %s",
+                aws_error_name(aws_last_error()));
+            stream->thread_data.window_size_self = previous_window;
```
graebm, line 255:
```suggestion
```

graebm (approved):
all my feedback was style stuff. Looks good overall

PR #390, opened 2022-09-22
Destroy callback for http stream

TingDaoK:
- Add destroy callback for http stream so that user can just track one callback instead of the status of the stream


source/request_response.c
```diff
@@ -1070,11 +1070,18 @@ void aws_http_stream_release(struct aws_http_stream *stream) {
...
+
         struct aws_http_connection *owning_connection = stream->owning_connection;
         stream->vtable->destroy(stream);
 
         /* Connection needed to outlive stream, but it's free to go now */
         aws_http_connection_release(owning_connection);
+        if (on_destroy_callback) {
+            /* info user that destroy completed. */
+            on_destroy_callback(user_data);
+        }
```
graebm, line 1084: trivial: let's invoke the stream-destroy-callback before releasing the connection

it probably doesn't matter, but it seems a bit weird for actual destroy order to be:
stream then connection

but have the on-destroy-callbacks fire:
connection then stream

graebm (approved):
fix & ship!
nice work

PR #391, opened 2022-10-10
Event loop pin

bretambrose:
* Adds support for fixing the event loop used to house http and websocket connections


tests/test_connection.c
```diff
@@ -395,7 +425,8 @@ static int s_tester_clean_up(struct tester *tester) {
     aws_server_bootstrap_release(tester->server_bootstrap);
     aws_client_bootstrap_release(tester->client_bootstrap);
     aws_host_resolver_release(tester->host_resolver);
-    aws_event_loop_group_release(tester->event_loop_group);
+    aws_event_loop_group_release(tester->client_event_loop_group);
+    aws_event_loop_group_release(tester->server_event_loop_group);
```
sbSteveK, line 429: Assuming it's fine to release the same event loop group twice in the event that we're not using pin_event_loop and client_event_loop_group == server_event_loop_group.

sbSteveK (approved):

tests/test_connection.c
sbSteveK: it does a NULL check -_- 👍

bretambrose: When we use the same for both, we do a double acquire, requiring a double release

Issue #393, opened 2022-10-18
h2_sm_acquire_stream fails at Arch Linux


anatol:
I am trying to build aws-c-http project at Arch Linux.

The project compiles fine. But 2 tests are failing with the following cryptic error. Could anyone explain what is going on here?

```
645/647 Test #645: random_access_set_exist_test .....................................................   Passed    0.01 sec
        Start 646: random_access_set_remove_test
646/647 Test #646: random_access_set_remove_test ....................................................   Passed    0.01 sec
        Start 647: random_access_set_owns_element_test
647/647 Test #647: random_access_set_owns_element_test ..............................................   Passed    0.01 sec

99% tests passed, 2 tests failed out of 647

Total Test time (real) =  54.32 sec

The following tests FAILED:
	638 - h2_sm_acquire_stream (Failed)
	639 - h2_sm_acquire_stream_multiple_connections (Failed)
Errors while running CTest
Output from these tests are in: /mnt/nvme/sources/arch/community/aws-c-http/trunk/src/aws-c-http-0.6.22/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
make: *** [Makefile:71: test] Error 8
```

```
[TRACE] [2022-10-18T17:13:30Z] [00007f84977fe6c0] [socket] - id=0x7f8484001140 fd=14: read of 175
[TRACE] [2022-10-18T17:13:30Z] [00007f84977fe6c0] [socket-handler] - id=0x7f8484012d00: read 175 from socket
[TRACE] [2022-10-18T17:13:30Z] [00007f84977fe6c0] [channel] - id=0x7f8484001da0: sending read message of size 175, from slot 0x7f8484006330 to slot 0x7f84840062e0 with handler 0x7f8484012f40.
[TRACE] [2022-10-18T17:13:30Z] [00007f84977fe6c0] [tls-handler] - id=0x7f8484012f40: Downstream window 18446744073709551615
[TRACE] [2022-10-18T17:13:30Z] [00007f84977fe6c0] [channel] - id=0x7f8484001da0: acquired message 0x7f848400e580 of capacity 16384 from pool 0x7f8484001b60. Requested size was 18446744073709551615
[TRACE] [2022-10-18T17:13:30Z] [00007f84977fe6c0] [tls-handler] - id=0x7f8484012f40: Bytes read 13
[...]

PR #395, opened 2022-10-28
Empty path

DmitriyMusatkin:
*Issue #, if available:*
https://github.com/awslabs/aws-c-http/issues/393

*Description of changes:*
URI spec allows for path to be empty.
However, HTTP spec says that paths that are empty should be sent as /.
Our URI parser was recently changed to treat empty path as is to be spec compliant.
This change updates http client to treat empty path as / within http request

### DIFF
--- a/source/request_response.c
+++ b/source/request_response.c
@@ -646,7 +646,14 @@ int aws_http_message_set_request_path(struct aws_http_message *request_message, 
     if (request_message->request_data) {
         switch (request_message->http_version) {
             case AWS_HTTP_VERSION_1_1:
+            {
+                /* Its valid for uri path to be empty, but http spec requires empty paths to
+                be sent as "/", so default it here. */
+                if (path.len == 0) {
+                    path = s_default_empty_path;
+                }
                 return s_set_string_from_cursor(&request_message->request_data->path, path, request_message->allocator);
+            }
             case AWS_HTTP_VERSION_2:
                 return aws_http2_headers_set_request_path(request_message->headers, path);
             default:
--- a/source/request_response.c
+++ b/source/request_response.c
@@ -372,7 +372,7 @@ int aws_http2_headers_get_request_path(const struct aws_http_headers *h2_headers
 
         return AWS_OP_ERR;
     }
-        
+
     return AWS_OP_SUCCESS;
 }
 
@@ -683,14 +683,12 @@ int aws_http_message_get_request_path(
 
     if (request_message->request_data) {
         switch (request_message->http_version) {
-            case AWS_HTTP_VERSION_1_1:
-                {
-                    *out_path = request_message->request_data->path != NULL ? 
-                        aws_byte_cursor_from_string(request_message->request_data->path) :
-                        s_default_empty_path;
-                    return AWS_OP_SUCCESS;
-                }
-                break;
+            case AWS_HTTP_VERSION_1_1: {
+                *out_path = request_message->request_data->path != NULL
+                                ? aws_byte_cursor_from_string(request_message->request_data->path)
+                                : s_default_empty_path;
+                return AWS_OP_SUCCESS;
+            } break;
             case AWS_HTTP_VERSION_2:
                 return aws_http2_headers_get_request_path(request_message->headers, out_path);
             default:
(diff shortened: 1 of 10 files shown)


waahm7 (approved):

source/request_response.c
```diff
@@ -630,15 +630,24 @@ int aws_http_message_get_request_method(
...
+    }
+
     if (request_message->request_data) {
         switch (request_message->http_version) {
-            case AWS_HTTP_VERSION_1_1:
+            case AWS_HTTP_VERSION_1_1: {
                 return s_set_string_from_cursor(&request_message->request_data->path, path, request_message->allocator);
+            }
             case AWS_HTTP_VERSION_2:
                 return aws_http2_headers_set_request_path(request_message->headers, path);
```
TingDaoK, line 652: Let's also update for `aws_http2_headers_set_request_path`, when the path cursor is empty, use the default one

source/request_response.c
```diff
@@ -630,15 +630,24 @@ int aws_http_message_get_request_method(
...
+
 int aws_http_message_set_request_path(struct aws_http_message *request_message, struct aws_byte_cursor path) {
     AWS_PRECONDITION(request_message);
     AWS_PRECONDITION(aws_byte_cursor_is_valid(&path));
     AWS_PRECONDITION(request_message->request_data);
 
+    /* Its valid for uri path to be empty, but http spec requires empty paths to
+    be sent as "/", so default it here. */
+    if (path.len == 0) {
+        path = s_default_empty_path;
```
graebm, line 643: You should change aws_http_message_get_request_path() to return "/" when the path isn't set yet.

It feels weird if we're autocorrecting empty paths to be "/" but we're no autocorrecting unset paths to be "/"

graebm (changes requested):

source/request_response.c
DmitriyMusatkin: agree. addressed in the next commit.

source/request_response.c
DmitriyMusatkin: i missed that aws_http2_headers_set_request_path was public api. added logic to both paths

source/request_response.c
```diff
@@ -361,7 +363,17 @@ int aws_http2_headers_get_request_authority(
 }
 
 int aws_http2_headers_get_request_path(const struct aws_http_headers *h2_headers, struct aws_byte_cursor *out_path) {
-    return aws_http_headers_get(h2_headers, aws_http_header_path, out_path);
+    if (aws_http_headers_get(h2_headers, aws_http_header_path, out_path)) {
+        if (aws_last_error() == AWS_ERROR_HTTP_HEADER_NOT_FOUND) {
+            *out_path = s_default_empty_path;
+            aws_reset_error();
+            return AWS_OP_SUCCESS;
+        }
```
graebm, line 371: I take it all back don't make any of these changes

A principal we followed when designing all this was "keep requests exactly as-is".
So we don't automagically insert any required headers, like "host"
and in HTTP/2 there are many required headers, and `:path:` is just one of them
and now we're adding special handling for that 1
and this change would introduce bugs because the HTTP/2 implementation just iterates over the headers and sends them as they are, it doesn't specifically call `get_request_path()` so it wouldn't get this magic override

and now this is all seeming like a slippery slope into "automagic complexity" town

so I take it all back
just handle the edge-case back in your code like

```
if (uri.len == 0)
    aws_http_message_set_request_path("/");
```

graebm (commented):
:(
sorry for being wishy washy

source/request_response.c
DmitriyMusatkin: reverted it in the latest commit

TingDaoK:
I think what Mike meant is to revert all the change about handling empty path from `set_request_path` as well. I also agreed. If people called `set_request_path` with “”, that’s how it is. So, just change the tests to set path to be "/" when it's empty.

However, I think it will be nice if we can handle it from our end as it’s a requirement from HTTP spec. It's basically everyone wants (No one wants to break the spec with empty path). But, I’d prefer to handle it from encoder. So, for h1, it's h1_encoder.c, and for h2, it's hpack_encoder.c

graebm (approved):

TingDaoK (approved):

PR #402, opened 2022-11-30
Proxy h2

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- add support for using proxy to setup http/2 connection
- right now only supports talking h1 with proxy and create the http/2 connection from the tunneling proxy. 
- Turns out google doesn't like `:authority` with `host`, I am totally okay with what ever the real world want. :)

### TODO:
The `AWS_TEST_HTTPS_H2_PROXY_URL` now has a proxy host that can talk http/2 with client. Used apache server with mod_proxy_http2. You can simply use that host for the proxy option and update the tls of proxy options to have h2 in the alpn string to create the h2 connection with proxy. However, I don't know what's the expected behavior for client to talk to a h2 proxy host. 

**`curl` doesn't seem supporting it.**

So, put it as TODO, but still have the host here incase we want to implement it in the future.


include/aws/http/private/proxy_impl.h
```diff
@@ -103,7 +103,10 @@ struct aws_http_proxy_user_data {
     struct aws_socket_options original_socket_options;
     bool original_manual_window_management;
     size_t original_initial_window_size;
+    bool prior_knowledge_http2;
     struct aws_http1_connection_options original_http1_options;
+    struct aws_http2_connection_options original_http2_options; /* allocated with userdata */
```
graebm, line 108: not sure why `/* allocated with userdata */` comment is on `original_http2_options` which is a by-value thing

graebm (approved):
some trivial feedback, but looks good overall

include/aws/http/private/connection_impl.h
```diff
@@ -140,6 +140,8 @@ struct aws_http_client_bootstrap {
     struct aws_http_connection *connection;
 };
 
+int aws_validate_http_client_connection_options(const struct aws_http_client_connection_options *options);
```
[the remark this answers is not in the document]
TingDaoK: I was trying to use from proxy connect, but figure out it doesn't fit well. will revert the change

include/aws/http/private/proxy_impl.h
TingDaoK: The options has setting array, which need to be allocated from heap. The comment was for that, will clarify it more.

PR #409, opened 2022-12-22
Breaking change to websocket API: setup callback can report body of response

graebm:
**Issue:**
If the server rejected a websocket connection, there was no way to see the body of the response. The response body often has useful information explaining why the connection was rejected.

**Underlying cause:**
The websocket code would close the connection immediately after receiving headers, because that was the simplest way to do things.

**Description of changes:**
Change the setup callback signature, so that the body can be reported.

The setup code keeps the connection open while receiving a rejection, making a copy of the body, and ultimately reporting it in the setup callback.


include/aws/http/websocket.h
```diff
@@ -37,27 +37,35 @@ enum aws_websocket_opcode {
...
-    int error_code,
-    int handshake_response_status,
-    const struct aws_http_header *handshake_response_header_array,
-    size_t num_handshake_response_headers,
-    void *user_data);
+struct aws_websocket_on_connection_setup_data {
+    int error_code;
+    struct aws_websocket *websocket;
+    const int *handshake_response_status;
+    const struct aws_http_header *handshake_response_header_array;
```
TingDaoK, line 57: Why not use `const struct *aws_http_headers`? Do we really need to have a deep copy of the headers to an array here?

graebm: I actually did that, and then reverted it.

There were some downstream uses where the array was being forwarded along, so they would need to deconstruct it back into a raw array to preserve their own APIs.

graebm: and it's not a "deep copy" it's just one allocation long enough to hold an array.
it's still using the underlying strings stored in the aws_http_headers struct

source/websocket_bootstrap.c
```diff
@@ -559,6 +516,55 @@ static int s_ws_bootstrap_on_handshake_response_header_block_done(
...
+    ws_bootstrap->got_full_response_headers = true;
+
+    if (header_block == AWS_HTTP_HEADER_BLOCK_INFORMATIONAL) {
+        if (ws_bootstrap->response_status == AWS_HTTP_STATUS_CODE_101_SWITCHING_PROTOCOLS) {
+            /* OK, got 101 response, proceed with upgrade! */
+            return s_ws_bootstrap_validate_response_and_install_websocket_handler(ws_bootstrap, http_connection);
+
+        } else {
+            /* It would be weird to get any other kind of 1xx response, but anything is possible.
+             * Another, response should come eventually. Just forget the headers from this one... */
```
TingDaoK, line 548:
```suggestion
             * Another response should come eventually. Just ignore the headers from this one... */
```

TingDaoK (approved):

PR #410, opened 2022-12-24
Validate websocket handshake response

graebm:
**Issue:**
The websocket client did not fully validate the server's response, as specified in [RFC-6455 Section 4.1](https://www.rfc-editor.org/rfc/rfc6455#section-4.1)

The client was doing 1/6 required steps (checking the status-code, but not checking any headers)

**Description of changes:**
Validate the "Upgrade", "Connection", and "Sec-WebSocket-Accept" headers.

Now we're doing 4/6 required steps. Still need to check "Sec-WebSocket-Extensions" and "Sec-WebSocket-Protocol". That will be in a followup PR.


include/aws/http/websocket.h
```diff
@@ -245,14 +245,14 @@ struct aws_websocket_client_connection_options {
 
     /**
      * Called repeatedly as payload data arrives.
-     * Required if `on_incoming_frame_begin` is set.
+     * Optional.
```
graebm, line 248: I don't know why I made these 3 callbacks have a weird "all or none" requirement.
Now they're just 3 independent callbacks, you can set any combination you want.
No reason this had to change, it was just weird and I was touching nearby code

source/websocket.c
```diff
@@ -1373,16 +1373,14 @@ static int s_decoder_on_payload(struct aws_byte_cursor data, void *user_data) {
 
 /* Invoke user cb */
 static int s_decoder_on_user_payload(struct aws_websocket *websocket, struct aws_byte_cursor data) {
-    if (!websocket->on_incoming_frame_payload) {
-        return AWS_OP_SUCCESS;
-    }
+    if (websocket->on_incoming_frame_payload) {
+        if (!websocket->on_incoming_frame_payload(
```
DmitriyMusatkin, line 1377: really minor: does it have to be 2 ifs? if (cb && !cb(...)) seems like a common pattern

DmitriyMusatkin: looks like the old code was exiting early if no callback was defined and now we are updating window and doing other things. was it a bug in old code?

source/websocket_bootstrap.c
```diff
@@ -160,10 +165,16 @@ int aws_websocket_client_connect(const struct aws_websocket_client_connection_op
     ws_bootstrap->websocket_frame_payload_callback = options->on_incoming_frame_payload;
     ws_bootstrap->websocket_frame_complete_callback = options->on_incoming_frame_complete;
     ws_bootstrap->handshake_request = aws_http_message_acquire(options->handshake_request);
+    aws_byte_buf_init(&ws_bootstrap->expected_sec_websocket_accept, ws_bootstrap->alloc, 0);
```
TingDaoK, line 168: super trivial: Can we init the buffer within `s_ws_bootstrap_calculate_sec_websocket_accept`? Just that `aws_byte_buf_init` and `aws_byte_buf_reserve` to one single `aws_byte_buf_init`? Unless we plan to use `s_ws_bootstrap_calculate_sec_websocket_accept` somewhere else that already have a buffer initialized.

TingDaoK (approved):

source/websocket.c
graebm: All the callbacks in this file are invoked with 2 ifs: one to check whether to invoke the callback, and the next to invoke it and see if it fails

I think it's easier to read

source/websocket_bootstrap.c
graebm: sure

DmitriyMusatkin (approved):

source/websocket_bootstrap.c
```diff
@@ -229,11 +241,76 @@ static void s_ws_bootstrap_destroy(struct aws_websocket_client_bootstrap *ws_boo
...
+    struct aws_byte_buf *out_buf,
+    struct aws_allocator *alloc) {
+
+    AWS_ASSERT(out_buf && !out_buf->allocator && out_buf->len == 0); /* expect buf to be uninitialized */
+
+    /* ignore leading and trailing whitespace */
+    sec_websocket_key = aws_strutil_trim_http_whitespace(sec_websocket_key);
+
+    /* concatenation of key with magic string (store temporarily in out_buf) */
+    struct aws_byte_cursor magic_string = aws_byte_cursor_from_c_str("258EAFA5-E914-47DA-95CA-C5AB0DC85B11");
```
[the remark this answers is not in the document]
graebm: ooh. clever!
doing it...

PR #420, opened 2023-01-10
Stop hitting example.com in tests

graebm:
it's timing out too often

### DIFF
--- a/integration-testing/http_client_test.py
+++ b/integration-testing/http_client_test.py
@@ -82,7 +82,7 @@ class SimpleTests(unittest.TestCase):
         run_command(simple_get_args)
     def test_simple_get_h1(self):
         """make a simple GET request via HTTP/1.1 and make sure it succeeds"""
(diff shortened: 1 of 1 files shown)


integration-testing/http_client_test.py
```diff
@@ -82,7 +82,7 @@ def test_simple_get_google(self):
         run_command(simple_get_args)
     def test_simple_get_h1(self):
         """make a simple GET request via HTTP/1.1 and make sure it succeeds"""
-        simple_get_args = elasticurl_cmd_prefix + ['-v', 'TRACE', '--http1_1', 'http://example.com']
```
graebm, line 85: using `httpbin.org` instead of `www.amazon.com` here because there's a separate test `test_simple_get_amazon` that explicitly hits `www.amazon.com`. Other tests were already hitting `httpbin.org` so this doesn't add any NEW dependencies

TwistedTwigleg (approved):

TingDaoK (approved):

PR #421, opened 2023-01-11
Add ProxyConfig new from ProxyOptions & TLS info

waahm7:
*Description of changes:*
- Adds a new function proxy_config new from proxy_options & tls which will be used by `s3_client_new` to make a copy of proxy options.

### DIFF
--- a/include/aws/http/proxy.h
+++ b/include/aws/http/proxy.h
@@ -507,18 +507,16 @@ struct aws_http_proxy_config *aws_http_proxy_config_new_from_proxy_options(
     const struct aws_http_proxy_options *options);
 
 /**
- * Create a persistent proxy configuration from non-persistent proxy options.
+ * @brief Determine proxy_connection_type based on tls_options configuration if connection_type is legacy.
  *
- * @param allocator memory allocator to use
- * @param options http proxy options to source proxy configuration from
- * @param tls_connection_options tls connection options to determine connection_type if connection type is legacy
- * @return
+ * @param proxy_connection_type
+ * @param tls_options
+ * @return AWS_HTTP_API enum
  */
 AWS_HTTP_API
-struct aws_http_proxy_config *aws_http_proxy_config_new_from_proxy_options_with_tls_options(
-    struct aws_allocator *allocator,
-    const struct aws_http_proxy_options *proxy_options,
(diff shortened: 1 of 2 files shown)


include/aws/http/proxy.h
```diff
@@ -506,6 +506,20 @@ struct aws_http_proxy_config *aws_http_proxy_config_new_from_proxy_options(
     struct aws_allocator *allocator,
     const struct aws_http_proxy_options *options);
 
+/**
+ * Create a persistent proxy configuration from non-persistent proxy options.
+ *
+ * @param allocator memory allocator to use
+ * @param options http proxy options to source proxy configuration from
+ * @param tls_connection_options tls connection options to determine connection_type if connection type is legacy
```
TingDaoK, line 514: Why not expose `s_determine_proxy_connection_type` instead?

waahm7: `s_determine_proxy_connection_type` is an implementation detail. This gives us the the flexibility to change implementation later without caring about breaking consumers.

TingDaoK: The whole point of using `tls_connection_options` is to determine what connection type is. Just have a simple if/else maybe better.

waahm7: I have updated the `s_determine_proxy_connection_type` to use a bool instead so that we can have simple if/else. If we make expose `s_determine_proxy_connection_type`, we will have to expose `s_aws_http_proxy_config_new` as well.

TingDaoK (approved):
fix & ship

include/aws/http/proxy.h
```diff
@@ -506,6 +506,20 @@ struct aws_http_proxy_config *aws_http_proxy_config_new_from_proxy_options(
     struct aws_allocator *allocator,
     const struct aws_http_proxy_options *options);
 
+/**
+ * Create a persistent proxy configuration from non-persistent proxy options.
+ *
+ * @param allocator memory allocator to use
+ * @param options http proxy options to source proxy configuration from
+ * @param is_tls_connection tls connection info to determine connection_type if connection type is legacy
```
TingDaoK, line 514: be specific about this `is_tls_connection` is talking about the main connection

PR #422, opened 2023-01-30
Adapt to renamed class from aws-c-common

graebm:
Adapt to: https://github.com/awslabs/aws-c-common/pull/981

### DIFF
--- a/include/aws/http/private/websocket_decoder.h
+++ b/include/aws/http/private/websocket_decoder.h
@@ -42,7 +42,7 @@ struct aws_websocket_decoder {
     /* True while processing a TEXT "message" (from the start of a TEXT frame,
      * until the end of the TEXT or CONTINUATION frame with the FIN bit set). */
     bool processing_text_message;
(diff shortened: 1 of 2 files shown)


include/aws/http/private/websocket_decoder.h
```diff
@@ -42,7 +42,7 @@ struct aws_websocket_decoder {
     /* True while processing a TEXT "message" (from the start of a TEXT frame,
      * until the end of the TEXT or CONTINUATION frame with the FIN bit set). */
     bool processing_text_message;
-    struct aws_utf8_validator *text_message_validator;
```
waahm7, line 45: trivial: We should rename the variable as well.

waahm7 (approved):

include/aws/http/private/websocket_decoder.h
graebm: keep it as is.
I realized they're different, but I'm still only doing validation here, so I think this name is better

PR #430, opened 2023-03-15
Throw better errors for h2 manual write instead of invalid state error

waahm7:
*Description of changes:*
- Throw `AWS_ERROR_HTTP_MANUAL_WRITES_NOT_ENABLED` if manual writes are not enabled but someone tries to call it.
- Throw `AWS_ERROR_HTTP_STREAM_NOT_ACTIVATED` or `AWS_ERROR_HTTP_STREAM_HAS_COMPLETED` if stream is not in active state.
- Throw `AWS_ERROR_HTTP_MANUAL_WRITES_HAS_COMPLETED` if someone calls manualWrite after sending end_stream.


source/h2_stream.c
```diff
@@ -1254,6 +1254,14 @@ static int s_stream_write_data(
     struct aws_http_stream *stream_base,
     const struct aws_http2_stream_write_data_options *options) {
     struct aws_h2_stream *stream = AWS_CONTAINER_OF(stream_base, struct aws_h2_stream, base);
+    if (!stream->manual_write) {
```
ganeshnj, line 1257: this would have saved some of debugging time.

ganeshnj (approved):

source/h2_stream.c
```diff
@@ -1254,6 +1254,14 @@ static int s_stream_write_data(
...
+        AWS_LOGF_ERROR(
+            AWS_LS_HTTP_STREAM,
+            "Manual writes are not enabled. You need to enable manual writes using by setting "
+            "`http2_use_manual_data_writes` to true in `aws_http_make_request_options`, stream=%p",
```
graebm, line 1261: trivial: I have no idea if there's anything wrong with backticks in a log message, but it is unusual? single-quote is more typical

Also, use `AWS_H2_STREAM_LOGF()` so we format our log messages consistently from this file. Looks like a few vanilla `AWS_LOGF_ERROR()` slipped into this file; they should probably be changed too

tests/test_h2_client.c
```diff
@@ -5336,6 +5336,59 @@ TEST_CASE(h2_client_manual_data_write) {
...
+
+    /* Try writing the data */
+    struct aws_input_stream *data_stream = s_h2_client_manual_data_write_generate_data(&test_ctx);
+    int64_t stream_length = 0;
+    ASSERT_SUCCESS(aws_input_stream_get_length(data_stream, &stream_length));
+    struct aws_http2_stream_write_data_options write_options = {
+        .data = data_stream,
+    };
+    ASSERT_FAILS(aws_http2_stream_write_data(stream, &write_options));
+    ASSERT_INT_EQUALS(aws_last_error(), AWS_ERROR_HTTP_MANUAL_WRITE_NOT_ENABLED);
```
graebm, line 5378: we have a special assert for this
```suggestion
    ASSERT_ERROR(AWS_ERROR_HTTP_MANUAL_WRITE_NOT_ENABLED, aws_http2_stream_write_data(stream, &write_options));
```
or keep it as two lines, but in our macros you put the expected value first (if it fails it prints out that it expected the value to be A but got B)
```suggestion
    ASSERT_FAILS(aws_http2_stream_write_data(stream, &write_options));
    ASSERT_INT_EQUALS(AWS_ERROR_HTTP_MANUAL_WRITE_NOT_ENABLED, aws_last_error());
```

source/h2_stream.c
waahm7: Thanks, updated.

graebm (approved):

PR #431, opened 2023-03-17
Handle connection close corner case

TingDaoK:
*Issue #, if available:*

https://github.com/awslabs/aws-c-s3/issues/264

*Description of changes:*

- When connection closed and the response finished for the outgoing stream, which still have not finished sending data, we should treat the stream finished successfully.
- consistent with [spec](https://httpwg.org/specs/rfc9112.html#persistent.tear-down) about receiving connection: closed
> A client that receives a "close" connection option MUST cease sending requests on that connection and close the connection after reading the response message containing the "close" connection option


source/h1_connection.c
```diff
@@ -1066,6 +1080,27 @@ static int s_decoder_on_header(const struct aws_h1_decoded_header *header, void
...
                 aws_h1_connection_unlock_synced_data(connection);
             } /* END CRITICAL SECTION */
+
+            /* For server side, server should finish sending the response. */
+            if (connection->base.client_data) {
+                /**
+                 * RFC-9112 section 9.6.
+                 * A client that receives a "close" connection option MUST cease sending
+                 *requests on that connection and close the connection after reading the
+                 *response message containing the "close" connection option.
```
graebm, line 1090: Ugh. Even in the new RFC this is ambiguous. Is it talking about "don't start any further requests" or saying "don't complete sending the current request"? You didn't capture the full sentence here. There's a semicolon and it goes on to talk specifically about pipelining, which makes me thing this means "don't start any further requests"

> A client that receives a "close" connection option MUST cease sending requests on that connection and close the connection after reading the response message containing the "close" connection option; if additional pipelined requests had been sent on the connection, the client SHOULD NOT assume that they will be processed by the server.

source/h1_connection.c
```diff
@@ -1066,6 +1080,27 @@ static int s_decoder_on_header(const struct aws_h1_decoded_header *header, void
                 connection->synced_data.new_stream_error_code = AWS_ERROR_HTTP_CONNECTION_CLOSED;
                 aws_h1_connection_unlock_synced_data(connection);
             } /* END CRITICAL SECTION */
+
+            /* For server side, server should finish sending the response. */
+            if (connection->base.client_data) {
```
graebm, line 1085: trivial: this comment confused me, more than helped me

I expect a comment before an if-statement to summarize what the if-statement is about
But this comment is explaining the else-case, before I've read what the if-statement even says.

source/h1_connection.c
graebm: Up in [section 9.5](https://www.rfc-editor.org/rfc/rfc9112#section-9.5-5) there's this:
> A client sending a message body SHOULD monitor the network connection for an error response while it is transmitting the request. If the client sees a response that indicates the server does not wish to receive the message body and is closing the connection, the client SHOULD immediately cease transmitting the body and close its side of the connection.

So ... we should definitely stop sending after a 404 Not Found "Connection: close"

BUT if we got a 100 Continue "Connection: close" 😳 we should NOT stop sending, because we still haven't received the full response.

But what about a 200 OK "Connection: close"? What's the right course of action there?

TingDaoK: Yeah, it's a bit ambiguous, but I think we can still just close connection and stop sending for those response. It will not hurt for the connection that server will stop sending anymore and only waits for the `corresponding close by the client`

source/h1_connection.c
```diff
@@ -546,6 +546,34 @@ static void s_stream_complete(struct aws_h1_stream *stream, int error_code) {
...
     }
 
+    if (error_code != AWS_ERROR_SUCCESS) {
+        if (stream->base.client_data && stream->is_incoming_message_done && stream->is_outgoing_message_done == false) {
+            /* As a request that finished receiving the response, we ignore error and
+             * consider it finished successfully */
+            AWS_LOGF_DEBUG(
+                AWS_LS_HTTP_STREAM,
+                "id=%p: Request stream haven't finished sending, but error code %d (%s) happened and full response "
+                "received. Stop sending and treated as succeed",
```
graebm, line 556:
```suggestion
                "id=%p: Ignoring error code %d (%s). The response has been fully received,
                "so the stream will complete successfully."
```

source/h1_connection.c
```diff
@@ -546,6 +546,34 @@ static void s_stream_complete(struct aws_h1_stream *stream, int error_code) {
...
+            error_code = AWS_ERROR_SUCCESS;
+        }
+        if (stream->base.server_data && stream->is_outgoing_message_done && stream->is_incoming_message_done == false) {
+            /* As a server finished sending the response, but still failed with the request was not finished receiving.
+             * We ignore error and consider it finished successfully
+             */
+            AWS_LOGF_DEBUG(
+                AWS_LS_HTTP_STREAM,
+                "id=%p: Server finishes sending response, but error code %d (%s) happens before receives the full "
+                "response. Treated as succeed",
```
graebm, line 569: similar feedback, do we need to check `is_incoming_message_done`?

and maybe

"id=%p: Ignoring error code %d (%s). The response has  been fully sent,"
" so the stream will complete successfully"

source/h1_connection.c
```diff
@@ -1066,6 +1094,31 @@ static int s_decoder_on_header(const struct aws_h1_decoded_header *header, void
...
             } /* END CRITICAL SECTION */
+
+            if (connection->base.client_data) {
+                /**
+                 * RFC-9112 section 9.6.
+                 * A client that receives a "close" connection option MUST cease sending
+                 * requests on that connection and close the connection after reading the
+                 * response message containing the "close" connection option.
+                 *
+                 * Mark the stream has no outgoing_message
```
graebm, line 1105:
```suggestion
                 * Mark the stream's outgoing message as complete,
                 * so that we stop sending, and stop waiting for it to finish sending.
```

source/h1_connection.c
```diff
@@ -1066,6 +1094,31 @@ static int s_decoder_on_header(const struct aws_h1_decoded_header *header, void
...
+                 **/
+                AWS_LOGF_DEBUG(
+                    AWS_LS_HTTP_STREAM,
+                    "id=%p: Request stream receives a \"close\" connection option. Stop writing to the connection "
```
graebm, line 1109:
```suggestion
                    "id=%p: Received 'Connection: close' header, no more request data will be sent."
```

graebm: [debatable] wonder if it's worth putting this within an if-block, so that it's not logged when the outgoing message is aready complete?
```
if (!incoming_stream->is_outgoing_message_done) {
    LOGF(... no more request data will be sent);
    incoming_stream->is_outgoing_message_done = true;
}
s_stop(...)
```

source/h1_connection.c
```diff
@@ -1066,6 +1094,31 @@ static int s_decoder_on_header(const struct aws_h1_decoded_header *header, void
...
+                    "anymore.",
+                    (void *)&incoming_stream->base);
+                incoming_stream->is_outgoing_message_done = true;
+                /* Stop writing right now and the shutdown will be scheduled right after finishing parsing the response
+                 */
```
graebm, line 1114: [utterly trivial] the autoformatted line wrap bugs me
```suggestion
                /* Stop writing right now.
                 * Shutdown will be scheduled after we finishing parsing the response *.
```

tests/CMakeLists.txt
```diff
@@ -142,6 +142,8 @@ add_test_case(h1_client_midchannel_requires_switching_protocols)
 add_test_case(h1_client_switching_protocols_fails_pending_requests)
 add_test_case(h1_client_switching_protocols_fails_subsequent_requests)
 add_test_case(h1_client_switching_protocols_requires_downstream_handler)
+add_test_case(h1_client_server_close_connection_before_request_finishes)
```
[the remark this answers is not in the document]
TingDaoK: It's still a client side test, maybe `h1_client_connection_close_before_request_finishes`

source/h1_connection.c
TingDaoK: I think it's worth to be more clear.

source/h1_connection.c
```diff
@@ -546,6 +546,34 @@ static void s_stream_complete(struct aws_h1_stream *stream, int error_code) {
         }
     }
 
+    if (error_code != AWS_ERROR_SUCCESS) {
+        if (stream->base.client_data && stream->is_incoming_message_done && stream->is_outgoing_message_done == false) {
+            /* As a request that finished receiving the response, we ignore error and
+             * consider it finished successfully */
```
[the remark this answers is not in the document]
TingDaoK: I think if we finish sending, something could go wrong from the callbacks. And, for those, it's worth to still error. But, it should not happen in real. Yeah, should remove the check.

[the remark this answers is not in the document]
graebm: oh damn, good point, if something goes wrong delivering the body to the user, that SHOULD fail things.

hmmm but looking at our code, we only set [is_incoming_message_done = true](https://github.com/awslabs/aws-c-http/blob/291b06cf7e4f53f8b558810be8ceee72e5f616f4/source/h1_connection.c#L1218) after we've completely successfully decoded the message. Any errors from callbacks would prevent stop us from setting, `is_incoming_message_done = true`. So we win!

source/h1_connection.c
```diff
@@ -546,6 +546,34 @@ static void s_stream_complete(struct aws_h1_stream *stream, int error_code) {
...
+                "so the stream will complete successfully.",
+                (void *)&stream->base,
+                error_code,
+                aws_error_name(error_code));
+            error_code = AWS_ERROR_SUCCESS;
+        }
+        if (stream->base.server_data && stream->is_outgoing_message_done) {
+            /* As a server finished sending the response, but still failed with the request was not finished receiving.
+             * We ignore error and consider it finished successfully
+             */
```
graebm, line 565: [utterly utterly trivial]
```suggestion
             * We ignore error and consider it finished successfully */
```

graebm (approved):

PR #433, opened 2023-04-05
HTTP stream telemetry api

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/h1_connection.c
```diff
@@ -1261,6 +1280,15 @@ static int s_decoder_on_done(void *user_data) {
...
     }
+    if (incoming_stream->base.on_metrics) {
+        aws_high_res_clock_get_ticks(&incoming_stream->base.metrics.receive_end_timestamp_ns);
+        AWS_ASSERT(incoming_stream->base.metrics.receive_start_timestamp_ns != 0);
+        AWS_ASSERT(
+            incoming_stream->base.metrics.receive_end_timestamp_ns >=
+            incoming_stream->base.metrics.receive_start_timestamp_ns);
+        incoming_stream->base.metrics.receiving_duration_ns = incoming_stream->base.metrics.receive_end_timestamp_ns -
+                                                              incoming_stream->base.metrics.receive_start_timestamp_ns;
+    }
```
waahm7, line 1291: We should move this block after 
```
    /* Otherwise the incoming stream is finished decoding and we will update it if needed */
    incoming_stream->is_incoming_message_done = true;
```
this won't be correct if we recieve an informational response.

source/h2_stream.c
```diff
@@ -706,7 +713,11 @@ int aws_h2_stream_on_activated(struct aws_h2_stream *stream, enum aws_h2_stream_
         AWS_H2_STREAM_LOGF(ERROR, stream, "Failed to create HEADERS frame: %s", aws_error_name(aws_last_error()));
         goto error;
     }
-
+    if (stream->base.on_metrics) {
+        stream->base.metrics.stream_id = stream->base.id;
+        AWS_ASSERT(stream->base.metrics.send_start_timestamp_ns == 0);
+        aws_high_res_clock_get_ticks(&stream->base.metrics.send_start_timestamp_ns);
+    }
```
waahm7, line 720: Similar to h1, this seems misleading. We haven’t actually started sending the data, only scheduled it.

source/h2_stream.c
```diff
@@ -734,6 +745,13 @@ int aws_h2_stream_on_activated(struct aws_h2_stream *stream, enum aws_h2_stream_
         }
     }
     aws_h2_connection_enqueue_outgoing_frame(connection, headers_frame);
+    if (stream->base.on_metrics && !with_data) {
+        /* The frame has not sent by the connection, but the stream is considered to end sending more data. */
+        AWS_ASSERT(stream->base.metrics.send_end_timestamp_ns == 0);
+        aws_high_res_clock_get_ticks(&stream->base.metrics.send_end_timestamp_ns);
+        stream->base.metrics.sending_duration_ns =
+            stream->base.metrics.send_end_timestamp_ns - stream->base.metrics.send_start_timestamp_ns;
+    }
```
waahm7, line 754: The `send_end_timestamp` represents "the time stamp when the request has finished being sent to the HTTP level". I expected this timestamp to include everything, such as the end frame, and for the stream to not send anything after this point. 

If we assume that send_end_timestamp only applies to the data, it can be confusing when there is no data to send. This could result in a misleading metric with a very short duration, indicating that the data was sent quickly even though there was nothing to send.

source/h2_stream.c
```diff
@@ -791,12 +809,19 @@ int aws_h2_stream_encode_data_frame(
...
     if (input_stream_complete && ends_stream) {
+        if (stream->base.on_metrics) {
+            AWS_ASSERT(stream->base.metrics.send_start_timestamp_ns != 0);
+            AWS_ASSERT(stream->base.metrics.send_end_timestamp_ns == 0);
+            aws_high_res_clock_get_ticks(&stream->base.metrics.send_end_timestamp_ns);
+            AWS_ASSERT(stream->base.metrics.send_end_timestamp_ns >= stream->base.metrics.send_start_timestamp_ns);
+            stream->base.metrics.sending_duration_ns =
+                stream->base.metrics.send_end_timestamp_ns - stream->base.metrics.send_start_timestamp_ns;
+        }
         /* Done sending data. No more data will be sent. */
```
waahm7, line 825: trivial: move this if block after the comment `/* Done sending data. No more data will be sent. */`

include/aws/http/request_response.h
```diff
@@ -194,6 +194,42 @@ typedef void(aws_http_on_stream_complete_fn)(struct aws_http_stream *stream, int
  */
 typedef void(aws_http_on_stream_destroy_fn)(void *user_data);
 
+/**
+ * Tracing metrics for aws_http_stream.
+ * Data maybe not be available if the data of stream was never sent/received before it completes.
+ */
+struct aws_http_stream_metrics {
+    /* The time stamp when the request started to be sent to HTTP level. 0 means data not available. */
+    uint64_t send_start_timestamp_ns;
```
DmitriyMusatkin, line 203: nit: its a bit unclear what http level refers to here? time when we gave the first byte of request to os networking stack? or something else?

include/aws/http/request_response.h
```diff
@@ -194,6 +194,42 @@ typedef void(aws_http_on_stream_complete_fn)(struct aws_http_stream *stream, int
  */
 typedef void(aws_http_on_stream_destroy_fn)(void *user_data);
 
+/**
+ * Tracing metrics for aws_http_stream.
+ * Data maybe not be available if the data of stream was never sent/received before it completes.
+ */
+struct aws_http_stream_metrics {
```
DmitriyMusatkin, line 201: a couple more things to consider adding: 
- tls overhead
- connection acquisition (do we have connection in the pool available or we have to get a new one)
- retries if we have visibility into http level retries
- number of bytes sent/received
- errors/status codes

source/h1_connection.c
```diff
@@ -731,6 +738,14 @@ static struct aws_h1_stream *s_update_outgoing_stream_ptr(struct aws_h1_connecti
     /* If current stream is done sending data... */
     if (current && !aws_h1_encoder_is_message_in_progress(&connection->thread_data.encoder)) {
         current->is_outgoing_message_done = true;
+        if (current->base.on_metrics) {
+            AWS_ASSERT(current->base.metrics.send_end_timestamp_ns == 0);
```
DmitriyMusatkin, line 742: if im reading docs correctly, aws_high_res_clock_get_ticks cannot be directly translated into system time. should we be more careful when exposing this value to user and warn them that timestamps are only good for relative operations?

include/aws/http/request_response.h
TingDaoK: - tls overhead is more a connection related metrics, we can add it as a separate connection level thing. 
- connection acquisition (do we have connection in the pool available or we have to get a new one)
  - that's connection manager level, we have `struct aws_http_manager_metrics` [here](https://github.com/awslabs/aws-c-http/blob/main/include/aws/http/connection_manager.h#L31), not really telling people the connection is acquired from poll or created, but provide an overall status of the connection 
- I don't think we have retries from http level (?)
-  number of bytes sent/received
  - I think that's something we can consider, even though we have callbacks for headers/body, the un-decoded bytes length can still be helpful.
- errors/status codes
  - There are complete callback that we provide the error code.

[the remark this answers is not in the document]
TingDaoK: I don't think we should keep the metrics private:
- we have `struct aws_http_manager_metrics` [here](https://github.com/awslabs/aws-c-http/blob/main/include/aws/http/connection_manager.h#L31), which we already exposed. 
- We are not having any memory that needs to be managed by us, and probably will not have any in the future, as we have header/body callbacks for the stream to provide those info. 
- For public metrics structure, we can still add metrics in the future,  we just cannot remove/change the existing one.

include/aws/http/request_response.h
TingDaoK: I was planned to use the time stamp sent to network stack, but it's awkward to do for h2. Since we may invoke the complete callback for h2 stream before any of the frame sent to the network stack, when the stream reset by client before any network stuff happens.

So, I changed to be `sent from http stream`, which is the time when stream encode the frame/data to the connection to send. For h1, it's the same as sending to the networking stack.

source/h1_connection.c
```diff
@@ -801,6 +816,10 @@ static struct aws_h1_stream *s_update_outgoing_stream_ptr(struct aws_h1_connecti
         s_set_outgoing_stream_ptr(connection, current);
 
         if (current) {
+            if (current->base.on_metrics) {
+                AWS_ASSERT(current->base.metrics.send_start_timestamp_ns == 0);
+                aws_high_res_clock_get_ticks(&current->base.metrics.send_start_timestamp_ns);
+            }
```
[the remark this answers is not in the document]
TingDaoK: We were not just scheduling the work, they all in the same thread, and get send right after it.

Here is the first time for the stream to be picked up by the connection to send. So, I think it's better to be here.

source/h2_stream.c
TingDaoK: I changed the document, it's meant to represent the time stamp when the stream finish sending to the connection.

source/h2_stream.c
TingDaoK: I updated the documentation, https://github.com/awslabs/aws-c-http/pull/433/files#r1163348036

source/h1_connection.c
TingDaoK: yeah, switched to aws_sys_clock_get_ticks

source/h1_connection.c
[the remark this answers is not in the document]
waahm7: Is it possible that we are just waiting for chunks after picking up the stream?

source/h2_stream.c
waahm7: I am still confused. How can we say the stream finished sending the data here, 'If the stream has no body, then the HEADERS frame marks the end of outgoing data'? Should we mark the end after the header is sent?

include/aws/http/request_response.h
waahm7: What about how much time we spend awaiting input?

include/aws/http/request_response.h
waahm7: I am bit confused about the difference between `from http stream` vs `http level`. It might be helpful to add some documentation about what it means by `from http stream`. In the case, where we may invoke complete callback before any of the frame sent to the network stack, I would expect the data to be not available.

Receive timestamp is still from http level, maybe forgot to update?

include/aws/http/request_response.h
JonathanHenson: don't we already have tls level metrics?

include/aws/http/request_response.h
```diff
@@ -234,6 +270,13 @@ struct aws_http_make_request_options {
      */
     aws_http_on_incoming_body_fn *on_response_body;
 
+    /**
+     * Invoked right before stream is complete, whether successful or unsuccessful
+     * Optional.
+     * See `aws_http_on_stream_metrics_fn`
+     */
+    aws_http_on_stream_metrics_fn *on_metrics;
```
graebm, line 278: [debatable] Would it be simpler to have a query function, instead of a callback? similar existing query functions such as aws_http_stream_get_incoming_response_status(...)

```
// data not available until the stream is complete
const struct aws_http_stream_metrics *aws_http_stream_get_metrics(const struct aws_http_stream *stream);
```

TingDaoK: I think if we use callbacks, we can skip gathering metrics if the callback was not set, which can improve the performance.

source/h2_connection.c
```diff
@@ -623,6 +624,19 @@ static void s_on_channel_write_complete(
 
     CONNECTION_LOG(TRACE, connection, "Message finished writing to network. Rescheduling outgoing frame task");
 
+    while (!aws_linked_list_empty(&connection->thread_data.finish_encoding_streams_list)) {
+        struct aws_linked_list_node *node =
+            aws_linked_list_pop_front(&connection->thread_data.finish_encoding_streams_list);
+        struct aws_h2_stream *stream = AWS_CONTAINER_OF(node, struct aws_h2_stream, node);
+        AWS_ASSERT(stream->base.metrics.send_start_timestamp_ns != 0);
+        AWS_ASSERT(stream->base.metrics.send_end_timestamp_ns == 0);
+        aws_high_res_clock_get_ticks(&stream->base.metrics.send_end_timestamp_ns);
```
graebm, line 633: it might be possible that we haven't set send_end_timestamp_ns yet, when we fire the completion callback

ughghhghgh I vote we just keep this simple, and forget about completion callbacks, and just take the time at our existing state-machine transitions, instead of adding complexity with more states

include/aws/http/connection.h
```diff
@@ -512,6 +512,12 @@ enum aws_http_version aws_http_connection_get_version(const struct aws_http_conn
 AWS_HTTP_API
 struct aws_channel *aws_http_connection_get_channel(struct aws_http_connection *connection);
 
+/**
+ * Returns the remote endpoint of the HTTP connection.
+ */
+AWS_HTTP_API
+struct aws_byte_cursor aws_http_connection_get_remote_endpoint_address(struct aws_http_connection *connection);
```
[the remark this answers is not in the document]
JonathanHenson: Yes to this! a.) The CRT's underlying transport will not always be TCP. In some cases it will be QUIC, in others it will be weird experimental stuff that aren't even using a socket api. If you return the whole struct we can convey the type of socket. We could even expand it over time to include things like MAC address as we need to without an API breakage. You could do it without an ABI break as well if you return it as ````const struct aws_socket_endpoint  *```` as well.

source/h2_connection.c
TingDaoK: We had another discussion later yesterday:

What about set the `send_end_timestamp_ns` from right before invoke complete callback if not set before.

So that, the `send_end_timestamp_ns` will be a bit off the real time at the corner case. We can document it out.

User will properly be interested in the time when the first byte is sent to the wire.

.github/workflows/ci.yml
```diff
@@ -158,7 +158,7 @@ jobs:
     - name: Build and test
       run: |
         python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder.pyz')"
-        python3 builder.pyz build -p aws-c-http --cmake-extra=-DENABLE_LOCALHOST_INTEGRATION_TESTS=ON
+        python3 builder.pyz build -p aws-c-http --cmake-extra=-DENABLE_LOCALHOST_INTEGRATION_TESTS=ON --config Debug
```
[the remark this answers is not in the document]
TingDaoK: I was using assert instead of real test to make sure the time metrics are gathered as expected. Enable debug for the most cover tests.

source/h1_connection.c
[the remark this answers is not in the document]
TingDaoK: It's not, we only encode one stream per time, when we update the stream, it will always be the initial state for the encoder. I can add an assert.

tests/stream_test_helper.c
```diff
@@ -116,6 +116,14 @@ static int s_on_body(struct aws_http_stream *stream, const struct aws_byte_curso
...
     return AWS_OP_SUCCESS;
 }
+static void s_on_metrics(
+    struct aws_http_stream *stream,
+    const struct aws_http_stream_metrics *metrics,
+    void *user_data) {
+    (void)stream;
+    struct client_stream_tester *tester = user_data;
+    tester->metrics = *metrics;
+}
```
waahm7, line 126: trivial: assert that all values are positive?

TingDaoK: we cannot really do that, because we have tests that the stream was never sent, so data can be unavailable

waahm7: We can add a separate test for metrics and maybe a test where data is unavailable.

source/h1_stream.c
```diff
@@ -361,6 +361,8 @@ static struct aws_h1_stream *s_stream_new_common(
     stream->base.on_incoming_body = on_incoming_body;
     stream->base.on_complete = on_complete;
     stream->base.on_destroy = on_destroy;
+    stream->base.metrics.receiving_duration_ns = -1;
+    stream->base.metrics.sending_duration_ns = -1;
```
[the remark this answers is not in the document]
TingDaoK: good catch

include/aws/http/private/h2_frames.h
```diff
@@ -239,6 +250,18 @@ struct aws_h2_frame *aws_h2_frame_new_headers(
...
+ * Same as `aws_h2_frame_new_headers`, but the frame will be associated with a stream and update the stream's metric
+ * when the frame is sent, if the stream has on_metrics callback set.
+ */
+AWS_HTTP_API
+struct aws_h2_frame *aws_h2_frame_new_headers_with_stream(
+    struct aws_h2_stream *h2_stream,
+    const struct aws_http_headers *headers,
+    bool end_stream,
+    uint8_t pad_length,
+    const struct aws_h2_frame_priority_settings *optional_priority);
```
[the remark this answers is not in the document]
waahm7: Multiple functions / a function which is only used in tests is confusing for someone who does not know the history of the codebase. We do this when we need to maintain backward compatibility and can avoid it in this case.

source/h1_stream.c
```diff
@@ -361,6 +361,12 @@ static struct aws_h1_stream *s_stream_new_common(
     stream->base.on_incoming_body = on_incoming_body;
     stream->base.on_complete = on_complete;
     stream->base.on_destroy = on_destroy;
+    stream->base.metrics.send_start_timestamp_ns = -1;
+    stream->base.metrics.send_end_timestamp_ns = -1;
+    stream->base.metrics.sending_duration_ns = -1;
+    stream->base.metrics.receive_start_timestamp_ns = -1;
+    stream->base.metrics.receive_end_timestamp_ns = -1;
+    stream->base.metrics.receiving_duration_ns = -1;
```
waahm7, line 369: [debatable] an init function might be better.

source/h2_stream.c
```diff
@@ -240,10 +241,13 @@ struct aws_h2_stream *aws_h2_stream_new_request(
     stream->base.on_incoming_headers = options->on_response_headers;
     stream->base.on_incoming_header_block_done = options->on_response_header_block_done;
     stream->base.on_incoming_body = options->on_response_body;
+    stream->base.on_metrics = options->on_metrics;
     stream->base.on_complete = options->on_complete;
     stream->base.on_destroy = options->on_destroy;
     stream->base.client_data = &stream->base.client_or_server_data.client;
     stream->base.client_data->response_status = AWS_HTTP_STATUS_CODE_UNKNOWN;
+    stream->base.metrics.receiving_duration_ns = -1;
+    stream->base.metrics.sending_duration_ns = -1;
```
waahm7, line 250: I think we need to update here as well.

source/h2_stream.c
```diff
@@ -446,6 +450,20 @@ void aws_h2_stream_complete(struct aws_h2_stream *stream, int error_code) {
     s_h2_stream_destroy_pending_writes(stream);
 
     /* Invoke callback */
+    if (stream->base.on_metrics) {
+        if (stream->base.metrics.send_end_timestamp_ns == 0 && stream->base.metrics.send_start_timestamp_ns != 0) {
```
waahm7, line 454: Should be on `-1` now?

source/h1_connection.c
waahm7, line 742: All asserts should be updated to use `-1` instead of `0`.

source/h2_stream.c
waahm7: Just an idea (not sure good or bad), we can also have a const variable with the default value so that we do not need to update it everywhere manually.

include/aws/http/private/h2_connection.h
```diff
@@ -125,6 +125,13 @@ struct aws_h2_connection {
         uint64_t outgoing_timestamp_ns;
         /* Timestamp when connection has data to receive, which is when there is an active stream */
         uint64_t incoming_timestamp_ns;
+
+        /**
+         * Contains struct aws_http_stream *
+         * The streams that encoded the last frame between it's encoded to the `aws_io_message` to the
+         *`aws_io_message.on_completion` invoked.
+         **/
+        struct aws_array_list finish_encoding_streams_list;
```
graebm, line 134: IMHO, the extra precision doesn't seem worth all the complexity and potential bugs

graebm: Like, maybe we just wrap the input_stream, first call to read() sets start time, and when get_status() says it's done, that's the end time

graebm: knowing the write completion callbacks is ... nice. But the complexity and weirdness they introduce make it so scary

TingDaoK: Get back to the simpler solution to track the time that stream thinks it starts/ends.

include/aws/http/request_response.h
```diff
@@ -194,6 +194,46 @@ typedef void(aws_http_on_stream_complete_fn)(struct aws_http_stream *stream, int
...
+    int64_t sending_duration_ns;
+
+    /* The time stamp when the response started to be received from the network channel. -1 means data not available.
+     * Timestamp are from `aws_high_res_clock_get_ticks` */
+    int64_t receive_start_timestamp_ns;
+    /* The time stamp when the response finished to be received from the network channel. -1 means data not available.
+     * Timestamp are from `aws_high_res_clock_get_ticks` */
+    int64_t receive_end_timestamp_ns;
+    /* The time duration for the request from start sending to finish sending. receive_end_timestamp_ns -
+     * receive_start_timestamp_ns. -1 means data not available. */
```
waahm7, line 219:
```suggestion
    /* The time duration for the request from start receiving to finish receiving. receive_end_timestamp_ns -
     * receive_start_timestamp_ns. -1 means data not available. */
```

source/h1_connection.c
```diff
@@ -1270,6 +1294,13 @@ static int s_decoder_on_done(void *user_data) {
 
     /* Otherwise the incoming stream is finished decoding and we will update it if needed */
     incoming_stream->is_incoming_message_done = true;
+    aws_high_res_clock_get_ticks((uint64_t *)&incoming_stream->base.metrics.receive_end_timestamp_ns);
+    AWS_ASSERT(incoming_stream->base.metrics.receive_start_timestamp_ns != 0);
```
waahm7, line 1298:
```suggestion
    AWS_ASSERT(incoming_stream->base.metrics.receive_start_timestamp_ns != -1);
```

source/h2_stream.c
```diff
@@ -812,6 +828,11 @@ int aws_h2_stream_encode_data_frame(
             stream->thread_data.state = AWS_H2_STREAM_STATE_HALF_CLOSED_LOCAL;
             AWS_H2_STREAM_LOG(TRACE, stream, "Sent END_STREAM. State -> HALF_CLOSED_LOCAL");
         }
+
+        AWS_ASSERT(stream->base.metrics.send_end_timestamp_ns == -1);
+        aws_high_res_clock_get_ticks((uint64_t *)&stream->base.metrics.send_end_timestamp_ns);
+        stream->base.metrics.sending_duration_ns =
+            stream->base.metrics.send_end_timestamp_ns - stream->base.metrics.send_start_timestamp_ns;
```
waahm7, line 835: Should we move this to start of the if block like before so that it is set in the error case as well?

waahm7 (approved):
Good job!

source/request_response.c
```diff
@@ -1107,6 +1107,17 @@ int aws_http_stream_send_response(struct aws_http_stream *stream, struct aws_htt
     return stream->owning_connection->vtable->stream_send_response(stream, response);
 }
 
+struct aws_http_stream *aws_http_stream_acquire(struct aws_http_stream *stream) {
+    if (!stream) {
+        return stream;
+    }
```
graebm, line 1113:
```suggestion
    AWS_ASSERT(stream);
```

Only clean_up/destroy/release() functions protect against passing in null-ish data

acquire() functions should not allow NULL

tests/stream_test_helper.c
graebm: You could still AWS_FATAL_ASSERT some things:
- that stream_id is set
- if _duration is set, then _end is set
- if _end is set, then _begin is set

source/h2_stream.c
```diff
@@ -798,6 +815,11 @@ int aws_h2_stream_encode_data_frame(
      */
     if (input_stream_complete && ends_stream) {
         /* Done sending data. No more data will be sent. */
+        AWS_ASSERT(stream->base.metrics.send_end_timestamp_ns == -1);
+        aws_high_res_clock_get_ticks((uint64_t *)&stream->base.metrics.send_end_timestamp_ns);
```
graebm, line 819: yay, I love that the metrics code now lives with existing state-transition code

graebm (approved):
fix & ship

.github/workflows/ci.yml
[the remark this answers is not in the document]
DmitriyMusatkin: are we are not losing coverage by switching all local host to debug? dont want to end up with the case where the bug only occurs in release build and localtest would have caught it if it was running release

DmitriyMusatkin (approved):

PR #439, opened 2023-04-30
Use promises in elasticurl

evanmiller:
*Issue #, if available:*

*Description of changes:*

[`promise.h`](https://github.com/awslabs/aws-c-common/blob/main/include/aws/common/promise.h) neatly encapsulates what the mutex and condition variable are doing in the `elasticurl` code. I found the code to be more succinct and comprehensible by using it.

Tested locally with the success and failure code paths.

### DIFF
--- a/bin/elasticurl/main.c
+++ b/bin/elasticurl/main.c
@@ -6,12 +6,11 @@
 #include <aws/http/request_response.h>
 
 #include <aws/common/command_line_parser.h>
-#include <aws/common/condition_variable.h>
 #include <aws/common/hash_table.h>
 #include <aws/common/log_channel.h>
 #include <aws/common/log_formatter.h>
 #include <aws/common/log_writer.h>
-#include <aws/common/mutex.h>
+#include <aws/common/promise.h>
 #include <aws/common/string.h>
 
 #include <aws/io/channel_bootstrap.h>
@@ -37,8 +36,7 @@ struct elasticurl_ctx {
     struct aws_allocator *allocator;
     const char *verb;
     struct aws_uri uri;
-    struct aws_mutex mutex;
-    struct aws_condition_variable c_var;
+    struct aws_promise *promise;
     bool response_code_written;
     const char *cacert;
     const char *capath;
@@ -63,7 +61,6 @@ struct elasticurl_ctx {
     const char *trace_file;
     enum aws_log_level log_level;
     enum aws_http_version required_http_version;
-    bool exchange_completed;
 };
 
 static void s_usage(int exit_code) {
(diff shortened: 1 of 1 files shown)


TingDaoK (commented):
The code looks good, but tsan finds a race condition. It's our fault, we need to fix the issue. Will prioritize that based on our tasks.

graebm:
Solid change, but promise.h hasn't been used in any public code yet. I was actually considering removing it from the codebase. Less code is best code, its unit test occasionally fail, and since it only offers a blocking API, it's not useful to building our non-blocking libraries. It's only handy in unit test and samples like this.

Sorry for being a jerk. I'm going to let this PR hang in limbo until the fate of promise.h is decided (this month, likely). We'll remove it, fix it, or replace it with something similar.

evanmiller:
Thank you for the swift attention to my patch.

Based on my reading of [this SO thread](https://stackoverflow.com/questions/47308494/is-there-a-data-race-if-pthread-cond-signal-is-the-last-call-in-a-thread) and a [similar data race here](https://github.com/Kistler-Group/sdbus-cpp/issues/108), it seems that `aws_promise_complete` and `aws_promise_fail` should perform their CV notifications while still holding the mutex rather than immediately after. If there is a spurious wakeup in the main thread while the worker thread is between these two lines

https://github.com/awslabs/aws-c-common/blob/9848a8cf3979c9a820027f705ca985091885afc0/source/promise.c#L85-L86

then the main thread may proceed to destroy the condition variable before the worker can execute its notification. According to this theory, swapping those two lines of code (as well as the last two lines of `aws_promise_fail`) will fix the race condition.

evanmiller:
As to the question of whether to keep `promise.h`: The race condition discovered here would also have presented itself if the original `elasticurl` code had attempted to clean up its condition variable. Mutexes and condition variable logic is hard to get right; as long as the library provides mutex and cvar primitives, my vote would be to keep (and fix) promises in order to prevent similar data race conditions in client code.

bretambrose:
> > Just a minor nitpick, but the current elasticurl implementation does not have a race condition. Predicate-based waits do not suffer from problems when a spurious wakeup occurs.
> 
> I must be misunderstanding what a spurious wakeup entails. My mental model for the data race is:
> 
> Worker sets `exchange_completed` to true Worker unlocks mutex Main thread spuriously wakes up (returns from `pthread_cond_wait`) Main thread checks predicate condition – condition is true Main thread returns from `aws_condition_variable_wait_pred` Main thread continues Worker performs notify
> 
> In my current understanding, there no data race only because `elasticurl` leaks the condition variable resource rather than freeing it. I would appreciate any clarification though.

Invoking the library cleanup will also invoke a join function (https://github.com/awslabs/aws-c-common/blob/main/include/aws/common/thread.h#L169-L178) that does not return until all threads spawned by any of the aws-c-* libraries are fully joined.

Now if you cleaned up the condition variable before that happened, then yes, you could theoretically destroy it while it was still in use.

evanmiller:
With https://github.com/awslabs/aws-c-common/pull/1023 merged, it would be great if someone could re-run the [failed sanitizer test](https://github.com/awslabs/aws-c-http/actions/runs/4853241741/jobs/8649179498?pr=439).

waahm7:
Thank you for the PR, sorry; we have decided to remove the promise.h class, so I am closing this PR as well for now.

PR #440, opened 2023-05-09
Fix warnings in public headers

waahm7:
*Description of changes:*
- Adds `AWS_PUSH_SANE_WARNING_LEVEL` and `AWS_POP_SANE_WARNING_LEVEL` to the public headers. The public header should contain at least one include; otherwise, we will receive a macro not found error. Therefore, I decided to skip adding these macros if there are no includes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,8 @@ on:
       - 'main'
 
 env:
-  BUILDER_VERSION: v0.9.40
-  BUILDER_SOURCE: releases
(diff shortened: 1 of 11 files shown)


TingDaoK (approved):
Looks good, but we need to update the builder version [here](https://github.com/awslabs/aws-c-http/blob/main/codebuild/linux-integration-tests.yml#L13) as well.

Or we can do the following:
https://github.com/awslabs/aws-crt-java/blob/main/codebuild/cd/generic-unix-build.sh#L28-L30

PR #446, opened 2023-07-06
skip empty string for proxy env var

TingDaoK:
*Issue #, if available:*

- We previously try to parse the env var for proxy even if it's empty
- I did some quick search, people think it's unexpected https://dev.gnupg.org/T685
- https://github.com/awslabs/aws-crt-java/issues/644

*Description of changes:*
- If the env var is set to empty, ignore it as it's unset.


graebm (approved):
wonder if we make this mistake in any other aws-c libraries?

PR #451, opened 2023-08-23
Improve Monitoring Report Interval for low allowable_throughput_failure_interval_seconds

waahm7:
*Description of changes:*
The report interval was 1 second by default, which can be slow if `allowable_throughput_failure_interval_seconds` is also low like `1 second` in the [IMDS Client](https://github.com/awslabs/aws-c-auth/blob/main/source/aws_imds_client.c#L168). It took around 3 seconds to figure out if the connection was stuck. This change improves that time to around `1.5 seconds`.

### DIFF
--- a/source/connection_monitor.c
+++ b/source/connection_monitor.c
@@ -188,8 +188,10 @@ static void s_destroy(struct aws_crt_statistics_handler *handler) {
 
 static uint64_t s_get_report_interval_ms(struct aws_crt_statistics_handler *handler) {
     (void)handler;
-
-    return 1000;
+    struct aws_statistics_handler_http_connection_monitor_impl *impl = handler->impl;
+    uint64_t allowable_throughput_failure_interval_ms = aws_timestamp_convert(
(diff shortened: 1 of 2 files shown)


graebm (approved):
We talked offline and you gave the stats of:
>/3 = ~1.6 seconds
> /5 = ~1.4 seconds
> /10 = ~1.2 seconds

maybe we should go for `/10`? I mean ... being 20% off still isn't fantastic, and if you're setting the timeout to 1sec, you obviously want that timeout to be fast

TingDaoK (commented):
Oh, we are using `minimum_throughput_bytes_per_second`.

So that if we are sampling the data less than 1 secs, we will have a inaccurate rate per second :(

The failing test has
```
    struct aws_http_connection_monitoring_options monitor_opt = {
        .allowable_throughput_failure_interval_seconds = 1,
        .minimum_throughput_bytes_per_second = 1000,
    };
```

The test failing because is doing a 900 Byte transfer after a second, so as we sampling the data every 300 ms, there will be two sampling result in 0 B/s, but one sampling results in 3000 B/s.

And the test allows 1 secs to failure. So, in 1 secs, there always be one sampling is larger than the minimal, so that it never shutdown the connection as expected. 

:(

PR #456, opened 2023-10-24
request idle timeout

TingDaoK:
*Issue #, if available:*

- introduce the idle timeout for request to track the number when request finish sending, and server haven't response in a certain time

*Description of changes:*

- Schedule a timeout task to the connection eventloop to kill the connection after the time
- If the response received, we can safely cancel the timeout task as it's the same thread.


include/aws/http/connection.h
```diff
@@ -407,6 +407,16 @@ struct aws_http_client_connection_options {
...
     const struct aws_host_resolution_config *host_resolution_config;
+
+    /**
+     * Optional.
+     * MUST be larger than 0.
+     * If the time between the request finishes to be sent and the first byte of the response received larger than the
+     * set timeout, connection will be closed.
+     * Applies to all the request made from the connection.
+     * TODO: Only supported in HTTP/1.1 now, support it in HTTP/2
+     */
```
graebm, line 418:
```suggestion
    /**
     * Optional (ignored if 0).
     * After a request is fully sent, if the server does not begin responding within N milliseconds,
     * then fail with AWS_ERROR_HTTP_REQUEST_IDLE_TIMEOUT.
     * This can be overridden per-request by aws_http_make_request_options.idle_request_timeout.
     * TODO: Only supported in HTTP/1.1 now, support it in HTTP/2
     */
```

include/aws/http/connection.h
```diff
@@ -407,6 +407,16 @@ struct aws_http_client_connection_options {
...
+    uint64_t idle_timeout_ms;
```
graebm, line 419: I get that it's technically safest to put this at the end of the options-struct, but consider adding it near the `monitoring_options` since they're thematically similar

include/aws/http/http.h
```diff
@@ -59,6 +59,7 @@ enum aws_http_errors {
     AWS_ERROR_HTTP_WEBSOCKET_PROTOCOL_ERROR,
     AWS_ERROR_HTTP_MANUAL_WRITE_NOT_ENABLED,
     AWS_ERROR_HTTP_MANUAL_WRITE_HAS_COMPLETED,
+    AWS_ERROR_HTTP_REQUEST_IDLE_TIMEOUT,
```
graebm, line 62: ugh, I'm super on the fence about this name.
we don't have any other errors that refer to "REQUEST"
and it's not the "request" that times out, it's the "response"

I get that people often refer to the whole operation as "request" (when "stream" or "exchange" would be more accurate)

`AWS_ERROR_HTTP_RESPONSE_TIMEOUT`?

source/h1_connection.c
```diff
@@ -721,6 +722,74 @@ static void s_client_update_incoming_stream_ptr(struct aws_h1_connection *connec
...
 }
 
+static void s_http_request_idle_timeout_task(struct aws_task *task, void *arg, enum aws_task_status status) {
+    struct aws_h1_stream *stream = arg;
+    if (status == AWS_TASK_STATUS_CANCELED) {
+        goto cleanup;
+    }
+    struct aws_http_connection *connection_base = stream->base.owning_connection;
+    struct aws_h1_connection *connection = AWS_CONTAINER_OF(connection_base, struct aws_h1_connection, base);
+    /* Time out happened, close the connection manually */
```
graebm, line 732: utterly trivial:
```suggestion
    /* Timeout happened, close the connection */
```

source/h1_connection.c
```diff
@@ -721,6 +722,74 @@ static void s_client_update_incoming_stream_ptr(struct aws_h1_connection *connec
...
+    s_stop(
+        connection,
+        false /*stop_reading*/,
+        false /*stop_writing*/,
+        true /*schedule_shutdown*/,
+        AWS_ERROR_HTTP_REQUEST_IDLE_TIMEOUT);
+    connection_base->client_data->request_idle_timeout_task = NULL;
+
+cleanup:
+    aws_mem_release(stream->base.alloc, task);
```
graebm, line 753: trivial: don't need to allocate this, store the task by-value
use a separate bool, or check whether task.task_fn is NULL, to check whether it's been scheduled

source/h1_connection.c
```diff
@@ -721,6 +722,74 @@ static void s_client_update_incoming_stream_ptr(struct aws_h1_connection *connec
...
+            connection->client_data->request_idle_timeout_task =
+                aws_mem_calloc(stream->base.alloc, 1, sizeof(struct aws_task));
+            aws_task_init(
+                connection->client_data->request_idle_timeout_task,
+                s_http_request_idle_timeout_task,
+                stream,
+                "http_request_idle_timeout_task");
+            uint64_t now_ns = 0;
+            int error = aws_channel_current_clock_time(channel, &now_ns);
+            AWS_FATAL_ASSERT(!error);
```
graebm, line 783: I mean, it would suck if the clock failed, but an HTTP connection closing early is better than an application crashing
```suggestion
```

source/h1_connection.c
```diff
@@ -721,6 +722,74 @@ static void s_client_update_incoming_stream_ptr(struct aws_h1_connection *connec
...
+            struct aws_event_loop *connection_loop = aws_channel_get_event_loop(channel);
+            aws_event_loop_schedule_task_future(
```
graebm, line 785: use an aws_channel_task, not an event-loop task

the channel will ensure all channel-tasks have been cancelled before the channel destroys itself/slots/handlers

whereas event-loop tasks will give you all kind of lifetime problems

source/h1_connection.c
```diff
@@ -735,13 +804,7 @@ static struct aws_h1_stream *s_update_outgoing_stream_ptr(struct aws_h1_connecti
...
     /* If current stream is done sending data... */
     if (current && !aws_h1_encoder_is_message_in_progress(&connection->thread_data.encoder)) {
-        current->is_outgoing_message_done = true;
-        AWS_ASSERT(current->base.metrics.send_end_timestamp_ns == -1);
-        aws_high_res_clock_get_ticks((uint64_t *)&current->base.metrics.send_end_timestamp_ns);
-        AWS_ASSERT(current->base.metrics.send_start_timestamp_ns != -1);
-        AWS_ASSERT(current->base.metrics.send_end_timestamp_ns >= current->base.metrics.send_start_timestamp_ns);
-        current->base.metrics.sending_duration_ns =
-            current->base.metrics.send_end_timestamp_ns - current->base.metrics.send_start_timestamp_ns;
+        s_request_finished_sending(current);
```
graebm, line 807: unfortunately, there's still a server implementation in here

it's not necessarily a "request" that just finished sending

```suggestion
        s_set_outgoing_message_done(current);
```

source/h1_connection.c
```diff
@@ -721,6 +722,74 @@ static void s_client_update_incoming_stream_ptr(struct aws_h1_connection *connec
...
+    stream->is_outgoing_message_done = true;
+    AWS_ASSERT(stream->base.metrics.send_end_timestamp_ns == -1);
+    aws_high_res_clock_get_ticks((uint64_t *)&stream->base.metrics.send_end_timestamp_ns);
+    AWS_ASSERT(stream->base.metrics.send_start_timestamp_ns != -1);
+    AWS_ASSERT(stream->base.metrics.send_end_timestamp_ns >= stream->base.metrics.send_start_timestamp_ns);
+    stream->base.metrics.sending_duration_ns =
+        stream->base.metrics.send_end_timestamp_ns - stream->base.metrics.send_start_timestamp_ns;
+    if (stream->base.metrics.receive_start_timestamp_ns == -1) {
+        /* We haven't receive any message, schedule the request idle timeout task */
+        uint64_t idle_timeout_ms = stream->base.client_data->idle_timeout_ms == 0
```
graebm, line 770: client_data will be NULL if this is a server

source/h1_connection.c
```diff
@@ -721,6 +722,74 @@ static void s_client_update_incoming_stream_ptr(struct aws_h1_connection *connec
...
+cleanup:
+    aws_mem_release(stream->base.alloc, task);
+}
+
+static void s_request_finished_sending(struct aws_h1_stream *stream) {
+    struct aws_http_connection *connection = stream->base.owning_connection;
+    struct aws_channel *channel = aws_http_connection_get_channel(connection);
+    AWS_ASSERT(aws_channel_thread_is_callers_thread(channel));
+
+    stream->is_outgoing_message_done = true;
```
graebm, line 761: In the current flow, I'm not 100% sure whether this function might end up being called twice.
probably want to check whether `is_outgoing_message_done` is already true

source/h1_connection.c
TingDaoK: channel task doesn't have a cancel, although I can still access the wrapper task to use the eventloop cancel

source/h1_connection.c
```diff
@@ -721,6 +731,82 @@ static void s_client_update_incoming_stream_ptr(struct aws_h1_connection *connec
...
+                                   ? connection_base->client_data->idle_timeout_ms
+                                   : stream->base.client_data->idle_timeout_ms;
+    AWS_LOGF_INFO(
+        AWS_LS_HTTP_CONNECTION,
+        "id=%p: Closing connection as timeout after request sent to the first byte received happened. Idle time "
+        "is %" PRIu64 " ms",
+        (void *)connection_base,
+        idle_timeout_ms);
+
+    AWS_ZERO_STRUCT(stream->base.client_data->idle_timeout_task);
```
graebm, line 755: trivial: maybe just zero this at the top of the function, instead of doing it from 2 different places

/* zero-out task to indicate that it's no longer scheduled */

include/aws/http/connection.h
```diff
@@ -307,6 +307,15 @@ struct aws_http_client_connection_options {
...
     const struct aws_http_connection_monitoring_options *monitoring_options;
 
+    /**
+     * Optional (ignored if 0).
+     * After a request is fully sent, if the server does not begin responding within N milliseconds,
+     * then fail with AWS_ERROR_HTTP_RESPONSE_TIMEOUT.
+     * This can be overridden per-request by aws_http_make_request_options.idle_request_timeout.
+     * TODO: Only supported in HTTP/1.1 now, support it in HTTP/2
+     */
+    uint64_t idle_timeout_ms;
```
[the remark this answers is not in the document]
TingDaoK: discussed, sold on `response_first_byte_timeout_ms`

PR #458, opened 2023-12-14
Add support for cancel stream

TingDaoK:
- Add support for cancel stream
- A bit nicer interface for you to provide an error code, that will result in the complete callback
- Basically doing the same thing as close connection for H1 and reset stream for H2


source/h1_connection.c
```diff
@@ -388,6 +388,36 @@ int aws_h1_stream_activate(struct aws_http_stream *stream) {
...
+            connection->synced_data.is_open == false) {
+            /* Not active, nothing to cancel. */
+            aws_h1_connection_unlock_synced_data(connection);
+            AWS_LOGF_DEBUG(AWS_LS_HTTP_STREAM, "id=%p: Stream not active, nothing to cancel.", (void *)stream);
+            return;
+        }
+        connection->synced_data.is_open = false;
+        connection->synced_data.new_stream_error_code = AWS_ERROR_HTTP_CONNECTION_CLOSED;
+
+        aws_h1_connection_unlock_synced_data(connection);
```
graebm, line 408: debatable: I'd prefer we release the lock and call s_stop() (which yes I know grabs the lock again) but reduces code duplication

source/h2_stream.c
```diff
@@ -586,6 +588,46 @@ static int s_stream_reset_stream(struct aws_http_stream *stream_base, uint32_t h
...
+    bool cross_thread_work_should_schedule = false;
+
+    { /* BEGIN CRITICAL SECTION */
+        s_lock_synced_data(stream);
+
+        reset_called = stream->synced_data.reset_called;
+        stream_is_active = stream->synced_data.api_state == AWS_H2_STREAM_API_STATE_ACTIVE;
+        if (!reset_called && stream_is_active) {
+            cross_thread_work_should_schedule = !stream->synced_data.is_cross_thread_work_task_scheduled;
+            stream->synced_data.reset_called = true;
```
graebm, line 610: this looks a lot like the implementation of s_stream_reset_stream_internal(). Why not just call that? If there's some difference, can we just modify that function? it's internal

source/h1_connection.c
```diff
@@ -388,6 +388,34 @@ int aws_h1_stream_activate(struct aws_http_stream *stream) {
     return AWS_OP_SUCCESS;
 }
 
+void aws_h1_stream_cancel(struct aws_http_stream *stream, int error_code) {
```
graebm, line 391: no-op: I was going to say "we should ensure they don't pass in 0", but turns it's fine. If they passed 0:
- the channel will end up shutting down with error_code 0
- BUT if the channel's error_code is 0, the stream will default to [AWS_ERROR_HTTP_CONNECTION_CLOSED](https://github.com/awslabs/aws-c-http/blob/0c965027a5747f427b3ac6b70e528be93ab21be4/source/h1_connection.c#L2081-L2082)

graebm (approved):

PR #462, opened 2024-02-10
Fix a bunch of places we forget to aws_raise_error()

graebm:
See: https://github.com/awslabs/aws-c-common/pull/1089

### DIFF
--- a/source/h1_encoder.c
+++ b/source/h1_encoder.c
@@ -403,10 +403,7 @@ int aws_h1_encoder_message_init_from_response(
         goto error;
     }
 
-    err = aws_byte_buf_init(&message->outgoing_head_buf, allocator, head_total_len);
-    if (err) {
-        return AWS_OP_ERR;
-    }
+    aws_byte_buf_init(&message->outgoing_head_buf, allocator, head_total_len);
 
     bool wrote_all = true;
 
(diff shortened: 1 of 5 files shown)


source/h1_encoder.c
```diff
@@ -403,10 +403,7 @@ int aws_h1_encoder_message_init_from_response(
         goto error;
     }
 
-    err = aws_byte_buf_init(&message->outgoing_head_buf, allocator, head_total_len);
-    if (err) {
-        return AWS_OP_ERR;
-    }
+    aws_byte_buf_init(&message->outgoing_head_buf, allocator, head_total_len);
```
sbSteveK, line 406: It looks like aws_byte_buf_init could return an AWS_OP_ERR resulting in a zeroed out outgoing_head_buf. Just making sure we shouldn't be checking the return and goto error here to clean up message if that happens.

sbSteveK (approved):

waahm7 (approved):

source/h1_encoder.c
graebm: In the past, aws_buf_buf_init() could fail if the allocator failed. But a few years back, we made it impossible for memory allocation to fail (the program will abort before the allocator can return NULL)

PR #329, opened 2021-08-05
Cutomized alpn

TingDaoK:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #474, opened 2024-07-11
adapt change from "TLS deliver buffer data during shutdown"

TingDaoK:
Change the HTTP/1 connection and websocket to adapt the change from TLS deliver the buffered data during shutdown
- In case of user initial the shutdown process, to prevent hanging, open up the flow control window to max to keep the data flow, but as read has stopped, drop anything from TLS directly. 

No need for HTTP/2, since HTTP/2 has it own flow-control method, and not using the channel flow control.

Other unrelated changes:
- We previously disabled the TLS test for Mac because of https://github.com/actions/runner-images/issues/3861
- After GitHub fixed the issue, we brought back one test but forgot the other one https://github.com/awslabs/aws-c-http/pull/329
- Bring back the test, which should be not skipped now.
- MacOS 11 is gone for github runner image. 

TODO:
- HTTP/1 also has the buffer data will met the same issue during shutdown as TLS handler

### DIFF
--- a/source/h1_connection.c
+++ b/source/h1_connection.c
@@ -146,7 +146,7 @@ static void s_stop(
         connection->thread_data.is_reading_stopped = true;
         /* Increase the window size after shutdown starts, to prevent deadlock when data still pending in the TLS
          * handler. */
-        connection->base.channel_slot->window_size = SIZE_MAX;
+        aws_channel_slot_increment_read_window(connection->base.channel_slot, SIZE_MAX);
     }
 
     if (stop_writing) {
--- a/source/h1_connection.c
+++ b/source/h1_connection.c
@@ -207,6 +207,9 @@ static void s_shutdown_from_off_thread(struct aws_h1_connection *connection, int
             connection->synced_data.shutdown_requested = true;
             connection->synced_data.shutdown_requested_error_code = error_code;
         }
+        /* Connection has shutdown, new streams should not be allowed. */
+        connection->synced_data.is_open = false;
+        connection->synced_data.new_stream_error_code = AWS_ERROR_HTTP_CONNECTION_CLOSED;
         aws_h1_connection_unlock_synced_data(connection);
     } /* END CRITICAL SECTION */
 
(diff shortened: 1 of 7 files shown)


source/h1_connection.c
```diff
@@ -785,11 +840,11 @@ static void s_http_stream_response_first_byte_timeout_task(
...
-    /* Don't stop reading/writing immediately, let that happen naturally during the channel shutdown process. */
+    /* Shutdown the connection. */
     s_stop(
         connection,
-        false /*stop_reading*/,
-        false /*stop_writing*/,
+        true /*stop_reading*/,
+        true /*stop_writing*/,
         true /*schedule_shutdown*/,
         AWS_ERROR_HTTP_RESPONSE_FIRST_BYTE_TIMEOUT);
```
graebm, line 849:
```suggestion
    s_shutdown_due_to_error(connection, AWS_ERROR_HTTP_RESPONSE_FIRST_BYTE_TIMEOUT);
```

source/h1_connection.c
```diff
@@ -189,14 +192,52 @@ static void s_shutdown_due_to_error(struct aws_h1_connection *connection, int er
...
  * Public function for closing connection.
  */
 static void s_connection_close(struct aws_http_connection *connection_base) {
     struct aws_h1_connection *connection = AWS_CONTAINER_OF(connection_base, struct aws_h1_connection, base);
+    { /* BEGIN CRITICAL SECTION */
+        aws_h1_connection_lock_synced_data(connection);
+        /* Even if we're not scheduling shutdown just yet (ex: sent final request but waiting to read final response)
+         * we don't consider the connection "open" anymore so user can't create more streams */
+        connection->synced_data.is_open = false;
+        connection->synced_data.new_stream_error_code = AWS_ERROR_HTTP_CONNECTION_CLOSED;
```
graebm, line 235: why are we only doing this from connection_close() but not stream_cancel()?
I would just move this behavior into the common s_shutdown_from_off_thread() function

graebm (approved):

graebm (approved):

PR #479, opened 2024-07-29
Connection Manager Acquisition Timeout

waahm7:
*Description of changes:*
Introduce a new `pending_connections_acquire_timeout_ms` similar to `max_connection_idle_in_milliseconds`. I have changed the `s_schedule_connection_culling` to a generic `s_schedule_culling` which culls both the idle connections and pending acquisitions.


source/connection_manager.c
```diff
@@ -801,9 +794,69 @@ static void s_schedule_connection_culling(struct aws_http_connection_manager *ma
...
+        aws_ref_count_acquire(&manager->internal_ref_count);
+    }
+
+    if (manager->cull_event_loop == NULL) {
+        manager->cull_event_loop = aws_event_loop_group_get_next_loop(manager->bootstrap->event_loop_group);
+    }
+    AWS_FATAL_ASSERT(manager->cull_event_loop != NULL);
+
+    uint64_t cull_task_time = s_calculate_idle_connection_cull_task_time(manager);
+    uint64_t connection_acquire_timeout = s_calculate_pending_connections_acquire_cull_task_time(manager);
```
graebm, line 851: trivial: "time" is always ambiguous. length of time? point in time?

be consistent here, don't call one `_time` and the other `_timeout`. Call them both `_time` or `_timestamp`

source/connection_manager.c
```diff
@@ -801,9 +794,69 @@ static void s_schedule_connection_culling(struct aws_http_connection_manager *ma
...
 
-    return;
+    if (manager->max_connection_idle_in_milliseconds != 0 && manager->pending_connections_acquire_timeout_ms != 0) {
+        cull_task_time = aws_min_u64(cull_task_time, connection_acquire_timeout);
+    } else if (manager->pending_connections_acquire_timeout_ms != 0) {
+        cull_task_time = connection_acquire_timeout;
+    }
```
graebm, line 857: trivial: if you had these helpers default to returning UINT64_MAX, instead of 0, you could make this simpler like. something like:
```
    uint64_t cull_task_time = aws_min_u64(s_calculate_idle_connection_cull_task_time(manager),
                                          s_calculate_pending_connections_acquire_cull_task_time(manager));
```
or
```
uint64_t idle_cull_time = s_calculate_idle_connection_cull_task_time(manager);
uint64_t acquisition_cull_time = s_calculate_pending_connections_acquire_cull_task_time(manager);
uint64_t cull_task_time = aws_min_u64(idle_cull_time, acquisition_cull_time);
```

TingDaoK, line 857: do we need two separate helper function to get the proper time for next cull task?
it seems more complicated than it needs to be.

source/connection_manager.c
```diff
@@ -801,9 +794,69 @@ static void s_schedule_connection_culling(struct aws_http_connection_manager *ma
...
+    uint64_t cull_task_time = 0;
+    aws_mutex_lock(&manager->lock);
+
+    const struct aws_linked_list_node *end = aws_linked_list_end(&manager->pending_acquisitions);
+    struct aws_linked_list_node *oldest_node = aws_linked_list_begin(&manager->pending_acquisitions);
+    if (oldest_node != end) {
+        /*
+         * Since the pending acquires are in FIFO order in the list, the front of the list has the closest
+         * cull time.
+         */
```
TingDaoK, line 814: it's just confusing that the comments above is
```
        /*
         * Since the connections are in LIFO order in the list, the front of the list has the closest
         * cull time.
         */
```
And here, everything is the same but FIFO order to be mentioned...

I think for the acquisition it's not as complicated as the connections. It won't back to the queue after poped. So, it may be simpler to just mention that the front will be the closest cull time.

source/connection_manager.c
```diff
@@ -801,9 +794,69 @@ static void s_schedule_connection_culling(struct aws_http_connection_manager *ma
...
+        struct aws_http_connection_acquisition *oldest_pending_acquire =
+            AWS_CONTAINER_OF(oldest_node, struct aws_http_connection_acquisition, node);
+        cull_task_time = oldest_pending_acquire->timeout_timestamp;
+    } else {
+        /*
+         * There are no connections in the list, so the absolute minimum anything could be culled is the full
```
TingDaoK, line 820:
```suggestion
         * There are no acquisition in the list, so the absolute minimum anything could be culled is the full
```

source/connection_manager.c
TingDaoK, line 851: Keep the naming consistent with the existing naming. acquire -> acquisition

source/connection_manager.c
```diff
@@ -1594,17 +1665,61 @@ static void s_cull_idle_connections(struct aws_http_connection_manager *manager)
...
+    struct aws_connection_management_transaction work;
+    s_aws_connection_management_transaction_init(&work, manager);
+
+    aws_mutex_lock(&manager->lock);
+
+    /* Only if we're not shutting down */
+    if (manager->state == AWS_HCMST_READY) {
+        const struct aws_linked_list_node *end = aws_linked_list_end(&manager->pending_acquisitions);
+        struct aws_linked_list_node *current_node = aws_linked_list_begin(&manager->pending_acquisitions);
+        while (current_node != end) {
```
graebm, line 1689: maybe a bit less copy/pasty to move the cull_acquisitions and cull_idle logic into the same function. just have them both put their actions in the same `work` struct

fewer locks, clocks, and if-we're-not-shutting-down blocks

source/connection_manager.c
waahm7: I started with one function, but it was more confusing since we need to cast the list node to different structs and check a different variable for full cull timestamp. It was the same amount of code duplication. So I split them into two separate functions.

source/connection_manager.c
waahm7: Thanks, updated with the simple version.

source/connection_manager.c
waahm7: Thanks, updated to what Mike suggested.

include/aws/http/connection_manager.h
```diff
@@ -125,6 +125,12 @@ struct aws_http_connection_manager_options {
      */
     uint64_t max_connection_idle_in_milliseconds;
 
+    /**
+     * If set to a non-zero value, then pending connections acquire will fail with
+     * AWS_ERROR_HTTP_CONNECTION_MANAGER_PENDING_ACQUIRE_TIMEOUT error after the specified timeout.
+     */
+    uint64_t pending_connections_acquire_timeout_ms;
```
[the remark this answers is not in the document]
waahm7: Thanks, updated.

source/connection_manager.c
```diff
@@ -801,9 +794,69 @@ static void s_schedule_connection_culling(struct aws_http_connection_manager *ma
...
+    return cull_task_time;
+}
+
+static uint64_t s_calculate_pending_connections_acquire_cull_task_time(struct aws_http_connection_manager *manager) {
+    if (manager->pending_connections_acquire_timeout_ms == 0) {
+        return 0;
+    }
+
+    uint64_t cull_task_time = 0;
+    aws_mutex_lock(&manager->lock);
```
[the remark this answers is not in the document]
waahm7: Thanks, updated to one lock. I did start with one function but IMO it was more confusing so I split them into two functions.

source/connection_manager.c
waahm7: Thanks, updated to _time.

source/connection_manager.c
waahm7: Thanks, updated to return UINT64_MAX.

tests/CMakeLists.txt
```diff
@@ -519,6 +519,7 @@ add_net_test_case(connection_customized_alpn_error_with_unknown_return_string)
 # unit tests where connections are mocked
 add_net_test_case(test_connection_manager_setup_shutdown)
 add_net_test_case(test_connection_manager_acquire_release_mix_synchronous)
+add_net_test_case(s_test_connection_manager_acquisition_timeout)
```
graebm, line 522:
```suggestion
add_net_test_case(test_connection_manager_acquisition_timeout)
```

graebm (approved):

TingDaoK (approved):

PR #483, opened 2024-08-15
Fix the websocket shutdown behavior

TingDaoK:
The bug was introduced in [PR #474](https://github.com/awslabs/aws-c-http/pull/474/files#diff-ee776c7576cfff50a64158d59a6173ab9a0aa373150574aa9987b4f8726b58e3)
  - `is_writing_stopped = true` shouldn't be set directly, there's a helper function `s_stop_writing()` that ensures subsequent calls to `aws_websocket_send_frame()` will fail.

Let's take a whole new approach these channel-shutdown-window-deadlock issues:
- add `s_stop_reading_and_dont_block_shutdown()` function that sets `is_reading_stopped = true`, but also increments the read window so that channel shutdown won't deadlock.
    - Most places that were setting `is_reading_stopped = true` now use this helper instead
- Revamp how `aws_channel_shutdown()` is called. Lots of channel behavior has changed since [this websocket code was written](https://github.com/awslabs/aws-c-http/pull/48).
  - If on the channel-thread, just call `aws_channel_shutdown()`
      - now that [aws_channel_shutdown()](https://github.com/awslabs/aws-c-io/pull/172) is always async, we don't need to defensively schedule a task to call it
  - If off-thread, use `s_schedule_channel_shutdown_from_offthead()`
      - now that this is only called from `aws_websocket_close()`, or when the refcount goes to zero, we can assume the user is OK if reading stops, and it can call `s_stop_reading_and_dont_block_shutdown()` on the way to shutting down.
- Add the test to verify that send after close should fail


sbSteveK (approved):

source/websocket.c
```diff
@@ -989,13 +989,13 @@ static void s_shutdown_channel_task(struct aws_channel_task *task, void *arg, en
 
     s_unlock_synced_data(websocket);
     /* END CRITICAL SECTION */
-    websocket->thread_data.is_reading_stopped = true;
-    websocket->thread_data.is_writing_stopped = true;
```
graebm, line 993: Oh man, I totally didn't notice these changes in that other PR.

I'll take over this PR. I'm oncall

TingDaoK (commented):
Cannot approve, but :shipit:

graebm (approved):

PR #502, opened 2025-02-28
Apple Network Framework Support

xiazhvera:
*Issue #, if available:*

*Description of changes:*
The consequence changes for Apple Network Framework Support. Depends on https://github.com/awslabs/aws-c-io/pull/662

### DIFF
--- a/source/connection.c
+++ b/source/connection.c
@@ -11,8 +11,6 @@
 
 #include <aws/http/private/proxy_impl.h>
 
-#include <aws/io/private/socket_impl.h>
-
 #include <aws/common/condition_variable.h>
 #include <aws/common/hash_table.h>
 #include <aws/common/mutex.h>
@@ -731,9 +729,15 @@ struct aws_http_server *aws_http_server_new(const struct aws_http_server_options
     };
 
     int listen_error = AWS_OP_SUCCESS;
+#if (defined(AWS_ENABLE_DISPATCH_QUEUE) && !defined(AWS_ENABLE_KQUEUE)) || defined(AWS_USE_APPLE_NETWORK_FRAMEWORK)
+#    define AWS_NETWORK_FRAMEWORK_ENABLED 1
+#else
+#    define AWS_NETWORK_FRAMEWORK_ENABLED 0
+#endif
+
     if (bootstrap_options.socket_options->impl_type == AWS_SOCKET_IMPL_APPLE_NETWORK_FRAMEWORK ||
-        (bootstrap_options.socket_options->impl_type == AWS_SOCKET_IMPL_PLATFORM_DEFAULT &&
-         aws_socket_get_default_impl_type() == AWS_SOCKET_IMPL_APPLE_NETWORK_FRAMEWORK)) {
+        (AWS_NETWORK_FRAMEWORK_ENABLED &&
+         bootstrap_options.socket_options->impl_type == AWS_SOCKET_IMPL_PLATFORM_DEFAULT)) {
         /*
          * WARNING!!!!
          * For Apple Network Framework, socket listen is an async function, we would need block here waiting for
@@ -750,12 +754,14 @@ struct aws_http_server *aws_http_server_new(const struct aws_http_server_options
         listen_error = server_user_data->setup_error_code;
     } else {
         server->socket = aws_server_bootstrap_new_socket_listener(&bootstrap_options);
-        listen_error = aws_last_error();
+        if (!server->socket) {
+            listen_error = aws_last_error();
+        }
     }
 
     s_server_unlock_synced_data(server);
 
-    if (!server->socket || listen_error) {
+    if (listen_error) {
         AWS_LOGF_ERROR(
             AWS_LS_HTTP_SERVER,
             "static: Failed creating new socket listener, error %d (%s). Cannot create server.",
--- a/source/connection.c
+++ b/source/connection.c
@@ -11,6 +11,7 @@
 
 #include <aws/http/private/proxy_impl.h>
 
+#include <aws/common/condition_variable.h>
 #include <aws/common/hash_table.h>
 #include <aws/common/mutex.h>
 #include <aws/common/string.h>
@@ -431,6 +432,15 @@ void aws_http_connection_release(struct aws_http_connection *connection) {
     }
 }
 
+struct aws_server_user_data {
+    struct aws_allocator *alloc;
+    struct aws_http_server *server;
+    struct aws_mutex mutex;
+    struct aws_condition_variable condition_variable;
+    int setup_error_code;
+    bool setup_complete;
+};
+
 /* At this point, the server bootstrapper has accepted an incoming connection from a client and set up a channel.
  * Now we need to create an aws_http_connection and insert it into the channel as a channel-handler.
  * Note: Be careful not to access server->socket until lock is acquired to avoid race conditions */
@@ -442,7 +452,8 @@ static void s_server_bootstrap_on_accept_channel_setup(
 
     (void)bootstrap;
     AWS_ASSERT(user_data);
-    struct aws_http_server *server = user_data;
+    struct aws_server_user_data *server_user_data = user_data;
+    struct aws_http_server *server = server_user_data->server;
     bool user_cb_invoked = false;
     struct aws_http_connection *connection = NULL;
     if (error_code) {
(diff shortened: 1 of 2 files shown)


bretambrose (approved):
Assuming Mike is ok with the wait.

source/h1_decoder.c
```diff
@@ -329,6 +329,7 @@ static int s_linestate_chunk_size(struct aws_h1_decoder *decoder, struct aws_byt
 
         /* Expected empty newline and end of message. */
         decoder->doing_trailers = true;
+
```
graebm, line 332: trivial: unnecessary edit
```suggestion
```

source/connection.c
```diff
@@ -431,6 +432,15 @@ void aws_http_connection_release(struct aws_http_connection *connection) {
     }
 }
 
+struct aws_server_user_data {
+    struct aws_allocator *alloc;
+    struct aws_http_server *server;
+    struct aws_mutex mutex;
+    struct aws_condition_variable condition_variable;
+    int setup_error_code;
+    bool setup_complete;
```
graebm, line 441: This is a perfect use case for [aws_future](https://github.com/awslabs/aws-c-io/blob/318f7e57e7871e5b0d48a281cc5dcb7f79ccecdd/include/aws/io/future.h), which uses a mutex, condition variable, error code, and optional "success" value, plus a bunch of helper functions that can do exactly this kind of thing
```suggestion
    struct aws_future_void *setup_future;
```

source/connection.c
```diff
@@ -617,8 +629,30 @@ static void s_server_bootstrap_on_accept_channel_shutdown(
...
+    struct aws_server_bootstrap *bootstrap,
+    int error_code,
+    void *user_data) {
+    (void)bootstrap;
+    struct aws_server_user_data *server_user_data = user_data;
+    aws_mutex_lock(&server_user_data->mutex);
+    server_user_data->setup_error_code = error_code;
+    server_user_data->setup_complete = true;
+    aws_condition_variable_notify_one(&server_user_data->condition_variable);
+    aws_mutex_unlock(&server_user_data->mutex);
```
graebm, line 655:
```suggestion
    if (error_code) {
        aws_future_void_set_error(server_user_data->setup_future, error_code);
    } else {
        aws_future_void_set_result(server_user_data->setup_future);
    }
```

source/connection.c
```diff
@@ -672,6 +706,14 @@ struct aws_http_server *aws_http_server_new(const struct aws_http_server_options
         server->is_using_tls = true;
     }
 
+    struct aws_server_user_data *server_user_data =
+        aws_mem_calloc(options->allocator, 1, sizeof(struct aws_server_user_data));
+    server_user_data->server = server;
+    server_user_data->alloc = options->allocator;
+    aws_mutex_init(&server_user_data->mutex);
+    aws_condition_variable_init(&server_user_data->condition_variable);
+    server_user_data->setup_complete = false;
```
graebm, line 715:
```suggestion
    server_user_data->setup_future = aws_future_void_new(options->allocator);
```

source/connection.c
```diff
@@ -680,21 +722,41 @@ struct aws_http_server *aws_http_server_new(const struct aws_http_server_options
...
+    // if server setup properly, waiting for setup callback
+    if (server->socket) {
+        aws_mutex_lock(&server_user_data->mutex);
+        aws_condition_variable_wait_pred(
+            &server_user_data->condition_variable,
+            &server_user_data->mutex,
+            s_listener_connected_predicate,
+            server_user_data);
+        aws_mutex_unlock(&server_user_data->mutex);
+        listen_error = server_user_data->setup_error_code;
```
graebm, line 748:
```suggestion
        aws_future_void_wait(server_user_data->setup_future, UINT64_MAX);
        listen_error = aws_future_void_get_error(server_user_data->setup_future);
```

source/connection.c
```diff
@@ -617,8 +629,30 @@ static void s_server_bootstrap_on_accept_channel_shutdown(
 static void s_server_bootstrap_on_server_listener_destroy(struct aws_server_bootstrap *bootstrap, void *user_data) {
     (void)bootstrap;
     AWS_ASSERT(user_data);
-    struct aws_http_server *server = user_data;
+    struct aws_server_user_data *server_user_data = user_data;
+    struct aws_http_server *server = server_user_data->server;
     s_http_server_clean_up(server);
+    aws_mem_release(server_user_data->alloc, server_user_data);
```
graebm, line 635: You're probably leaking `server_user_data` if this fails in the async-setup phase.

Also, forgot to call aws_mutex_clean_up() and aws_condition_variable_clean_up(), but now need to call aws_future_void_destroy(). Probably want to do this cleanup in a helper function ... oh hey there's already an s_http_server_clean_up() function. Probably should alter it to take `aws_server_user_data  *` instead of `aws_http_server *`, since the server always lives inside the user-data struct now

source/connection.c
```diff
@@ -579,6 +587,12 @@ static void s_http_server_clean_up(struct aws_http_server *server) {
     aws_mem_release(server->alloc, server);
 }
 
+static void s_http_server_user_data_clean_up(struct aws_server_user_data *server_user_data) {
```
graebm, line 590: trivial/style: combine this function with the existing s_http_server_clean_up() so there's only 1 clean-up function. 

Cleaner to have 1 clean-up function, than 2 clean-up functions where someone might accidentally call the wrong one

source/connection.c
```diff
@@ -672,6 +701,12 @@ struct aws_http_server *aws_http_server_new(const struct aws_http_server_options
         server->is_using_tls = true;
     }
 
+    struct aws_server_user_data *server_user_data =
+        aws_mem_calloc(options->allocator, 1, sizeof(struct aws_server_user_data));
+    server_user_data->server = server;
+    server_user_data->alloc = options->allocator;
+    server_user_data->setup_future = aws_future_void_new(options->allocator);
```
graebm, line 708: wait a minute ... what is even the point of the `struct server_user_data`? Can't we just put the `setup_future` inside the `struct server`

EZPZ

source/connection.c
```diff
@@ -709,6 +758,8 @@ struct aws_http_server *aws_http_server_new(const struct aws_http_server_options
     return server;
 
 socket_error:
+    aws_future_void_release(server_user_data->setup_future);
```
graebm, line 761: oh man ... this is old-fashioned style CRT error handling, where we'd have different goto labels for each tiny piece that could go wrong. And then try to selectively clean up ONLY the struct members that were successfully initialized.

Please rewrite this to just be 1 goto label that always calls `s_http_server_clean_up()`.
All our misc release() and destroy() and clean_up() functions tolerate being called on something that was never fully initialized. So it's fine to just call `s_http_server_clean_up()`, and have it naively clean up the mutex and hash table even if they'd failed initialization. It's simpler, and safer than having 2 different places in code that clean up the struct members. As as example of why it's better, observe how these goto labels forgot to clean up some socket members like `aws_server_bootstrap_release(server->bootstrap)`

xiazhvera: Did not see a better way to determine if we should clean up the mutex, therefore added a flag to indicate if the mutex is inited or not.

source/connection.c
```diff
@@ -680,23 +700,37 @@ struct aws_http_server *aws_http_server_new(const struct aws_http_server_options
...
+    /*
+     * WARNING & TODO!!!!
+     * aws_server_bootstrap_new_socket_listener has async callback, we would block here waiting for
+     * setup complete. aws-c-http library need to be updated with a proper async API.
+     */
+    server->socket = aws_server_bootstrap_new_socket_listener(&bootstrap_options);
+    // if server setup properly, waiting for setup callback
+    if (server->socket) {
+        aws_future_void_wait(server->setup_future, UINT64_MAX /*timeout*/);
+        listen_error = aws_future_void_get_error(server->setup_future);
```
graebm, line 720: in this path, `aws_raise_error()` is never called, so the aws_http_server_new() is going to return NULL, but if the user calls aws_last_error() to find out why they'll get 0 or whatever stale old error was on the stack

source/connection.c
```diff
@@ -680,23 +699,37 @@ struct aws_http_server *aws_http_server_new(const struct aws_http_server_options
...
         AWS_LOGF_ERROR(
             AWS_LS_HTTP_SERVER,
             "static: Failed creating new socket listener, error %d (%s). Cannot create server.",
-            aws_last_error(),
-            aws_error_name(aws_last_error()));
+            listen_error,
+            aws_error_name(listen_error));
 
-        goto socket_error;
+        goto server_error;
```
graebm, line 732:
```suggestion
        aws_raise_error(listen_error);
        goto server_error;
```

graebm (approved):
fix & ship

PR #506, opened 2025-03-11
HTTP/1: Support streaming requests of unknown length

graebm:
**Issue:**

aws-sdk-cpp needs to send streaming requests of unknown content length. This isn't currently supported in our HTTP/1 client. (our HTTP/2 client *does* already support this)

**Background:**

The HTTP/1 client already supported these use cases:
- Send streaming request body, but declare Content-Length up front
- Send chunked request body, but user must declare length of each chunk via `write_chunk()` API.

But it wasn't possible to send a stream of unknown length without doing extra copies (read to intermediate buffer, then send that buffer via write_chunk() API).

**Description of changes:**

Support input streams of unknown length via chunked encoding.

When reading from the stream, leave a bit of space at the start and end of the aws_io_message buffer for the chunk prefix and suffix. Then go back and write in the prefix and suffix afterwards. No unnecessary copies are involved.


source/h1_encoder.c
```diff
@@ -755,6 +750,134 @@ static int s_state_fn_unchunked_body(struct aws_h1_encoder *encoder, struct aws_
...
+        return AWS_OP_SUCCESS;
+    }
+
+    if (aws_byte_buf_write_from_whole_cursor(dst, last_chunk) == true) {
+        ENCODER_LOG(TRACE, encoder, "Last chunk complete");
+        return s_switch_state(encoder, AWS_H1_ENCODER_STATE_CHUNK_TRAILER);
+    } else {
+        /* Remain in state until there's enough space to write */
+        return AWS_OP_SUCCESS;
+    }
```
TingDaoK, line 878: don't we just checked the space couple line before? should we just assert on the return val?

TingDaoK (approved):
minor comments

source/h1_encoder.c
```diff
@@ -755,6 +750,134 @@ static int s_state_fn_unchunked_body(struct aws_h1_encoder *encoder, struct aws_
...
+        /* Write chunk-prefix: LENGTH-IN-HEX CRLF */
+        char hexbuf[padded_hex_len + 1] = {0};
+        snprintf(hexbuf, sizeof(hexbuf), padded_hex_fmt, body_sub_buf.len);
+        aws_byte_buf_write_from_whole_cursor(dst, aws_byte_cursor_from_c_str(hexbuf));
+        s_write_crlf(dst);
+
+        /* Increment dst->len, since we already copied body in there via sub-buffer */
+
+        dst->len += body_sub_buf.len;
+        AWS_ASSERT(dst->len <= dst->capacity);
```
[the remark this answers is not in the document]
graebm: I added mildly different asserts

rather than assert I'm going to have enough space, just assert afterwards that all the aws_byte_buf_write() calls succeeded via the old  `wrote_all &= aws_byte_buf_write_fn_that_returns_bool(...); AWS_ASSERT(wrote_all)` trick

source/h1_encoder.c
```diff
@@ -755,6 +750,134 @@ static int s_state_fn_unchunked_body(struct aws_h1_encoder *encoder, struct aws_
...
+     * (else no body data streamed, so dst remains untouched. Maybe we've
+     * reached end of stream, maybe user just doesn't have data yet to send) */
+    if (body_sub_buf.len > 0) {
+        encoder->chunk_count++;
+        ENCODER_LOGF(
+            TRACE, encoder, "Sending chunk #%" PRIu64 " with size %zu", encoder->chunk_count, body_sub_buf.len);
+
+        /* Write chunk-prefix: LENGTH-IN-HEX CRLF */
+        char hexbuf[padded_hex_len + 1] = {0};
+        snprintf(hexbuf, sizeof(hexbuf), padded_hex_fmt, body_sub_buf.len);
```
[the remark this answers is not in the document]
graebm: it's impossible due to math and clamping we did earlier
added AWS_ASSERT and comment

`AWS_ASSERT(body_sub_buf.len <= max_hex_value_given_padding); /* guaranteed, b/c we clamped .capacity earlier */`

source/h1_encoder.c
graebm: this is a the `CHUNKED_BODY_STREAM_LAST_CHUNK` state, that we go into after aws_input_stream has reported EOF

so like imagine:
-  in the `AWS_H1_ENCODER_STATE_CHUNKED_BODY_STREAM` state, the aws_input_stream nearly fills the aws_io_message, there are only 2 bytes left over. The state checks if the stream is EOF, and it is, so the state machine moves to the next state...
- This is the AWS_H1_ENCODER_STATE_CHUNKED_BODY_STREAM_LAST_CHUNK, which we're looking at. It want to write "0\r\n", it needs 3 bytes, so we `/* Remain in state until there's enough space to write */`
- The h1_encoder state machine sees that the state didn't change, so returns control back to the h1_connection, which sends the aws_io_message and runs the encoder again in some future tick with a fresh aws_io_message...

graebm: whoops, didn't mean to be redundant with the checks

PR #509, opened 2025-03-23
h1_decoder error on multiple content-length headers

### LINKED ISSUE #508: aws-c-http accepts requests with multiple content-length headers
### Describe the bug

RFC 7230 3.3.3 states:
> If a message is received without Transfer-Encoding and with
 either multiple Content-Length header fields having differing
 field-values or a single Content-Length header field having an
 invalid value, then the message framing is invalid and the
 recipient MUST treat it as an unrecoverable error.

Therefore, messages with multiple conflicting Content-Length headers are invalid, and must be rejected with 400 status.

aws-c-http does not implement this behavior. When aws-c-http receives an HTTP/1.1 request with multiple content-length headers, the last content-length header is prioritized, and the request is not rejected.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The request should have been rejected with status 400.

### Current Behavior

The request is incorrectly accepted, prioritizing the second content-length header.

### Reproduction Steps

1. Run an aws-c-http based server that echoes received request bodies, such as [this one](https://github.com/narfindustries/http-garden/blob/main/images/aws_c_http/server.c). 
2. Send it  a request that contains multiple content-length headers.
```sh
printf 'POST / HTTP/1.1\r\nHost: whatever\r\nContent-Length: 1\r\nContent-Length: 2\r\n\r\nZZ' \
    | ncat localhost 80 \
    | grep "body" \
    | jq '.["body"]' \
    | xargs echo \
    | base64 -d
```
3. Observe that the message was accepted, and the second content-length header was prioritized:
```
ZZ
```

### Possible Solution

Respond with status 400 to any request with multiple conflicting content-length headers.

### Additional Information/Context

_No response_

### aws-c-http version used

e3a9cabc664630120df25c28ec710199b8e8b15b

### Compiler and version used

clang 19.1.7

### Operating System and version

Debian 13, Linux 6.12.16-1-lts

quinnj:
Fixes #508


tests/test_connection_manager.c
```diff
@@ -161,13 +161,11 @@ static int s_cm_tester_init(struct cm_tester_options *options) {
...
     }
 
-    tester->event_loop_group = aws_event_loop_group_new(
-        tester->allocator,
-        clock_fn,
-        1,
-        NULL,  // new_loop_fn
-        NULL,  // new_loop_user_data
-        NULL); // shutdown_options
+    struct aws_event_loop_group_options elg_options = {
```
quinnj: ah, is this because I had a newer aws-c-io version locally that I was running against? That makes sense.

quinnj: Yes, sorry, I meant the other way around. I was indeed on an older version.

source/h1_decoder.c
```diff
@@ -411,6 +414,14 @@ static int s_linestate_header(struct aws_h1_decoder *decoder, struct aws_byte_cu
 
     switch (header.name) {
         case AWS_HTTP_HEADER_CONTENT_LENGTH:
+            if (decoder->content_length_received) {
+                AWS_LOGF_ERROR(
+                    AWS_LS_HTTP_STREAM,
+                    "id=%p: Multiple incoming headers for content-length received. This is illegal.",
```
graebm, line 420: As I [mentioned](https://github.com/awslabs/aws-c-http/discussions/493#discussioncomment-11378308) a few months back, the server stuff was in an experimental state. Not something we'd recommend. We never really did a top-to-bottom read of the RFC when implementing HTTP/1. There are very very likely more security issues like this that would affect a server, which needs to worry about receiving malicious requests. At least, much more than a client needs to worry while receiving responses from AWS, which is what this library was primarily developed for.

graebm (approved):

fishb0x:
thanks dawgs.

PR #510, opened 2025-03-28
H2 manual write triggers the stream complete test

TingDaoK:
*Issue #, if available:*

- While I am binding out the function, I found this behavior confusing.
- The data write failed to be read from, but it will fail the data write with `AWS_ERROR_HTTP_STREAM_HAS_COMPLETED`
- But when I dig into the code, it does make sense, the write will fail with stream completed, and the stream will complete with the real error.
- Add one test to display this behavior.

*Description of changes:*


waahm7 (approved):

tests/test_h2_client.c
```diff
@@ -5422,6 +5422,70 @@ TEST_CASE(h2_client_manual_data_write) {
...
+
+    ASSERT_SUCCESS(aws_http2_stream_write_data(stream, &write));
+    aws_input_stream_release(data_stream);
+
+    testing_channel_drain_queued_tasks(&s_tester.testing_channel);
+    ASSERT_TRUE(stream_tester.complete);
+    /* The stream complete will get the error code from the input stream read. */
+    ASSERT_UINT_EQUALS(stream_tester.on_complete_error_code, AWS_IO_STREAM_READ_FAILED);
+    /* The write triggers the stream to complete with error, so the write failed as the stream completes. */
+    ASSERT_UINT_EQUALS(error_code, AWS_ERROR_HTTP_STREAM_HAS_COMPLETED);
```
graebm, line 5477: The docs say that it should work the way you assumed, that the stream-write-complete error-code should have been AWS_IO_STREAM_READ_FAILED, and not AWS_ERROR_HTTP_STREAM_HAS_COMPLETED. Is this easy to fix?

https://github.com/awslabs/aws-c-http/blob/e3a9cabc664630120df25c28ec710199b8e8b15b/include/aws/http/request_response.h#L379-L385

PR #512, opened 2025-04-11
report the error back to write complete

TingDaoK:
*Issue #, if available:*

- follow up on https://github.com/awslabs/aws-c-http/pull/510#discussion_r2027406726

*Description of changes:*

- if `AWS_ERROR_HTTP_STREAM_HAS_COMPLETED` reported, the error SHOULD not be related to the data write. 

### DIFF
--- a/include/aws/http/private/h2_frames.h
+++ b/include/aws/http/private/h2_frames.h
@@ -207,6 +207,7 @@ int aws_h2_encode_frame(
  * body_complete will be set true if encoder reaches the end of the body_stream.
  * body_stalled will be true if aws_input_stream_read() stopped early (didn't
  * complete, though more space was available).
+ * body_failed will be true if the aws_input_stream was the cause of an error.
  *
  * Each call to this function encodes a complete DATA frame, or nothing at all,
  * so it's always safe to encode a different frame type or the body of a different stream
@@ -224,7 +225,7 @@ int aws_h2_encode_data_frame(
     struct aws_byte_buf *output,
     bool *body_complete,
     bool *body_stalled,
-    bool *body_error);
+    bool *body_failed);
 
 AWS_HTTP_API
 void aws_h2_frame_destroy(struct aws_h2_frame *frame);
--- a/include/aws/http/private/h2_frames.h
+++ b/include/aws/http/private/h2_frames.h
@@ -223,7 +223,8 @@ int aws_h2_encode_data_frame(
     size_t *connection_window_size_peer,
     struct aws_byte_buf *output,
     bool *body_complete,
-    bool *body_stalled);
+    bool *body_stalled,
+    bool *body_error);
 
 AWS_HTTP_API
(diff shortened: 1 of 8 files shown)


include/aws/http/private/h2_frames.h
```diff
@@ -223,7 +223,8 @@ int aws_h2_encode_data_frame(
     size_t *connection_window_size_peer,
     struct aws_byte_buf *output,
     bool *body_complete,
-    bool *body_stalled);
+    bool *body_stalled,
+    bool *body_error);
```
graebm, line 227: utterly trivial: `error` isn't this most boolean-ish word
```suggestion
    bool *body_failed);
```

DmitriyMusatkin (approved):

source/h2_stream.c
```diff
@@ -1335,14 +1338,16 @@ static int s_stream_write_data(
...
+                s_stream_data_write_destroy(stream, pending_write);
                 AWS_H2_STREAM_LOG(ERROR, stream, "Cannot write DATA frames to an inactive or closed stream");
                 return aws_raise_error(error_code);
             }
 
             if (stream->synced_data.manual_write_ended) {
                 s_unlock_synced_data(stream);
-                s_stream_data_write_destroy(stream, pending_write, AWS_ERROR_HTTP_MANUAL_WRITE_HAS_COMPLETED);
+                pending_write->error_code = AWS_ERROR_HTTP_MANUAL_WRITE_HAS_COMPLETED;
+                s_stream_data_write_destroy(stream, pending_write);
```
graebm, line 1350: debatable/trivial: it seems a bit fragile that we be so careful about setting write->error_code before calling destroy(). It would be easy to miss a spot, and have a write complete with error_code 0 in some weird error situation.

Maybe go back to passing error_code to the destroy() function.
If a data_frame fails due a write, then destroy the write immediately, instead of leaving it in a list to get cleaned up later. That's how it works with HTTP/1 chunked encoding ([see code](https://github.com/awslabs/aws-c-http/blob/ef78cb890afb9bf768afea6528b7f0fc324f361d/source/h1_encoder.c#L927-L938)). If failure was caused by the chunk, the chunk is immediately destroyed with that error code. Then connection shutdown continues...

source/h2_stream.c
```diff
@@ -815,7 +812,13 @@ int aws_h2_stream_encode_data_frame(
             &connection->thread_data.window_size_peer,
             output,
             &input_stream_complete,
-            &input_stream_stalled)) {
+            &input_stream_stalled,
+            &input_stream_error)) {
+
+        if (input_stream_error) {
+            /* If the error comes from the input stream, propagate it to the current_write */
+            current_write->error_code = aws_last_error();
```
graebm, line 820: (continuing previous comment)
so like here, we could pop the current_write out of its list and destroy it

graebm (approved):

graebm (commented):
I pushed my own code, but you started this PR so can't Accept. Merge if you like, change if you don't

Issue #413, opened 2022-12-28
Support `NO_PROXY` environment variable


poconnor-lab49:
There does not appear to be a way to add a list of hosts which should not be proxied if `HTTP_PROXY` or `HTTPS_PROXY` are set.

The current workaround is to unset `HTTP_PROXY` whenever I want to run against localhost, but this obviously won't work if I need to reach a mix of resources which should and should not be be proxied.

Please add support for `NO_PROXY` and `no_proxy`  to match the `HTTP_PROXY`  env variable support.

davejoyce-lab49:
@poconnor-lab49 I think we may need to agree on `no_proxy` behavior, as it has an interesting history.

See: [We need to talk: Can we standardize NO_PROXY?](https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/)

poconnor-lab49:
> @poconnor-lab49 I think we may need to agree on `no_proxy` behavior, as it has an interesting history.
> 
> See: [We need to talk: Can we standardize NO_PROXY?](https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/)

I read the exact same article!  Very informative.

My gut says to follow the principle of least surprise and do roughly what curl does.  I'd target [the lowest common denominator](https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/#the-lowest-common-denominator) from that article and not bother with CIDR block matching.

bryanlb:
The lack of `NO_PROXY` support for this implementation definitely caught us by surprise when attempting to swap AWS clients. We default traffic through squid proxies and then use `NO_PROXY` to exclude a list of endpoints, which prevents us from using the above workaround.

yasminetalby:
Hello everyone,

Thank you very much for this submission and showcasing your interest in this feature.
To provide an update we have discussed this feature and it is currently in the team backlog.

I will provide all updates about this feature here.

Best regards,

Yasmine

poconnor-lab49:
Thank you and great timing!  See https://github.com/curl/curl/discussions/11962

waahm7:
Hello everyone,

[...]

PR #522, opened 2025-06-25
support no_proxy exactly like CURL

TingDaoK:
*Issue #, if available:*
- https://github.com/awslabs/aws-c-http/issues/413

*Description of changes:*
- Since NO_PROXY has no standard yet. https://about.gitlab.com/blog/we-need-to-talk-no-proxy/ and likely won't have any standard in the foreseeable future (there been many discussions but none of them really driving to a conclusion.).
- It's also very tricky about how to match the host and IP address, and how to handle concern cases and invalid cases.
- In the end, I decided to follow exactly the curl implementation from [noproxy.c](https://github.com/curl/curl/blob/curl-8_14_1/lib/noproxy.c) at 8.14.1
- The implementation basically replicated the CURL logic.
- All the CIDR and host validation all followed how CURL implement it. 
- Given no standards, we are taking curl as the the standard here. 


include/aws/http/private/no_proxy.h
```diff
@@ -0,0 +1,27 @@
...
+ * Check if a host should bypass the proxy based on the NO_PROXY environment variable.
+ * Since NO_PROXY has no standard yet. Follows the curl implementation from noproxy.c at 8.14.1.
+ * https://github.com/curl/curl/blob/curl-8_14_1/lib/noproxy.c
+ *
+ * NO_PROXY is a comma-separated list of domain names, hostnames, or IP addresses that
+ * should bypass the proxy.
+ *
+ * Returns true if the host should bypass the proxy.
+ */
+AWS_HTTP_API bool aws_check_no_proxy(struct aws_allocator *allocator, struct aws_byte_cursor host);
```
DmitriyMusatkin: from testing perspective should this just take a no_proxy string to match against and have some other code retrieve no_proxy from env?

DmitriyMusatkin: also the name could be better. maybe something like aws_http_host_matches_no_proxy?

source/no_proxy.c
```diff
@@ -0,0 +1,368 @@
...
+
+/**
+ * Determines whether a host string is an IPv4 address and stores the binary representation.
+ * Checks if the string follows IPv4 format
+ *
+ * @param host The host string to check
+ * @param addr_out Optional pointer to store the parsed binary address. Must be at least 4 bytes.
+ * @return true if the host is an IPv4 address, false otherwise
+ */
+static bool s_is_ipv4_address(const struct aws_byte_cursor *host, void *addr_out) {
```
DmitriyMusatkin: we have helpers for this and ipv6 in common

source/no_proxy.c
```diff
@@ -0,0 +1,368 @@
...
+ * s_cidr4_match() returns true if the given IPv4 address is within the
+ * specified CIDR address range.
+ * Based on the curl implementation Curl_cidr4_match().
+ *
+ * @param bits The number of network bits in the CIDR notation
+ * @param network_part The network pattern to match against (e.g., "192.168.0.0")\
+ * @param host_addr Pre-parsed binary representation of the host IP, or NULL to parse from host
+ * @return true if the IP address matches the CIDR pattern, false otherwise
+ */
+static bool s_cidr4_match(uint16_t bits, struct aws_byte_cursor network_part, const void *host_addr) {
```
DmitriyMusatkin: its probably fine here as its not as popular as ipv4/ipv6, but consider moving this to common

source/no_proxy.c
```diff
@@ -0,0 +1,368 @@
...
+        return false;
+    }
+
+    struct aws_string *no_proxy_str = aws_get_env_nonempty(allocator, s_no_proxy_env_var_low);
+    if (no_proxy_str == NULL) {
+        no_proxy_str = aws_get_env_nonempty(allocator, s_no_proxy_env_var);
+    }
+
+    if (no_proxy_str == NULL) {
+        aws_string_destroy(no_proxy_str);
```
DmitriyMusatkin: if its null, what would destroying the string do?

source/no_proxy.c
TingDaoK: I'd prefer to keep it here for now. Since this implementation is not really common enough... There are corner cases this implementation didn't handle.

I literally just duplicated the logic from curl and this is how to do CIDR check for no proxy. So... This is more no-proxy specific CIDR check...

source/no_proxy.c
```diff
@@ -0,0 +1,368 @@
...
+ * s_cidr6_match() returns true if the given IPv6 address is within the
+ * specified CIDR address range.
+ * Based on the curl implementation Curl_cidr6_match().
+ *
+ * @param bits The number of network bits in the CIDR notation
+ * @param network_part The network pattern to match against (e.g., "2001:db8::")
+ * @param host_addr Pre-parsed binary representation of the host IP, or NULL to parse from host
+ * @return true if the IP address matches the CIDR pattern, false otherwise
+ */
+static bool s_cidr6_match(uint16_t bits, struct aws_byte_cursor network_part, const void *host_addr) {
```
DmitriyMusatkin: im assuming the void*/casting/and memsets are all of curl? its probably fine, but it does look scary

source/no_proxy.c
```diff
@@ -0,0 +1,368 @@
...
+
+    /* All checks passed, addresses match within the CIDR range */
+    return true;
+}
+
+static bool s_is_dot(uint8_t c) {
+    return c == '.';
+}
+
+/* The host is expected to be the result from  */
```
DmitriyMusatkin: nit: result from where?

source/no_proxy.c
```diff
@@ -0,0 +1,368 @@
...
+        return true;
+    }
+    bool bypass = false;
+    struct aws_byte_cursor no_proxy_cur = aws_byte_cursor_from_string(no_proxy_str);
+    struct aws_array_list no_proxy_list;
+    if (aws_array_list_init_dynamic(&no_proxy_list, allocator, 10, sizeof(struct aws_byte_cursor))) {
+        goto cleanup;
+    }
+    /* Split the NO_PROXY string by commas */
+    if (aws_byte_cursor_split_on_char(&no_proxy_cur, ',', &no_proxy_list)) {
```
DmitriyMusatkin, line 136: probably want to confirm this works correctly against badly formatted input, like multiple commas back to back or commas at the start or end of no proxy

source/no_proxy.c
```diff
@@ -0,0 +1,368 @@
...
+                }
+                if (aws_byte_cursor_next_split(&pattern, '/', &substr)) {
+                    /* This substr will be number of bits. */
+                    if (substr.len > 8) {
+                        /* Invalid, ignore it. */
+                        continue;
+                    }
+                    memcpy(bits_buffer, substr.ptr, substr.len);
+                    bits_buffer[substr.len] = '\0'; /* Null-terminate for atoi */
+                    network_bits = (uint16_t)atoi(bits_buffer);
```
DmitriyMusatkin: https://github.com/awslabs/aws-c-common/blob/2b67a658e461520f1de20d64342b91ddcedc7ebb/include/aws/common/byte_buf.h#L943?

source/no_proxy.c
TingDaoK: yep, added tests https://github.com/awslabs/aws-c-http/pull/522/files#diff-6e4a6d16e821a9d15702525903e75d495b445d08639f9787b4445b70343a2a9eR119

DmitriyMusatkin (approved):

Issue #523, opened 2025-07-16
Compilation of v0.10.3 broken on FreeBSD


giordano:
### Describe the bug

```
[02:38:46] /workspace/srcdir/aws-c-http/source/no_proxy.c:41:19: error: use of undeclared identifier 'AF_INET'
[02:38:46]    41 |     if (inet_pton(AF_INET, aws_string_c_str(network_part), &check) != 1) {
[02:38:46]       |                   ^
[02:38:46] /workspace/srcdir/aws-c-http/source/no_proxy.c:82:19: error: use of undeclared identifier 'AF_INET6'
[02:38:46]    82 |     if (inet_pton(AF_INET6, aws_string_c_str(network_part), check) != 1) {
[02:38:46]       |                   ^
[02:38:46] /workspace/srcdir/aws-c-http/source/no_proxy.c:146:19: error: use of undeclared identifier 'AF_INET'
[02:38:46]   146 |     if (inet_pton(AF_INET, aws_string_c_str(host_str), &ipv4_addr) == 1) {
[02:38:46]       |                   ^
[02:38:46] /workspace/srcdir/aws-c-http/source/no_proxy.c:158:23: error: use of undeclared identifier 'AF_INET6'
[02:38:46]   158 |         if (inet_pton(AF_INET6, aws_string_c_str(host_str_copy), ipv6_addr) == 1) {
[02:38:46]       |                       ^
[02:38:46] 4 errors generated.
```
Compilation of v0.10.2 was successful.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

Compilation is successful

### Current Behavior

Compilation fails

### Reproduction Steps

Compile v0.10.3 on FreeBSD

### aws-c-http version used

v0.10.3

### Compiler and version used

Clang 19

### Operating System and version

FreeBSD

TingDaoK:
thanks for pointing out this issue. I just merged the fix and released the new version

PR #524, opened 2025-07-16
Add openBSD/freeBSD CI

TingDaoK:
*Issue #, if available:*

- https://github.com/awslabs/aws-c-http/issues/523

*Description of changes:*

- Add CI for FreeBSD/OpenBSD
- Apparently, FreeBSD requires different header file for `inet_pton`


bretambrose (commented):
As a short term fix seems fine, but long term I feel like there should not be platform-specific anything outside of common/io/cal

PR #531, opened 2025-08-08
Fix additional warnings found by the Undefined Behavior Sanitizer

bjosv:
Fixes runtime error warnings found when running tests in the [aws-crt-cpp](https://github.com/awslabs/aws-crt-cpp) repo with the undefined behavior sanitizer enabled.

```
cd aws-crt-cpp
mkdir build && cd build
CC=clang-18 CXX=clang++-18 cmake -G Ninja -DENABLE_SANITIZERS=ON -DSANITIZERS="undefined" ..
ninja
ninja test
grep "runtime error" Testing/Temporary/LastTest.log
...
aws-crt-cpp/crt/aws-c-common/source/hash_table.c:68:12: runtime error: call to function aws_byte_cursor_eq through pointer to incorrect function type 'bool (*)(const void *, const void *)'
...
```

*Description of changes:*
Add callback functions that uses correct signature [aws_hash_callback_eq_fn](https://github.com/awslabs/aws-c-common/blob/cde20c682d46ed8ee03ab6412c3f386f5dbdd0c7/include/aws/common/hash_table.h#L110) (as also seen in [aws-c-mqtt](https://github.com/awslabs/aws-c-mqtt/blob/f9aefbc95e63d00eada405fc4d69b2087cbc06de/source/client_impl_shared.c#L220)).

Since the CI-build is not using `-fno-sanitize-recover=all`, and the warnings are recoverable, they will not fail the CI build.


bjosv:
@TingDaoK I see that aws-c-http is running sanitizers in CI, but without the `-fno-sanitize-recover=all` flag.
Maybe we should add that as a default, what do you think?

PR #537, opened 2025-12-12
Add helper to check for transient errors

azkrishpy:
*Issue #, if available:*

*Description of changes:*
An extension of the [IO implementation](https://github.com/awslabs/aws-c-io/pull/782) to check for error codes indicating a transient error.

### DIFF
--- a/include/aws/http/http.h
+++ b/include/aws/http/http.h
@@ -123,9 +123,10 @@ void aws_http_library_init(struct aws_allocator *alloc);
 AWS_HTTP_API
 void aws_http_library_clean_up(void);
 
-/**
- * Helper used downstream to check if the HTTP error codes can be classified
- * as transient errors.
+/*
+ * This API provides a recommendation on whether an error code should be considered retryable for transient
+ * errors like host resolution, sockets, and TLS.
+ * Note: This is a recommendation only. Retry behavior should be determined based on your specific use case.
  */
 AWS_HTTP_API
 bool aws_http_error_code_is_retryable(int error_code);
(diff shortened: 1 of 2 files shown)


source/http.c
```diff
@@ -8,6 +8,7 @@
 #include <aws/http/private/hpack.h>
 #include <aws/http/private/http_impl.h>
 #include <aws/http/status_code.h>
+#include <aws/io/io.h>
```
TingDaoK: you probably don't need this. As http.h already includes io.h, right?

source/http.c
```diff
@@ -556,6 +557,15 @@ void aws_http_fatal_assert_library_initialized(void) {
     }
 }
 
+bool aws_http_error_code_is_transient(int error_code) {
+    switch (error_code) {
+        case AWS_ERROR_HTTP_CONNECTION_CLOSED:
+        case AWS_ERROR_HTTP_SERVER_CLOSED:
```
TingDaoK, line 565: consider adding `AWS_ERROR_HTTP_PROXY_CONNECT_FAILED_RETRYABLE`.

`AWS_ERROR_HTTP_STREAM_MANAGER_CONNECTION_ACQUIRE_FAILURE` is tricky that we basically throw this for everything failed to get the connection when using stream manager. 

Let's add a TODO/notes, chat with Swift SDK team, and see how they retries and how much they can be benefit from this change.

include/aws/http/http.h
```diff
@@ -123,6 +123,13 @@ void aws_http_library_init(struct aws_allocator *alloc);
 AWS_HTTP_API
 void aws_http_library_clean_up(void);
 
+/**
+ * Helper used downstream to check if the HTTP error codes can be classified
+ * as transient errors.
```
TingDaoK: update it

PR #538, opened 2025-12-19
Fix CI issues

azkrishpy:
*Issue #, if available:*

*Description of changes:*
- address changes in behavior since https://github.com/awslabs/aws-c-io/pull/758
- change in behavior for apple (Server side ALPN now works for MacOS as well)
- disabled proxy test for apple platforms. Should re-enabled later.


tests/test_connection.c
```diff
@@ -735,14 +735,18 @@ static int s_test_connection_customized_alpn_error_with_unknown_return_string(
     tester.wait_client_connection_num = 1;
     tester.wait_server_connection_num = 1;
 
-#ifndef __APPLE__ /* Server side ALPN doesn't work for MacOS */
+    // TODO: Investigate further and fix behavior taking into account the recent migration
+    // to newer Apple Netowrk Frameowrk on aws-c-io. Implemented temporary fix for CI tests.
+    // CI fails on local for some machines. Still under investigation.
```
TingDaoK: why not put the notes of what we found yesterday here?

tests/CMakeLists.txt
```diff
@@ -652,7 +652,9 @@ add_test_case(test_no_proxy_invalid_patterns)
 add_test_case(test_no_proxy_invalid_host_inputs)
 
 add_test_case(http_tunnel_proxy_connection_success)
+if(NOT APPLE)
```
TingDaoK: skip from CMakeList will be hard to notice. 
do something similar to
```
TEST_CASE(multipart_upload_with_network_interface_names_mock_server) {
    (void)ctx;
#if defined(AWS_OS_WINDOWS) || defined(AWS_OS_APPLE)
    (void)allocator;
    return AWS_OP_SKIP;
#else
```

TingDaoK (approved):

PR #540, opened 2025-12-22
Automate the renew of the cert used in test

TingDaoK:
*Issue #, if available:*
- AI generated Github Action to automate the process to renew the cert used in the tests.
- We use cert in our tests for the local server and client the test the tls. But the cert cannot last forever, when it expires, we need to manually renew them.
- It's a bit hard to find where the issue is given the error is just tls time out mostly. So, automate this process to happen is much better comparing to let the devs to do this every ~800 days. 
- https://github.com/awslabs/aws-c-http/pull/541 tested the path to generate the new cert and create PR

*Description of changes:*
- https://github.com/awslabs/aws-c-http/actions/runs/20445411465/job/58747688996?pr=540 after we already merged the renewed cert
- https://github.com/awslabs/aws-c-http/actions/runs/20444935869/job/58746246658 before we merge the cert. It detects the cert has expired, and create new PR https://github.com/awslabs/aws-c-http/pull/541 that works.

TODO: should apply similarly to aws-c-io

### DIFF
--- a/include/aws/http/http.h
+++ b/include/aws/http/http.h
@@ -123,6 +123,14 @@ void aws_http_library_init(struct aws_allocator *alloc);
 AWS_HTTP_API
 void aws_http_library_clean_up(void);
 
+/*
+ * This API provides a recommendation on whether an error code should be considered retryable for transient
+ * errors like host resolution, sockets, and TLS.
+ * Note: This is a recommendation only. Retry behavior should be determined based on your specific use case.
+ */
+AWS_HTTP_API
+bool aws_http_error_code_is_retryable(int error_code);
+
 /**
  * Returns the description of common status codes.
  * Ex: 404 -> "Not Found"
--- a/source/http.c
+++ b/source/http.c
@@ -556,6 +556,19 @@ void aws_http_fatal_assert_library_initialized(void) {
     }
 }
 
+/*
+ * This might need to get updated with more http error codes based on consensus.
+ */
+bool aws_http_error_code_is_retryable(int error_code) {
+    switch (error_code) {
+        case AWS_ERROR_HTTP_CONNECTION_CLOSED:
+        case AWS_ERROR_HTTP_SERVER_CLOSED:
+        case AWS_ERROR_HTTP_PROXY_CONNECT_FAILED_RETRYABLE:
+            return true;
+    }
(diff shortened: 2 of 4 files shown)


.github/workflows/cert-renewal.yml
```diff
@@ -0,0 +1,145 @@
...
+          \`\`\`bash
+          openssl req -x509 -new -key unittests.key -config unittests.conf -out unittests.crt -days 824
+          openssl pkcs12 -export -out unittests.p12 -inkey unittests.key -in unittests.crt -password pass:1234 -keypbe PBE-SHA1-3DES -certpbe PBE-SHA1-3DES -macalg sha1
+          \`\`\`
+
+          ### Testing
+          Please verify that the unit tests pass with the new certificates before merging.
+
+          ---
+          🤖 This PR was automatically generated by the certificate renewal workflow." \
```
azkrishpy, line 143: are we okay with checking in emojis in code lol

.github/workflows/cert-renewal.yml
```diff
@@ -0,0 +1,145 @@
...
+
+          if [ ! -f "$CERT_FILE" ]; then
+            echo "Certificate file not found!"
+            echo "needs_renewal=true" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          # Get certificate expiration date in seconds since epoch
+          EXPIRY_DATE=$(openssl x509 -enddate -noout -in "$CERT_FILE" | cut -d= -f2)
+          EXPIRY_EPOCH=$(date -d "$EXPIRY_DATE" +%s 2>/dev/null || date -j -f "%b %d %H:%M:%S %Y %Z" "$EXPIRY_DATE" +%s)
```
sfod, line 36: What's the plan for when we add new certs to CI? I see two problems:
1) Checking multiple certs will be a mess. I guess we can synchronize renewal of all certs, so if `unittests.crt` is expired then all certs should be renewed.
2) A developer has to remember to add it to this job on adding a new cert to repo.

We don't add certs to this repo all the time, so these problems are minor. But still worth thinking about it.

TingDaoK: yeah, I think it makes sense for all the certs to be updated all at once for simplicity.

But, if any new certs to be added, we need to update this script as well. I can add it to the readme about the certs. good callout

PR #544, opened 2026-01-28
[fix] h2 stream manager initial settings not passed correctly & Log the headers

TingDaoK:
*Issue #, if available:*
1. When the initial settings set for h2 stream manager, the copy of the initial settings has a bug that didn't set the length of array correctly after the data copied. And end with the initial settings was not respected.
2. When converting http/1.1 header to h2, forget to skip `connection` header that is also not allowed
3. `An endpoint MUST NOT generate an HTTP/2 message containing connection-specific header fields.` so that let's check the headers we send out, and error out if those headers are found

*Description of changes:*
- Add tests. Also, issues warnings when the flow control window is 0 and stops the flow so that we can track this easier
- Fix the headers related bug
- Log the headers so that we can debug easier
- Skip the `authorization` headers in case of sensitive information.


source/h1_stream.c
```diff
@@ -431,6 +431,34 @@ struct aws_h1_stream *aws_h1_stream_new_request(
...
 
+    struct aws_http_headers *headers = aws_http_message_get_headers(options->request);
+
+    /* Log the headers that we are sending out. */
+    for (size_t i = 0; i < aws_http_headers_count(headers); i++) {
+        struct aws_http_header header;
+        aws_http_headers_get_index(headers, i, &header);
+        enum aws_http_header_name name_enum = aws_http_str_to_header_name(header.name);
+        switch (name_enum) {
+            case AWS_HTTP_HEADER_AUTHORIZATION:
```
DmitriyMusatkin, line 442: can we also skip session/security token?

DmitriyMusatkin (approved):

PR #547, opened 2026-02-16
Async streams for HTTP

azkrishpy:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/aws/http/private/h1_connection.h
+++ b/include/aws/http/private/h1_connection.h
@@ -120,6 +120,8 @@ struct aws_h1_connection {
         bool is_outgoing_stream_task_active : 1;
 
         bool is_processing_read_messages : 1;
+
+        struct aws_io_message *pending_async_message;
     } thread_data;
 
     /* Any thread may touch this data, but the lock must be held */
--- a/include/aws/http/request_response.h
+++ b/include/aws/http/request_response.h
@@ -958,7 +958,9 @@ void aws_http_message_set_body_stream(struct aws_http_message *message, struct a
  * The stream must not be destroyed until the message is complete.
  */
 AWS_HTTP_API
-void aws_http_message_set_async_body_stream(struct aws_http_message *message, struct aws_async_input_stream *async_body_stream);
+void aws_http_message_set_async_body_stream(
+    struct aws_http_message *message,
+    struct aws_async_input_stream *async_body_stream);
 
 /**
  * aws_future<aws_http_message*>
--- a/include/aws/http/request_response.h
+++ b/include/aws/http/request_response.h
@@ -936,6 +936,13 @@ AWS_HTTP_API
 struct aws_input_stream *aws_http_message_get_body_stream(const struct aws_http_message *message);
 
 /**
+ * Get the async body stream.
+ * Returns NULL if no async body stream is set.
+ */
+AWS_HTTP_API
+struct aws_async_input_stream *aws_http_message_get_async_body_stream(const struct aws_http_message *message);
+
+/**
  * Set the body stream.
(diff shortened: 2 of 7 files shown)


include/aws/http/request_response.h
```diff
@@ -874,6 +881,17 @@ struct aws_input_stream *aws_http_message_get_body_stream(const struct aws_http_
 AWS_HTTP_API
 void aws_http_message_set_body_stream(struct aws_http_message *message, struct aws_input_stream *body_stream);
 
+/**
+ * Set the async body stream.
+ * NULL is an acceptable value for messages with no body.
+ * Note: The message does NOT take ownership of the body stream.
+ * The stream must not be destroyed until the message is complete.
```
TingDaoK, line 958: we should, and the same comments for `aws_http_message_set_body_stream` is actually out dated, that I forgot to remove.

So, the input streams are refcounted, so that we should keep a refcount on it with the http message.

include/aws/http/private/h1_connection.h
```diff
@@ -120,6 +120,8 @@ struct aws_h1_connection {
         bool is_outgoing_stream_task_active : 1;
 
         bool is_processing_read_messages : 1;
+
+        struct aws_io_message *pending_async_message;
```
TingDaoK, line 124: comment on what does this do?

source/h1_connection.c
```diff
@@ -1113,7 +1130,16 @@ static void s_write_outgoing_stream(struct aws_h1_connection *connection, bool f
...
     }
 
-    if (msg->message_data.len > 0) {
+    if (connection->thread_data.encoder.async_error) {
+        /* Error receiving data asynchronously. Need to note when this is happening, but for now, abandon ship */
+        if (msg) {
+            aws_mem_release(msg->allocator, msg);
+        }
+        s_shutdown_due_to_error(connection, connection->thread_data.encoder.async_error);
+        return;
```
TingDaoK, line 1165: trivial: seems like we can just goto error, with a `aws_raise_error(connection->thread_data.encoder.async_error)`

source/h1_connection.c
```diff
@@ -1030,6 +1030,17 @@ static void s_on_channel_write_complete(
      * to run again instead of calling the function directly.
      * This way, if the message completes synchronously,
      * we're not hogging the network by writing message after message in a tight loop */
+
+    /* If encoder is waiting for async body read, don't reschedule - the async callback will do it */
+    if (connection->thread_data.encoder.state == AWS_H1_ENCODER_STATE_ASYNC_WAITING) {
```
TingDaoK, line 1058: the connection is a user of the encoder, so it is better to not know the implementation detail of the encoder. 

It's better to have a helper function to check if encoder is waiting for data async, instead of just querying the state of the encoder.

include/aws/http/request_response.h
TingDaoK: to make the design easier to use, let's say it's an error to set both. you can only set either one of the regular body stream or the async body stream

source/h1_connection.c
```diff
@@ -1552,6 +1593,9 @@ static struct aws_h1_connection *s_connection_new(
 
     aws_h1_encoder_init(&connection->thread_data.encoder, alloc);
 
+    /* hacking around with adding connection to encoder. there should be a better way to do it */
+    connection->thread_data.encoder.connection = connection;
```
TingDaoK, line 1623: the better way to decouple them is probably expose something like
```
aws_encoder_set_async_stream_read_done(callback, user_data);
```
So that we keep the logic of handling the stream read done within the connection instead of the encoder

Issue #546, opened 2026-02-05
aws-c-http does not read the entire response body after receiving Connection: Close


netapp-zwilliam:
### Describe the bug

aws mountpoint-s3 receives a Complete Multipart Upload response with a Connection: Close but does not finish reading the body of the Complete Multipart Upload response

See https://github.com/awslabs/mountpoint-s3/issues/1741 for more context.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

From https://datatracker.ietf.org/doc/html/rfc7230#section-6.6 

   A server that sends a "close" connection option MUST initiate a close
   of the connection (see below) after it sends the response containing
   "close".  The server MUST NOT process any further requests received
   on that connection.

   A client that receives a "close" connection option MUST cease sending
   requests on that connection and close the connection **after reading
   the response message containing the "close"**; if additional pipelined
   requests had been sent on the connection, the client SHOULD NOT
   assume that they will be processed by the server.

As noted in the spec that https://github.com/awslabs/aws-c-http/blob/a9745ea9998f679cd7456e7d23cc8820e38c97d4/source/h1_connection.c#L1455 is referring to, it should close the connection after reading the response message. Based on the logs above 

### Current Behavior

It stops reading data from the final stream, even though based on the spec it should only close the data **after* reading the response.

```
2025-09-17T03:19:46.570758Z TRACE awscrt::http-connection: id=0x7fba040229d0: Incoming message of size 271.
2025-09-17T03:19:46.570767Z TRACE awscrt::http-stream: id=0x7fba04045d40: Incoming response status: 200 (OK).
2025-09-17T03:19:46.570806Z TRACE awscrt::http-stream: id=0x7fba04045d40: Incoming header: Date: Wed, 17 Sep 2025 03:19:46 GMT
**2025-09-17T03:19:46.570833Z TRACE awscrt::http-stream: id=0x7fba04045d40: Incoming header: Connection: CLOSE
2025-09-17T03:19:46.570840Z TRACE awscrt::http-stream: id=0x7fba04045d40: Received 'Connection: close' header. This will be the final stream on this connection.**
2025-09-17T03:19:46.570844Z TRACE awscrt::http-stream: id=0x7fba04045d40: Incoming header: Server: <redacted>
2025-09-17T03:19:46.570847Z TRACE awscrt::http-stream: id=0x7fba04045d40: Incoming header: x-amz-request-id: #############
[...]

PR #550, opened 2026-02-20
Fix unkown length response

TingDaoK:
*Issue #, if available:*
- https://github.com/awslabs/aws-c-http/issues/546

*Description of changes:*
- The HTTP client currently treat the response completes after the headers if no transfer encoding nor content-length received. 
- But, the SPEC clearly mentioned that is a valid case for the response to provide indeterminate length response. (But request is not allowed to do so.) 
- Add support to handle the case by leave the decoder 


source/h1_decoder.c
```diff
@@ -688,6 +714,9 @@ static int s_linestate_response(struct aws_h1_decoder *decoder, struct aws_byte_
 
     /* RFC-7230 section 3.3 Message Body */
     decoder->body_headers_ignored |= code_val == AWS_HTTP_STATUS_CODE_304_NOT_MODIFIED;
+    if (decoder->body_headers_ignored_on_2xx) {
+        decoder->body_headers_ignored |= code_val / 200 == 1;
```
azkrishpy:
```
2xx (Successful) responses to a CONNECT request method (Section 4.3.6 of[RFC7231]) 
switch to tunnel mode instead of having a message body. All 1xx (Informational), 
204 (No Content), and 304 (Not Modified) responses do not include a message body.  
All other responses do include a message body, although the body might be of zero length.
```
we seem to ignore body for any value between 200 and 400? Is that correct?

TingDaoK: yeah, I was wrong about the math, this is supposed to be status code from 200-299. 
Good catch!

azkrishpy:
```suggestion
        decoder->body_headers_ignored |= code_val / 100 == 2;
```

PR #558, opened 2026-05-08
[fix] h2 double complete

TingDaoK:
*Issue #, if available:*

*Description of changes:*
- In the case where the the remote sends GOAWAY marks the stream to be half closed, we will try to mark the stream complete, since we get a GOAWAY
- Meanwhile, if user also want to cancel the stream and sending a RST_STREAM. Nothing prevents this, and spec also allows RST_STREAM to be sent after receiving GOAWAY. However, with sending RST_STREAM, we also try to mark the stream complete. And in which case, we complete the stream twice and everything blows up. 
- Add a simple boolean to prevent it from happening. 

Arguably, we can prevent the RST_STREAM from a various places (User invoke, running the cross thread task ,etc..) But, sending it is not the real problem, and we do protect a lot around those when the connection/stream is shutting down. So, I chose to prevent the complete to happen again, so, every place can safely try to complete the stream regardless if it's already completed or not.


tests/test_stream_manager.c
```diff
@@ -1152,6 +1152,63 @@ TEST_CASE(h2_sm_mock_goaway) {
     return s_tester_clean_up();
 }
 
+/* Test that cancelling a stream after GOAWAY does not crash from double stream completion.
+ * Reproduces P428540711: cancel() schedules cross-thread task, but s_finish_shutdown already
```
azkrishpy, line 1156: should we mention this _magic_ number?

azkrishpy (approved):

DmitriyMusatkin (approved):

tests/test_stream_manager.c
TingDaoK: it's a comment in the test, so pretty much for ourselves to reference.

PR #559, opened 2026-05-18
Fix rounding error in hpack resizing

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/hpack.c
+++ b/source/hpack.c
@@ -414,7 +414,7 @@ int aws_hpack_insert_header(struct aws_hpack_context *context, const struct aws_
     }
 
     /* If we're out of space in the buffer, grow it */
-    if (context->dynamic_table.num_e lements == context->dynamic_table.buffer_capacity) {
+    if (context->dynamic_table.num_elements == context->dynamic_table.buffer_capacity) {
         /* If the buffer is currently of 0 size, reset it back to its initial size */
         const size_t new_size =
             context->dynamic_table.buffer_capacity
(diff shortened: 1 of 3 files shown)


source/hpack.c
```diff
@@ -421,7 +418,8 @@ int aws_hpack_insert_header(struct aws_hpack_context *context, const struct aws_
         /* If the buffer is currently of 0 size, reset it back to its initial size */
         const size_t new_size =
             context->dynamic_table.buffer_capacity
-                ? (size_t)(context->dynamic_table.buffer_capacity * s_hpack_dynamic_table_buffer_growth_rate)
+                /* increase buffer by 1.5 rounded up. */
+                ? (size_t)(context->dynamic_table.buffer_capacity + ((context->dynamic_table.buffer_capacity + 1) / 2))
```
TingDaoK, line 422: we cannot get a 0 `buffer_capacity`?

Will it be more clear if we do a `min(context->dynamic_table.buffer_capacity / 2, 1)`?

TingDaoK (approved):

source/hpack.c
DmitriyMusatkin: from my reading of code, s_dynamic_table_resize_buffer supports new size of 0, which will result in buffer_capacity being 0. 
wouldn't that min always result in 1?

TingDaoK: oops, max not min

PR #561, opened 2026-06-18
Fix connection recycling with stalled HTTP 1.1 streams

sbSteveK:
The `HttpClientConnectionManager` using `manual_window_management = true` was able to release a connection back to the pool while a HTTP 1.1 stream was still in progress. e.g. the stream window filled to zero and the user never called `update_window`. The connection manager would be able to vend the same connection for a new request but because HTTP 1.1 processes streams in FIFO order, the new stream's response callback would never fire and be permanently blocked waiting for the stalled stream to complete. This caused requests to hang indefinitely after a previous request on the same connection was abandoned with an unconsumed response body.

This can be seen as a user error but we should handle it.

A num_streams_in_progress counter is added to `aws_h1_connection.synced_data` to track client streams that have been submitted but not yet completed. The counter is incremented when a stream is activated and decremented in the existing critical section of `s_stream_complete()`

An internal `aws_http_connection_is_connection_idle()` function is introduced, backed by a connection vtable function and wired through the connection manager's system vtable. For H1 connections, this returns false when num_streams_in_progress > 0. For H2 connections, it unconditionally returns true (H2 multiplexes streams independently and uses GOAWAY for lifecycle management).

The connection manager release path now checks both `new_requests_allowed` (connection alive/not errored) and `is_connection_idle` (no orphaned streams). If the connection is not idle, it is destroyed instead of recycled, preventing the stall.

Tests:
Check that new requests are allowed during pipelining.
Check that `is_connection_idle` returns false during active/stalled streams and true after completion.

CI:
split out the downstream ci job into its own workflow to allow us to retry the significantly shorter running jobs if they failed a flaky test.
fail-fast on matrix CI jobs to false so we can allow other jobs to continue running if one flaky test fails in one matrix element.


source/h1_connection.c
```diff
@@ -272,12 +272,14 @@ static bool s_connection_is_open(const struct aws_http_connection *connection_ba
...
     int new_stream_error_code;
+    size_t num_streams_in_progress;
     { /* BEGIN CRITICAL SECTION */
         aws_h1_connection_lock_synced_data(connection);
         new_stream_error_code = connection->synced_data.new_stream_error_code;
+        num_streams_in_progress = connection->synced_data.num_streams_in_progress;
         aws_h1_connection_unlock_synced_data(connection);
     } /* END CRITICAL SECTION */
-    return new_stream_error_code == 0;
+    return new_stream_error_code == 0 && num_streams_in_progress == 0;
```
TingDaoK: This will be invoked from a public function 
```
/**
 * Return whether the connection can make a new requests.
 * If false, then a new connection must be established to make further requests.
 */
AWS_HTTP_API
bool aws_http_connection_new_requests_allowed(const struct aws_http_connection *connection);
```

Why adding `num_streams_in_progress == 0;`? It means if we have any stream in progress, we don't allow new requests? 

This will be a behavior breaking change. We do allow creating multiple streams per connection, even though it will be piped by the connection to handle them one by one.

include/aws/http/private/h1_connection.h
```diff
@@ -136,6 +136,11 @@ struct aws_h1_connection {
         /* If non-zero, reason to immediately reject new streams. (ex: closing) */
         int new_stream_error_code;
 
+        /* Number of streams in new_client_stream_list + thread_data.stream_list.
+         * Incremented when a stream is submitted, decremented when s_stream_complete() fires.
+         * Used by new_requests_allowed() to prevent recycling connections with active streams. */
+        size_t num_streams_in_progress;
```
TingDaoK, line 142: why not use an atomic?

TingDaoK (commented):
Overall, I am not against the idea when user releases the connection back to the pool, close it when there is still streams alive or even error it out. 

If there is outstanding request/response and user release it back to the pool, it to me like freeing an object that is still actively in use. We can be proactive to either prevent it, or be nice to close the connection so that the requests will be closed as well.

include/aws/http/private/h1_connection.h
sbSteveK: We've decided to stick to non-atomic but to move the decrement into existing lock of `s_stream_complete()` function.

source/h1_connection.c
sbSteveK: pulled this out so we don't run into issued with multiple streams per connection and added a separate `is_connection_idle` that checks whether there are any active streams on the connection going.

source/h2_connection.c
```diff
@@ -2266,6 +2268,13 @@ static bool s_connection_new_requests_allowed(const struct aws_http_connection *
     return new_stream_error_code == 0;
 }
 
+static bool s_h2_connection_is_idle(const struct aws_http_connection *connection_base) {
+    /* HTTP/2 multiplexes streams and handles cleanup via GOAWAY.
+     * The connection manager uses separate H2-specific mechanisms for lifecycle management. */
+    (void)connection_base;
+    return true;
+}
```
TingDaoK, line 2294: i don't know what this means... 

I'd say either we raise error that it's not currently supported in http/2 or actually support this helper (not for the reason to control the lifetime from connection manager, but it is nice to have to check if the connection is idle or not)

sbSteveK: HTTP/2 doesn't run into the issue because it multiplexes streams (preventing a stream not completed from blocking the connection entirely) and uses GOAWAY ( https://github.com/awslabs/aws-c-http/blob/8acf604e12137d8e033c57108eecf3c352d4e0cc/include/aws/http/connection.h#L79-L93 ) from peer to prevent a `new_request_allowed()` from returning true. It also uses the stream manager to check if it has streams assigned to it before allowing it to be released. Re-reading this description, it's pretty weird without context...

This function being added to the v-table was a result of needing to get this info from a testing perspective in our test framework. That meant having to add it to HTTP/2 and figured it'd be easiest to just make it a no-op.

It makes sense to allow it to return the actual idle state of an HTTP/2 connection instead of just returning true since it's here. I've added a best-effort check on the idle state but removed this from the being checked on HTTP/2 connections during the recycle check because HTTP/2 connections have different behavior that doesn't require it.

PR #568, opened 2026-07-28
Enforce MAX_HEADER_LIST_SIZE

azkrishpy:
*Issue #, if available:*

*Description of changes:*
We enable an advisory setting `AWS_HTTP2_SETTINGS_MAX_HEADER_LIST_SIZE` to limit the size of headers we accept but we do not enforce it. This PR enforces the setting to ensure a malicious server cannot send us bloated number of headers. 


source/h2_decoder.c
```diff
@@ -1247,6 +1261,15 @@ static struct aws_h2err s_process_header_field(
         goto already_malformed;
     }
 
+    /* RFC-7541 6.5.1 header-list-size accounting, applied to every field (including cookie fields
```
TingDaoK: > RFC-7541 6.5.1

404

TingDaoK (commented):
we should have a default better than uint32 max, right?

azkrishpy:
> we should have a default better than uint32 max, right?

yep fully missed this

source/h2_decoder.c
azkrishpy: *crying *

tests/CMakeLists.txt
```diff
@@ -480,6 +480,7 @@ add_test_case(h2_client_conn_err_invalid_last_stream_id_goaway)
 
 # TODO add_test_case(h2_client_send_goaway_with_push_promises) id of 1st should be in GOAWAY 2nd should be ignored
 add_test_case(h2_client_change_settings_succeed)
+add_test_case(h2_client_max_header_list_size_setting_is_enforced)
```
sbSteveK: Trivial: This looks like it's testing increasing the size and the new size being enforced (a size over it fails).

We probably want a handful of other tests related to this:
test that checks against the default (32 KB) and that it's properly enforced (fails if > 32 KB).
test that sets that raising the max header list size allows a header that is close to but not exceeding that increased size doesn't fail.
test that lowers the max header list size and it's properly enforced (to check the other direction) from the existing test.

sbSteveK (approved):
Add tests for coverage in all directions. Everything else looks solid to me.

tests/CMakeLists.txt
azkrishpy: Makes sense. Will add other cases.

TingDaoK (approved):

PR #572, opened 2026-08-03
HPACK String Length Validation

sbSteveK:
Validate HPACK string length against SETTINGS_MAX_HEADER_LIST_SIZE at decode time.


.github/workflows/ci.yml
```diff
@@ -22,6 +22,7 @@ jobs:
   linux-compat:
     runs-on: ubuntu-24.04 # latest
     strategy:
+      fail-fast: false
```
azkrishpy, line 25: we don't have to set this to false, we can just look at whichever test failed and make sure it failed during hpack decode itself.

source/hpack_decoder.c
```diff
@@ -25,6 +25,9 @@ void aws_hpack_decoder_init(struct aws_hpack_decoder *decoder, struct aws_alloca
     aws_byte_buf_init(&decoder->progress_entry.scratch, allocator, s_hpack_decoder_scratch_initial_size);
 
     decoder->dynamic_table_protocol_max_size_setting = aws_hpack_get_dynamic_table_max_size(&decoder->context);
+
+    /* Default: no limit until h2_decoder sets it from SETTINGS_MAX_HEADER_LIST_SIZE */
+    decoder->max_header_list_size = UINT64_MAX;
```
azkrishpy: we should have a smaller limit than UINT64_MAX.

.github/workflows/ci.yml
sbSteveK: I'm setting this to false so that when we start a CI run, if one test within the matrix fails due to a flaky test, the other ones in the matrix can continue to completion so we don't have to restart all of them.

source/hpack_decoder.c
sbSteveK: I thought about this and I want the value to be set in one place. When the h2 decoder initializes, it sets the value for the hpack decoder. If the h2 decoder has this set, it also applies it to the hpack decoder. If we set it to the default here separately, we would have to remember to change it in both places in the future and there could also be a missmatch between this and the h2 decoder. Best to set it to max here and let it be set by the h2 decoder.

source/h2_decoder.c
```diff
@@ -340,6 +340,10 @@ struct aws_h2_decoder *aws_h2_decoder_new(struct aws_h2_decoder_params *params)
     decoder->settings.max_frame_size = aws_h2_settings_initial[AWS_HTTP2_SETTINGS_MAX_FRAME_SIZE];
     decoder->settings.max_header_list_size = aws_h2_settings_initial[AWS_HTTP2_SETTINGS_MAX_HEADER_LIST_SIZE];
 
+    /* Propagate the header-list size limit to the HPACK decoder so it can reject
+     * oversized strings at decode time, before a complete header field is assembled. */
+    aws_hpack_decoder_set_max_header_list_size(&decoder->hpack, decoder->settings.max_header_list_size);
```
azkrishpy: I think we can make this part of `aws_hpack_decoder_init()` since this is copying over a static setting, we don't need an api call.

source/hpack_decoder.c
azkrishpy: during `init` it is a static setting so we can just copy over `aws_h2_settings_initial[AWS_HTTP2_SETTINGS_MAX_HEADER_LIST_SIZE]`.

source/hpack_decoder.c
azkrishpy: we can also remove this comment now.

tests/test_h2_decoder.c
```diff
@@ -2980,3 +2980,30 @@ H2_DECODER_ON_SERVER_PREFACE_TEST(h2_decoder_err_bad_preface_from_client_3) {
...
+
+    /* clang-format off */
+    uint8_t input[] = {
+        /* HEADERS FRAME - no END_HEADERS, so a CONTINUATION is expected */
+        0x00, 0x00, 0x08,           /* Length (24): 8 bytes of payload */
+        AWS_H2_FRAME_T_HEADERS,     /* Type (8) */
+        0x00,                       /* Flags (8): no END_HEADERS */
+        0x00, 0x00, 0x00, 0x01,     /* Reserved (1) | Stream Identifier (31): stream 1 */
+        0x00,                       /* Literal header field, new name, never indexed */
+        0x7f, 0xa1, 0x8c, 0x06,    /* String length = 100000 */
```
azkrishpy: we should try to check default + 1 instead.

azkrishpy (approved):

source/h2_decoder.c
sbSteveK: I'm using an API call in case someone calls `aws_h2_decoder_set_setting_max_header_list_size` on h2.

source/hpack_decoder.c
sbSteveK: done

tests/test_h2_decoder.c
sbSteveK: debatable. I think here it's better to just test if it's over that it fails because we might someday change the default. I guess we could also try to get the static default and +1 it but I think we are testing over, not whether the >= works.

source/h2_decoder.c
azkrishpy: but we make that call from the `aws_h2_decoder_set_setting_max_header_list_size` API itself.

source/hpack_decoder.c
sbSteveK: discussed and resolved offline

source/h2_decoder.c
sbSteveK: Ah, missread the comment. I thought we were talking about `aws_hpack_decoder_set_max_header_list_size`

it's been removed

tests/test_h2_decoder.c
azkrishpy: true, we should not hardcode the default, instead use the same default from settings[] and check one above that. the comment can also just say "default from h2 settings array" instead of 32768 but I doubt that will change anytime soon. yes here the "edge case" is {default} + 1, anything over that is well over and it will fail if {default} + 1 fails. So we are trying to test the edge case I believe?

sbSteveK: sure, works for me

PR #577, opened 2026-09-01
Remove unreachable OOM handling

azkrishpy:
*Issue #, if available:* N/A

*Description of changes:*

CRT allocators abort on failure rather than returning NULL. They route through `AWS_PANIC_OOM`, an unconditional `exit(-1)`, so every OOM branch guarded by a NULL check was unreachable.

Removes those checks and the error paths behind them, and folds acquire-then-zero pairs into `aws_mem_calloc`. No assert replaces them: it would be dead in debug and compiled out in release, since the abort happens before control returns.

No behavior change. Existing suite passes unchanged.

### DIFF
--- a/source/connection_monitor.c
+++ b/source/connection_monitor.c
@@ -204,15 +204,13 @@ struct aws_crt_statistics_handler *aws_crt_statistics_handler_new_http_connectio
     struct aws_crt_statistics_handler *handler = NULL;
     struct aws_statistics_handler_http_connection_monitor_impl *impl = NULL;
 
-    if (!aws_mem_acquire_many(
-            allocator,
-            2,
-            &handler,
-            sizeof(struct aws_crt_statistics_handler),
-            &impl,
-            sizeof(struct aws_statistics_handler_http_connection_monitor_impl))) {
-        return NULL;
-    }
+    aws_mem_acquire_many(
+        allocator,
+        2,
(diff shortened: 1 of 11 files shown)


source/connection_monitor.c
```diff
@@ -204,15 +204,13 @@ struct aws_crt_statistics_handler *aws_crt_statistics_handler_new_http_connectio
     struct aws_crt_statistics_handler *handler = NULL;
     struct aws_statistics_handler_http_connection_monitor_impl *impl = NULL;
```
TingDaoK, line 206: trivial: let's add a comment to this function say this cannot fail. Since when we use it, we stopped to check for failure from this functions.

TingDaoK (approved):

source/connection_monitor.c
TingDaoK: Or, we can move the allocation out and restructure this function to `_init_`, which takes the pre-allocated address and make the init function as a void function instead.

DmitriyMusatkin (approved):

source/connection_monitor.c
azkrishpy: as discussed, restructure api is not a possibility now, however adding explicit comments in all header functions to convey that it will never return NULL as OOM will essentially cause an abort - Similar to the description on the PR.

PR #582, opened 2026-09-08
fix: harden HPACK decoding and table sizing

azkrishpy:
*Issue #, if available:* N/A

*Description of changes:*

Two HPACK conformance problems, both from TODOs left in the code.

The encoder took the peer's `SETTINGS_HEADER_TABLE_SIZE` at face value and passed it straight through as the requested table size. Anything above the 16MB this implementation supports is rejected outright, so a peer advertising a larger table made every resize fail, and with it every header block we tried to send. The value is now clamped to the supported ceiling, which RFC-7541 4.2 allows since the peer's number is a limit we may use, not one we must.

The decoder did not check the padding bits at the end of a Huffman-coded string. RFC-7541 5.2 requires padding to be the most significant bits of the EOS symbol and shorter than 8 bits, and says anything else MUST be treated as a decoding error. Both conditions are now enforced.

Worth knowing on review: the padding check rejects strings that previously decoded, and any HPACK failure becomes a connection error, so a peer that pads incorrectly now loses the whole connection rather than one stream. That is what the spec asks for, but the blast radius is the connection.


include/aws/http/private/hpack.h
```diff
@@ -10,6 +10,11 @@
 #include <aws/common/hash_table.h>
 #include <aws/compression/huffman.h>
 
+/* The largest dynamic table this implementation will allocate, whatever the peer's
+ * SETTINGS_HEADER_TABLE_SIZE says. A macro so both hpack.c and hpack_encoder.c can use it
+ * without exporting a new symbol. */
+#define AWS_HPACK_MAX_DYNAMIC_TABLE_SIZE (16 * 1024 * 1024)
```
azkrishpy, line 16: This is a guess there can probably be lower. Please leave a comment if you have a better estimate on what the lower number should be.

TingDaoK (approved):

PR #586, opened 2026-09-15
adding ResponseFirstByteTimeout config

kai-ion:
*Issue #, if available:*
https://github.com/aws/aws-sdk-cpp/issues/3923

*Description of changes:*
Implement response_first_byte_timeout_ms for HTTP/2.

On HTTP/2, a peer that accepts the stream but never begins responding leaves the request with no timeout anywhere in the chain, so the stream can stall indefinitely


source/h2_stream.c
```diff
@@ -680,6 +688,101 @@ static struct aws_h2err s_send_rst_and_close_stream(struct aws_h2_stream *stream
...
+    AWS_H2_STREAM_LOGF(
+        INFO,
+        stream,
+        "Timed out waiting for first byte of response after sending the full request, "
+        "response_first_byte_timeout_ms=%" PRIu64 ". Resetting stream.",
+        timeout_ms);
+
+    /* Reset the stream; it completes with AWS_ERROR_HTTP_RESPONSE_FIRST_BYTE_TIMEOUT. */
+    struct aws_h2err returned_h2err =
+        s_send_rst_and_close_stream(stream, aws_h2err_from_aws_code(AWS_ERROR_HTTP_RESPONSE_FIRST_BYTE_TIMEOUT));
```
TingDaoK: Trivial: ` aws_h2err_from_aws_code(AWS_ERROR_HTTP_RESPONSE_FIRST_BYTE_TIMEOUT)` 
This will return:
```
    return (struct aws_h2err){
        .h2_code = AWS_HTTP2_ERR_INTERNAL_ERROR,
        .aws_code = aws_error_code,
    };
```

The h2_code is the error code we embed in the rst_stream maps to https://httpwg.org/specs/rfc9113.html#ErrorCodes

I think it makes more sense to use `CANCEL` here instead of `INTERNAL_ERROR`.

